PR #68, opened 2026-01-22
test: add test cases for run.py 

### LINKED ISSUE #61: [Enhancement]: Add direct unit tests for RunMetrics and RunSummary
## Problem Statement

Run.py does not have direct unit tests; Run.add_decision(), record_outcome(), add_problem(), complete() 
are already tested via test_runtime.py through the Runtime wrapper.

## Proposed Solution

Methods/properties needing tests:
- RunMetrics.success_rate computed field (edge case: 0 decisions)
- Run.duration_ms computed field
- Run._generate_narrative() - the auto-narrative logic
- RunSummary.from_run() - aggregation and limiting logic

yumosx:
## Description

test: add test cases for run.py 

## Related Issues

fix #61

image


RichardTang-Aden:
Tried and the test failed for `test_duration_ms`

The test created two timestamps at nearly the same instant (datetime.now() called twice), resulting in a ~2 microsecond difference. The duration_ms property correctly returns int(0.002) = 0, but the test asserts against the raw float 0.002, causing a mismatch.

yumosx:
thank you for your review.
I've added `int()` conversion to the assertion. Please check.

TimothyZhang7 (approved):
Looks good to me

PR #180, opened 2026-01-24
feat(runtime): add configurable batch processing to EventBus

### LINKED ISSUE #41: [Feature]: Optimize Event Batch Size for High-Throughput Scenarios
## Problem Statement

When operating Aden at scale with 1000+ concurrent agents across multiple teams, event ingestion becomes a bottleneck. The SDK currently sends individual events to the backend, which results in:
- High network overhead and latency
- Inefficient database writes with excessive transaction overhead
- Potential event loss during network fluctuations
- High memory consumption on worker nodes processing individual events

## Proposed Solution

Implement adaptive batch sizing for SDK event transmission:

1. **Dynamic batch size calculation**: Adjust batch size based on:
   - Current network latency (increase batch size if latency is high)
   - Queue depth (if queue is backing up, increase batch size)
   - Available memory (decrease if memory pressure detected)
   - Target batch interval (e.g., 100ms or 1000 events, whichever comes first)

2. **Configuration options**:
   ```typescript
   interface BatchConfig {
     maxBatchSize: number;        // Default: 1000
     maxBatchIntervalMs: number;  // Default: 100ms
     minBatchSize: number;        // Default: 10
     adaptiveSizing: boolean;     // Default: true
   }
   ```

3. **Monitoring metrics**:
   - Batch size distribution
   - Average batch latency
   - Events dropped due to batching limits
   - CPU/memory impact of batching

## Alternatives Considered

1. **Fixed large batch size**: Simple but doesn't adapt to varying load conditions
2. **Compression**: Reduces bandwidth but adds CPU overhead
3. **Streaming events**: Low latency but high transaction overhead at scale

## Additional Context

This optimization is critical for the architecture to scale efficiently. Based on performance analysis of similar distributed systems, adaptive batching can reduce backend write load by 50-70% while maintaining <200ms end-to-end latency for critical events.

Suggested implementation:
- Start with fixed 500-event batches
- Add monitoring for 2 weeks
- Then implement adaptive sizing based on collected metrics

Ahmed-cloud-stack:
## Description

Adds batch processing to the EventBus for high-throughput scenarios. Currently events get processed one-by-one which becomes a bottleneck when you have many concurrent agents or frequent state updates. This adds optional batching that collects events and flushes them periodically, with critical events (like constraint violations) bypassing the batch for immediate delivery.

## Type of Change

## Related Issues

Fixes #41

## Changes Made

- Added `event_bus_enhanced.py` with configurable batch processing
- Added `test_event_bus_enhanced.py` with comprehensive tests
- EventBusConfig for tuning batch interval, size, and concurrency
- Priority system so critical events skip batching
- Metrics tracking for monitoring throughput and latency

## Checklist

- [x] I have added tests that prove my feature works

### DIFF
--- a/core/framework/runtime/event_bus_enhanced.py
+++ b/core/framework/runtime/event_bus_enhanced.py
@@ -1,0 +1,783 @@
+"""
+Event Bus - Pub/sub event system for inter-stream communication.
+
+Allows streams to:
+- Publish events about their execution
+- Subscribe to events from other streams
+- Coordinate based on shared state changes
+
+Enhanced with:
+- Configurable batch processing for high-throughput scenarios
+- Adaptive batch sizing based on load
+- Detailed metrics for monitoring
+- Priority-based event delivery
+"""
+
+import asyncio
+import logging
+import time
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+from typing import Any, Awaitable, Callable
+
+logger = logging.getLogger(__name__)
+
+
+class EventType(str, Enum):
+    """Types of events that can be published."""
+
+    # Execution lifecycle
+    EXECUTION_STARTED = "execution_started"
+    EXECUTION_COMPLETED = "execution_completed"
+    EXECUTION_FAILED = "execution_failed"
+    EXECUTION_PAUSED = "execution_paused"
+    EXECUTION_RESUMED = "execution_resumed"
+
+    # State changes
+    STATE_CHANGED = "state_changed"
+    STATE_CONFLICT = "state_conflict"
+
+    # Goal tracking
+    GOAL_PROGRESS = "goal_progress"
+    GOAL_ACHIEVED = "goal_achieved"
+    CONSTRAINT_VIOLATION = "constraint_violation"
+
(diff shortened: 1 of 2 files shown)


TimothyZhang7:
Can you explain the dynamic batching logic better? Also, do not use class names like enhanced_xxxx, as it is not favourable for canonical class names in the future

Ahmed-cloud-stack:
> Can you explain the dynamic batching logic better? Also, do not use class names like enhanced_xxxx, as it is not favourable for canonical class names in the future

Thanks for the review!
For the dynamic batching, basically it looks at how many events/sec are coming in. If it's below the threshold (default 100/sec), no batching. Once it goes above that, batch size scales up proportionally between min and max. So you're not adding latency when traffic is low but still get the throughput benefits under load.
For the naming, Good point I'll rename event_bus_enhanced.py to just event_bus.py and update the classes accordingly. Should I replace the existing event_bus.py or keep them separate with a different naming approach? Let me know your preference and I'll push the update.

Hundao:
Referenced issue is already closed, closing this PR.

PR #219, opened 2026-01-25
Improve error context for unknown agent resolution

Anshul-Sharma07:
This PR improves the error message when an unknown agent is referenced by including the list of available agents, making debugging easier without changing runtime behavior.

### DIFF
--- a/core/framework/runner/orchestrator.py
+++ b/core/framework/runner/orchestrator.py
@@ -273,7 +273,11 @@ class AgentOrchestrator:
             Response message from target agent
         """
         if to_agent not in self._agents:
-            raise ValueError(f"Unknown agent: {to_agent}")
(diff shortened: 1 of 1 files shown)


core/framework/runner/orchestrator.py
```diff
@@ -273,7 +273,11 @@ async def relay(
             Response message from target agent
         """
         if to_agent not in self._agents:
-            raise ValueError(f"Unknown agent: {to_agent}")
+           raise ValueError(
+    f"Unknown agent '{to_agent}'. "
+    f"Available agents: {list(self.agents.keys())}"
+)
```
akashb95, line 279: NIT: is there any way we can avoid this slightly unsightly de-dent?

PR #260, opened 2026-01-25
validation: enforce bounds on Option.confidence

Anvesh33:
Fixes #245 

What:
Adds validation to Option.confidence to ensure values remain between 0.0 and 1.0.

Why:
confidence represents probability-like certainty. Unbounded values (e.g. -1 or 500)
break semantic meaning and may cause downstream scoring issues.

Changes:
- Added Pydantic field constraints (ge=0.0, le=1.0), consistent with existing fields.

Risk:
Low. Validation pattern already exists in this schema.
Screenshot 2026-01-25 181841

### DIFF
--- a/core/framework/schemas/decision.py
+++ b/core/framework/schemas/decision.py
@@ -45,7 +45,7 @@ class Option(BaseModel):
     cons: list[str] = Field(default_factory=list)
 
     # Agent's confidence in this option (0-1)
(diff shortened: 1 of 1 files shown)


Anvesh33:
Thanks for the review and for taking a look!
Happy to adjust or contribute elsewhere if helpful.

PR #284, opened 2026-01-25
fix(graph): implement divergent cycle detection in graph validation

### LINKED ISSUE #283: [Bug]: Graph executor relies on step limits instead of cycle validation for termination
Description: The current graph framework treats all cycles as valid during the build/validation phase. It relies on a `max_steps` (default 100) safety net to halt execution of infinite loops. However, this is a "Path Length Limit" approach rather than a mathematical proof of termination, leading to wasted time/tokens.

There is a distinction between:

Convergent Cycles (Good): "Node A -> Node B. If `some_condition`, go to A." These are legitimate iterations with an exit condition.

Divergent Cycles (Bad): "Node A calls Node B. Node B calls Node A." These have no exit condition and represent a logic flaw in the graph.

Proposed Solution: Implement a `CycleDetector` that analyses the graph structure during validation. It should identify cycles where all edges are unconditional (Always, OnSuccess, etc.) and no higher-priority exit paths exist. If such a "Divergent Cycle" is found, validation should fail early rather than allowing the run to hit the execution ceiling.

WarriorOfLiberation:
Link to Issue: Closes #283 

### Summary

This PR fixes a critical flaw in the graph execution framework where divergent (inescapable) cycles were allowed to execute until the `max_steps` limit was reached.
A new static validation step now detects such cycles ahead of execution and fails fast with a clear error.

### Key Changes

* **New module:** `core/framework/graph/cycle_detector.py`

  * Implements DFS-based cycle detection
  * Classifies cycles as *divergent* when:

    * All edges in the cycle are unconditional, and
    * No higher-priority exit paths exist from any node in the cycle

* **Validation update:**

  * `GraphSpec.validate` (in `core/framework/graph/edge.py`) now invokes `CycleDetector`
  * Prevents execution of graphs containing divergent cycles

* **Tests added:** `core/tests/test_divergent_cycles.py`

  * Covers both valid convergent cycles and invalid divergent cycles
  * All tests passing (100%)

### Impact

* Prevents runaway execution caused by inescapable cycles
* Catches invalid graph structures at validation time rather than runtime
* Improves safety and predictability of graph execution

### Testing

* Added unit tests for cycle detection and classification
* Verified existing graph behavior remains unchanged for valid cycles

### DIFF
--- a/core/framework/graph/cycle_detector.py
+++ b/core/framework/graph/cycle_detector.py
@@ -1,0 +1,211 @@
+"""
+Cycle detection and validation module.
+
+Identifies "Divergent Input Cycles" - loops that have no logical exit condition
+and would spin infinitely (until max_steps).
+"""
+
+from typing import Any, List, Set, Dict
+from dataclasses import dataclass
+from framework.graph.edge import EdgeCondition, EdgeSpec
+
+@dataclass
+class Cycle:
+    """Represents a detected cycle in the graph."""
+    path: List[str]  # List of node IDs in order: A -> B -> C -> A
+    edges: List[EdgeSpec]  # The edges forming the cycle
+    is_divergent: bool
+    reason: str | None = None
+
+class CycleDetector:
+    """Detects cycles and validates if they are divergent."""
+
+    def __init__(self, graph_spec: Any):
+        """
+        Initialize with a GraphSpec.
+        
+        Args:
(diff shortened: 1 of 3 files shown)


austin931114:
@WarriorOfLiberation Great work here. Could you check the copilot comment? Also, there could be some minor lint issues in the code. Thanks!

Hundao:
Referenced issue is already closed, closing this PR.

PR #286, opened 2026-01-25
fix: add safety limit to view_file to prevent OOM crashes on large files

AdityaRajput8:
## Summary
This PR adds a safety mechanism to the `view_file` tool to prevent "Resource Exhaustion" (OOM) crashes when an agent attempts to read extremely large files.

## The Issue
Previously, `view_file.py` read the entire file content into memory:
`content = f.read()`

If an agent (or user) requested a massive file (e.g., a 5GB log file), this would cause:
1. High memory consumption (potential crash).
2. Context window overflow when passing the result back to the LLM.

## The Fix
I have implemented a `MAX_FILE_SIZE` constant (100KB).
- If the file is smaller than 100KB, it reads normally.
- If the file is larger, it reads the first 100KB and appends a `[...File truncated...]` warning.
- Returns a structured warning in the output so the agent knows the data is incomplete.

## Verification & Testing
I created a reproduction script (`verify_fix.py`) to simulate reading a 150KB file using the new logic.
- **Scenario 1 (Small File):** Verified that files under the limit are read fully.
- **Scenario 2 (Large File):** Verified that a 150KB file triggers the safety limit, truncates the content, and returns the correct warning message.

**Verification Output:**
> Status: SUCCESS (Truncated at 100KB)
> Warning Message: 'File too large... truncated'

*(See attached screenshot for execution proof)*
Screenshot 2026-01-25 at 7 01 24 PM

### DIFF
--- a/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py
+++ b/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py
@@ -35,15 +35,31 @@ def register_tools(mcp: FastMCP) -> None:
             if not os.path.exists(secure_path):
                 return {"error": f"File not found at {path}"}
 
+            # SAFETY CHECK: Prevent reading massive files that crash the agent
+            MAX_FILE_SIZE = 100 * 1024  # 100KB limit
+            file_size = os.path.getsize(secure_path)
+            
+            truncated = False
             with open(secure_path, "r", encoding="utf-8") as f:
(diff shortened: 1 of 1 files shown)


austin931114:
Hi @AdityaRajput8, thank you for your work on this.

This issue has actually been addressed in another ongoing workstream/PR, so I will close this one to avoid duplication. We appreciate the contribution, though!

PR #289, opened 2026-01-25
fix: disable ANSI colors when unsupported in setup script

Jbheemeswar:
## Description
Disables ANSI color escape sequences when the terminal does not support them (e.g., Windows Git Bash), preventing raw codes like `\033[0;34m` from appearing in setup output.

## Type of Change
- Bug fix (non-breaking)

## Testing
- Ran `./scripts/setup-python.sh` on Windows (Git Bash) with Python 3.12 venv

### DIFF
--- a/scripts/setup-python.sh
+++ b/scripts/setup-python.sh
@@ -8,12 +8,21 @@
 
 set -e
 
-# Colors for output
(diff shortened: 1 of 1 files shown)


RichardTang-Aden:

PR #310, opened 2026-01-25
docs(readme): clarify node vs python development workflows

Ayush9545:
This PR addresses issue **#298** by clarifying when Node.js vs Python tooling is required.

The goal is to improve onboarding for new contributors. Changes are documentation-only.


bryanadenhq:
Thanks for the contribution! However, the Node.js/npm workflow is actually archived and no longer needed for agent development.

Ayush9545:
Thanks for the clarification. Appreciate the context on the Node.js workflow being archived.

PR #326, opened 2026-01-25
Fix execute_command tests on Windows

MekalaKaveri18:
## Summary
Fixes `execute_command` unit tests failing on Windows due to unsupported shell utilities like `ls` and `tr`.

## Changes
- Replace `ls` with Windows-compatible directory listing command
- Skip pipe-based test on Windows where shell piping differs

## Why
These tests currently fail on Windows CI/local environments even though the tool works correctly. This makes onboarding and cross-platform development harder.

## Result
`pytest tools/tests` now passes cleanly on Windows.

### DIFF
--- a/tools/tests/tools/test_file_system_toolkits.py
+++ b/tools/tests/tools/test_file_system_toolkits.py
@@ -1,5 +1,8 @@
 """Tests for file_system_toolkits tools (FastMCP)."""
+import cmd
 import os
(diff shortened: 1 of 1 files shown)


MekalaKaveri18:
This PR fixes Windows-specific failures in execute_command tests by using cross-platform commands and skipping unsupported pipe behavior.

PR #331, opened 2026-01-25
fix(executor): allow resuming steps from AWAITING_APPROVAL status

### LINKED ISSUE #328: [Bug]: Execution Breaks after pausing for human approval
`FlexibleGraphExecutor` fails to resume after pausing for human approval. Test cases don't check for the behaviour either.

WarriorOfLiberation:
## Summary

Fixed a critical bug where `FlexibleGraphExecutor` would fail to resume execution after pausing for human input.

## Problem

When a step requires human approval (`requires_approval=True`), the executor correctly pauses execution and sets the step status to `AWAITING_APPROVAL`.

However, when `execute_plan` is called again to resume execution (for example, after the user approves the step), the executor fails to pick the step back up. This occurs because `PlanStep.is_ready()` strictly checked for `self.status == StepStatus.PENDING` and ignored steps in `AWAITING_APPROVAL`.

As a result, the executor incorrectly concluded that there were no executable steps available, causing a soft-lock where execution could not resume.

## Fix

Updated `PlanStep.is_ready()` in `core/framework/graph/plan.py` to treat both `PENDING` and `AWAITING_APPROVAL` statuses as valid candidate states for execution, subject to dependency checks.

## Verification

* Added regression test `test_execution_approval_and_resume` in `core/tests/test_flexible_executor.py` to validate the Pause -> Approve -> Resume workflow
* Verified all existing tests pass.

Fixes #328

### DIFF
--- a/core/framework/graph/plan.py
+++ b/core/framework/graph/plan.py
@@ -159,7 +159,7 @@ class PlanStep(BaseModel):
 
     def is_ready(self, completed_step_ids: set[str]) -> bool:
         """Check if this step is ready to execute (all dependencies met)."""
-        if self.status != StepStatus.PENDING:
+        if self.status not in (StepStatus.PENDING, StepStatus.AWAITING_APPROVAL):
             return False
         return all(dep in completed_step_ids for dep in self.dependencies)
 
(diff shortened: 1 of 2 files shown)


shiva9198 (commented):
```markdown
Excellent fix! The [is_ready()](cci:1://file:///Users/shiva/Documents/hive/core/framework/graph/plan.py:159:4-163:74) change is elegant and the test is comprehensive.
Tested locally - works as expected. The step correctly resumes after approval callback is set.
✅ LGTM

PR #335, opened 2026-01-25
clarify Python vs Node.js development workflows

haneeo3:
This PR adds a "Development Workflows" section to clarify when to use Python vs Node.js tooling. 

- Python setup: for building, testing, and running agents.
- Node.js setup: for JavaScript CLI tools or frontend scripts.
- Helps new contributors choose the right setup.


bryanadenhq:
Thanks for the contribution! However, the Node.js/npm workflow is actually archived and no longer needed for agent development.

PR #344, opened 2026-01-25
docs: clarify missing exports directory and agent validation steps

srinuk9570:
This PR clarifies the usage of the `exports/` directory in the setup guide.

Currently, new contributors encounter `ModuleNotFoundError` because
`exports/` is referenced but not present by default.

This documentation-only change:
- Clarifies when and how `exports/` is created
- Prevents confusion during agent validation
- Improves onboarding for contributors without Claude Code access


austin931114 (commented):
close for outdated doc

PR #379, opened 2026-01-25
Fix/executor max retries

Waqar53:
Summary
The current 
GraphExecutor
 implementation in 
core/framework/graph/executor.py
 ignores the node_spec.max_retries configuration, using a hardcoded value of 3 for all nodes regardless of user settings.

The Problem
Hardcoded Retry Limit: Line 184 declares max_retries_per_node = 3 as a fixed constant, completely bypassing the NodeSpec.max_retries field that users expect to control retry behavior.

User Configuration Ignored: When users define a node with max_retries=10 (for flaky operations) or max_retries=1 (for fail-fast scenarios), their configuration is silently discarded.

No Per-Node Control: All nodes in a graph are forced to use the same retry limit, preventing granular control over execution behavior for different node types.

Proposed Solution
Refactor the retry logic in GraphExecutor.execute() to read from node_spec.max_retries:

Remove Hardcoded Value:

Delete max_retries_per_node = 3 from the execution initialization.
Read Per-Node Configuration:

Inside the retry loop, read max_retries = node_spec.max_retries to respect each node's configured limit.
Update All References:

Update logging, error messages, and comparison logic to use the node-specific value.
Impact
Flexibility: Each node can now have its own retry strategy.
Backward Compatible: Default is still 3 - existing code works unchanged.
User Expectations Met: The 
max_retries
 field on 
NodeSpec
 now actually works.
Affected Files
core/framework/graph/executor.py
core/tests/test_executor_max_retries.py
 (new)
Good
Bad
0 Files With Changes
Review Changes

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -181,7 +181,6 @@ class GraphExecutor:
         total_tokens = 0
         total_latency = 0
         node_retry_counts: dict[str, int] = {}  # Track retries per node
-        max_retries_per_node = 3
 
         # Determine entry point (may differ if resuming)
         current_node_id = graph.get_entry_point(session_state)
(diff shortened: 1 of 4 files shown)


bryanadenhq:
resolved with #397

PR #395, opened 2026-01-25
Add token budgets, circuit breaker, and remove recursive LLM cleanup

agarwal-tanmay-work:
This PR hardens LLM execution against runaway token usage and cost amplification.

### Key changes
- Adds per-session and per-call token budgets to LiteLLMProvider
- Introduces a circuit breaker after consecutive failures
- Tracks session-level token usage and estimated cost
- Removes recursive LLM cleanup calls during JSON extraction
- Fails explicitly on malformed outputs instead of triggering hidden LLM calls

### Why this matters
Under retries, tool loops, or malformed LLM outputs, the previous behavior could result in
multiple cascading LLM calls from a single request, leading to unexpected API spend and
difficult-to-debug failures.

These changes make execution bounded, observable, and safer for production use.

### Scope
- `framework/llm/litellm.py`
- `framework/node.py`

Happy to adjust defaults or configuration based on feedback.

### DIFF
--- a/core/framework/graph/node.py
+++ b/core/framework/graph/node.py
@@ -693,63 +693,24 @@ class LLMNode(NodeProtocol):
             except json.JSONDecodeError:
                 pass
 
-        # All local extraction methods failed - use LLM as last resort
-        # Prefer Cerebras (faster/cheaper), fallback to Anthropic Haiku
-        import os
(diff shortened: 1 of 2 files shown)


bryanadenhq:
closing because stale, please rebase if still interested

PR #402, opened 2026-01-25
 Unsanitized expression evaluation in EdgeSpec (RCE Vulnerability)

Shamanth-8:
## Summary
Fixed a critical Remote Code Execution (RCE) vulnerability in `EdgeSpec` where `eval()` was used on unsanitized input.

## Changes
- Implemented `core/framework/graph/safe_eval.py` using `ast.parse` to allow only safe operations (arithmetic, comparisons, boolean logic) and whitelisted functions.
- Updated `core/framework/graph/edge.py` to use `safe_eval()` instead of `eval()`.
- Verified that arbitrary code execution (e.g., `__import__('os')`) is blocked while valid expressions work.

## Vulnerability Details
The previous implementation allowed execution of arbitrary Python code via the `condition_expr` field. This has been replaced with a strict AST walker that prevents access to `__builtins__`, imports, and private attributes.

### DIFF
--- a/core/framework/graph/edge.py
+++ b/core/framework/graph/edge.py
@@ -11,9 +11,10 @@ our edges can be created dynamically by a Builder agent based on the goal.
 
 Edge Types:
 - always: Always traverse after source completes
+- always: Always traverse after source completes
 - on_success: Traverse only if source succeeds
 - on_failure: Traverse only if source fails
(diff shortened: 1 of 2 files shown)


bryanadenhq (approved):
thanks for your contribution!

PR #407, opened 2026-01-25
Fix/starlette dependency

dadeteye1:
**Issue Addressed**

- The FastAPI-powered components within the tools package began failing when the Starlette dependency was upgraded beyond version 0.42.
- Starlette 0.42+ introduced breaking changes:
- Removal of anyio integration.
- Altered routing behaviors that FastAPI relied on.
- Because the dependency was previously unconstrained, fresh installs could inadvertently pull these newer versions.

**This led to two classes of problems:**

- Hard failures: runtime import errors when expected modules or functions were missing.
- Soft regressions: subtle request-handling differences that broke assumptions in the MCP tools server and FastAPI routes.

**Fix Applied**

Commit https://github.com/adenhq/hive/commit/ee2bc070cbdbc8598340dd4ca92b4e0e987d325d (chore: pin starlette for FastAPI compatibility) introduced a strict dependency bound in pyproject.toml:

toml
starlette >=0.40, <0.42

**This ensures that:**

- Pip installs only the Starlette versions (0.40 and 0.41) known to be compatible with the current FastAPI stack.
- Future installs are protected from pulling incompatible releases.
- The MCP tools server and FastAPI routes remain stable and predictable.

**Impact**

- Prevents runtime errors caused by missing imports.
- Guards against subtle routing regressions that could compromise request handling.
- Provides a clear compatibility boundary until the codebase can be updated to support newer Starlette releases.
- Improves developer experience by ensuring consistent behavior across environments and deployments.

### DIFF
--- a/CONTAINERIZATION.md
+++ b/CONTAINERIZATION.md
@@ -1,0 +1,55 @@
+# Containerized Hive Workspace
+
+This repo can be run entirely inside Docker so you don’t need to install every dependency on the host. The provided `Dockerfile` builds an image with the Python packages (`core`, `tools`) installed in editable mode, and the accompanying `docker-compose.yml` file orchestrates the agent shell plus the MCP tools server.
+
+## Build the image
+
+```bash
+docker build -t aden-hive:latest .
+```
+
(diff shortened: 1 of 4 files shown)


AadiSharma49:
Related Issues
Fixes #(issue number) please mention the issue you solved !!

PR #418, opened 2026-01-25
docs: clarify Python setup for PEP 668 environments

Cyril17-cyber:
This PR adds a note in the README files to help macOS/Homebrew users avoid the externally-managed-environment error when running ./scripts/setup-python.sh. 
It adds the note in all the languages README is available in.


austin931114 (commented):
close for outdated doc

PR #425, opened 2026-01-25
fix(tool-registry): implement runtime argument validation for tool execution

0bhi:
## Description

Implements runtime argument validation for tool execution in `ToolRegistry.register_function()` to prevent system crashes when LLMs provide incorrect or malformed arguments.

## Type of Change

## Related Issues

Fixes #168

## Changes Made

- Added validation for missing required parameters
- Added validation for extra/unexpected parameters (unless **kwargs present)
- Added basic type checking for int, float, bool, str annotations
- Used inspect.signature.bind_partial() for robust argument binding
- Improved error messages with descriptive ValueError exceptions

## Testing

## Checklist

- [ ] I have made corresponding changes to the documentation (if needed)


0bhi:
Hi @RichardTang-Aden

This PR implements runtime argument validation for tool execution. Could you please review it when you have a moment? Let me know if any changes are needed.

Thanks!

core/framework/runner/tool_registry.py
```diff
@@ -108,7 +109,116 @@ def register_function(
...
+                            f"got {type(value).__name__}"
+                        )
+                elif annotation is str:
+                    if not isinstance(value, str):
+                        raise ValueError(
+                            f"Tool '{tool_name}' parameter '{param_name}' must be str, "
+                            f"got {type(value).__name__}"
+                        )
+
+            # All validation passed, call the function with bound arguments
```
yumosx: Can you extract this verification and parsing into a separate function? It's too long to be placed here.

yumosx:
> Hi @RichardTang-Aden
> 
> This PR implements runtime argument validation for tool execution. Could you please review it when you have a moment? Let me know if any changes are needed.
> 
> Thanks!

Secondly, you need to assign it under the issue.

0bhi:
@yumosx @RichardTang-Aden I've addressed the feedback by extracting the validation into a separate method. The PR is ready for review. Thanks!

yumosx:
@0bhi can you add a unit test for it?

0bhi:
@yumosx  I've added tests. The PR is ready for review.

RichardTang-Aden (changes requested):
Tested and Core Claim Is Incorrect. Suggesting no furthur change for this PR

The issue (#168) states: *"System crashes when the LLM hallucinating arguments."* This is not accurate. The existing code at `tool_registry.py:215-220` already has a blanket exception handler in `get_executor()`:

```python
except Exception as e:
    return ToolResult(
        tool_use_id=tool_use.id,
        content=json.dumps({"error": str(e)}),
        is_error=True,
    )
```

**Every** exception from tool execution — including `TypeError` from bad arguments — is already caught, converted to a `ToolResult(is_error=True)`, and sent back to the LLM. The system does not crash. The LLM sees the error and can retry with corrected arguments.

### What This PR Actually Changes

| Behavior | Before PR | After PR |
|---|---|---|
| Missing required args | `TypeError` → caught → `ToolResult(is_error=True)` | `ValueError` → caught → `ToolResult(is_error=True)` |
| Extra unexpected args | `TypeError` → caught → `ToolResult(is_error=True)` | `ValueError` → caught → `ToolResult(is_error=True)` |
| Wrong types (e.g. str→int) | Error in function body → caught → `ToolResult(is_error=True)` | `ValueError` at validation → caught → `ToolResult(is_error=True)` |

The end result is identical in all cases: `ToolResult(is_error=True)` with an error message string. The PR changes `TypeError` to `ValueError` and slightly rewords the messages, but the observable behavior for the LLM (and the system stability) is the same.

### The Only New Behavior (Type Checking) Has Significant Issues

The basic type checking for `int`, `float`, `bool`, `str` is the one thing that doesn't exist today, but:

1. **Incomplete** — Only checks 4 primitive types. Ignores `list`, `dict`, `tuple`, and any complex/custom types.
2. **Limited value** — LLM tool calls come from JSON, which only has string, number, boolean, null, array, object. The tool definitions already include JSON Schema type information that constrains types at the API level.
3. **Bool/int bug** — In Python, `bool` is a subclass of `int`, so `isinstance(True, int)` returns `True`. The int check would incorrectly pass for boolean values.
4. **~150 lines of complexity** for marginal benefit over the existing `except Exception` handler.

### Other Concerns

- The `self`/`cls` filtering from inputs is dead code — bound methods already exclude `self` from their signature, and LLMs would never send `self` as a tool argument.
- The string annotation parsing (`split("|", 1)[0].strip()`) is a fragile heuristic that duplicates what `get_type_hints()` (already called above it) should handle.
- The test file doesn't include any integration test demonstrating the "crash" scenario it claims to fix.

### Suggestion

If clearer error messages for the LLM are desired, a much simpler approach would be to catch `TypeError` specifically in the existing executor closure and reformat the message — roughly 5 lines instead of 150:

```python
def executor(inputs: dict) -> Any:
    try:
        return func(**inputs)
    except TypeError as e:
        raise ValueError(f"Tool '{tool_name}' argument error: {e}") from e
```

This would be caught by the existing `except Exception` handler and produce a cleaner message without the complexity of full signature validation.

PR #426, opened 2026-01-25
docs: clarify optional presence of exports directory

mkk1254:
This PR clarifies the Project Structure section in DEVELOPER.md to note that
example agent packages under `exports/` may not be present in all repository
distributions or branches.

No functional changes.


austin931114 (commented):
close for outdated doc

PR #454, opened 2026-01-25
Clarify onboarding flow and agent generation requirements

panakantinandu:
### Summary
This PR clarifies the onboarding flow for first-time contributors by explaining
that exported agents are generated dynamically and are not present in a fresh
clone of the repository.

### Changes
- Document that the repo starts in a framework-only state
- Clarify when and how the `exports/` directory is created
- Call out the Anthropic API credit requirement for agent generation
- Provide a framework-only exploration path for contributors

### Motivation
While setting up the project, I noticed that the README assumes exported agents
already exist, which can be confusing for new contributors and candidates.
This change aims to reduce that initial friction without altering behavior.


bryanadenhq:
closing because duplicate

PR #475, opened 2026-01-26
Fix validate() fallback to warn on model-specific API keys

aiSynergy37:
## Summary\n- Use _get_api_key_env_var(self.model) when den_tools is unavailable\n- Add the provider-specific env var to missing_credentials\n- Include model name in the warning for clarity\n\n## Testing\n- Not run (logic-only change)\n\nCloses #474


TimothyZhang7 (approved):
nice catch

PR #479, opened 2026-01-26
Fix get_secure_path absolute path handling on Windows

MekalaKaveri18:
## Summary
Fixes sandbox path resolution so absolute paths like `/etc/passwd` are correctly treated as relative inside the session directory on Windows.

## Problem
`os.path.commonpath` behaved inconsistently when handling POSIX-style absolute paths on Windows, causing sandbox validation to raise incorrectly.

## Fix
- Normalize `session_dir` using `abspath`
- Strip both `/` and `\` prefixes from absolute paths before joining
- Preserve strict sandbox enforcement

## Test Results
```bash
pytest tools/tests/tools/test_security.py -v

### DIFF
--- a/tools/src/aden_tools/tools/file_system_toolkits/security.py
+++ b/tools/src/aden_tools/tools/file_system_toolkits/security.py
@@ -1,28 +1,27 @@
 import os
 
-# Use user home directory for workspaces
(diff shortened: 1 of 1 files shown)


bryanadenhq:
Hi @MekalaKaveri18, could you please tag the issue that this is related to?

PR #523, opened 2026-01-26
fix: add utf-8 encoding when saving run results on Windows

GauravNardia:
Fixes UnicodeEncodeError on Windows when saving run results.

Windows defaults to a non-UTF-8 encoding, causing failures when writing
Unicode characters. This change explicitly specifies UTF-8 when opening
run and summary files.

Fixes #505


bryanadenhq:
resolved with #762

PR #529, opened 2026-01-26
fix: replace deprecated asyncio.get_event_loop() with get_running_loop()

### LINKED ISSUE #525: [Bug]: ConcurrentStorage uses deprecated asyncio.get_event_loop() causing DeprecationWarning on Python 3.10+
## Bug Description
The `ConcurrentStorage` class in `core/framework/storage/concurrent.py` uses `asyncio.get_event_loop()` which is deprecated since Python 3.10. This causes `DeprecationWarning` messages and will break in future Python versions.

## Location
- **File:** `core/framework/storage/concurrent.py`
- **Lines:** 147, 172, 194, 209, 225, 235, 243, 250, 257, 357

## Environment
- Python: 3.11+ (as required by `pyproject.toml`)
- Hive version: main branch

## Steps to Reproduce
1. Run any code that uses `ConcurrentStorage` with Python 3.12+
2. Enable deprecation warnings: `python -W default::DeprecationWarning`
3. Observe the warnings about `get_event_loop()` being deprecated

## Expected Behavior
No deprecation warnings. Code should use `asyncio.get_running_loop()` since all usages are within async functions.

## Actual Behavior
`DeprecationWarning: There is no current event loop` may appear, and code will fail in future Python versions.

## Suggested Fix
Replace all 10 occurrences of `asyncio.get_event_loop()` with `asyncio.get_running_loop()`.

Since all usages are inside `async` functions where an event loop is guaranteed to be running, `get_running_loop()` is the correct choice.

## References
- [Python asyncio deprecation notice](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop)

Sourabsb:
## Summary
Replace all 10 occurrences of `asyncio.get_event_loop()` with `asyncio.get_running_loop()` in `ConcurrentStorage` class.

## Issue
Fixes #525 

## Changes
- Replaced deprecated `asyncio.get_event_loop()` with `asyncio.get_running_loop()` in `core/framework/storage/concurrent.py`
- Total: 10 replacements

## Why
- `asyncio.get_event_loop()` is deprecated since Python 3.10
- All usages are inside `async` functions where an event loop is guaranteed to be running
- `asyncio.get_running_loop()` is the correct replacement for this context

## Testing
The change is safe because:
1. All modified code is inside `async` functions
2. `get_running_loop()` will always work inside an async context
3. The behavior is identical - both return the current event loop


RichardTang-Aden:
We’ve received many issues regarding the Hive framework’s memory system, and we’ve encountered the same problems during internal testing. We’re currently redesigning the memory system. Please check the details in #2480.

PR #534, opened 2026-01-26
fix: properly close MCP session and STDIO context managers in disconnect()

### LINKED ISSUE #532: [Bug]: MCPClient.disconnect() doesn't properly close session and STDIO context managers (resource leak)
## Summary
The `MCPClient.disconnect()` method doesn't properly close the MCP session and STDIO context managers, causing resource leaks and potential zombie processes.

## Problem
In `_connect_stdio()`, context managers are entered with `__aenter__()`:
- `await self._stdio_context.__aenter__()`
- `await self._session.__aenter__()`

But in `disconnect()`, they are just set to `None` without calling `__aexit__()`:
- `self._session = None` 
- `self._stdio_context = None`

## Location
- **File:** `core/framework/runner/mcp_client.py`
- **Method:** `disconnect()` (lines 393-418)

## Why This Is a Bug
- MCP server subprocess may not terminate properly
- File handles/pipes remain open
- Potential zombie processes on Unix systems
- Memory leak from unclosed resources

## Fix
Properly call `__aexit__()` on both context managers before clearing references using `asyncio.run_coroutine_threadsafe()`.

Sourabsb:
## Summary
Added proper cleanup for MCP session and STDIO context managers in `MCPClient.disconnect()` to prevent resource leaks.

## Issue
Fixes #532

## Changes
- Added `_cleanup_stdio_async()` method to properly call `__aexit__()` on session and stdio_context
- Modified `disconnect()` to run async cleanup before stopping the event loop
- Added error handling with logging for cleanup failures

## Why
- Previously, context managers were just set to `None` without calling `__aexit__()`
- This caused resource leaks, unclosed file handles, and potential zombie processes
- Now properly closes resources using `asyncio.run_coroutine_threadsafe()`

## Testing
The change is safe because:
1. Cleanup runs with a 5-second timeout to prevent hanging
2. Errors during cleanup are caught and logged (best-effort cleanup)
3. Existing behavior is preserved - references are still cleared after cleanup


core/framework/runner/mcp_client.py
```diff
@@ -389,13 +389,44 @@ def _call_tool_http(self, tool_name: str, arguments: dict[str, Any]) -> Any:
...
-            # Stop event loop - this will cause context managers to clean up naturally
-            if self._loop and self._loop.is_running():
+            # Properly close session and context managers before stopping loop
+            if self._loop.is_running():
+                try:
+                    cleanup_future = asyncio.run_coroutine_threadsafe(
+                        self._cleanup_stdio_async(),
+                        self._loop
+                    )
+                    cleanup_future.result(timeout=5)  # Wait for cleanup with timeout
```
yumosx: I found that there is a function called `run_sync` in this module. Wouldn't it be better for us to just use it directly?

yumosx (changes requested):

core/framework/runner/mcp_client.py
Sourabsb: Thanks for the review! You're right that the code looks similar to `_run_async()`. However, there's a subtle but important difference:

`_run_async()` only checks `if self._loop is not None` before using `run_coroutine_threadsafe()`. But in `disconnect()`, I need to also check `if self._loop.is_running()` because:

1. **If loop is running**: Use `run_coroutine_threadsafe()` for cleanup, then stop the loop
2. **If loop exists but NOT running** (e.g., crashed or stopped externally): `run_coroutine_threadsafe()` would hang forever waiting for a result that never comes - the fallback uses `asyncio.run()` instead

If we used `_run_async()` directly when the loop exists but isn't running, it would deadlock on `future.result()`.

That said, I could potentially refactor `_run_async()` to also check `is_running()`, but that might change behavior for other callers. Let me know if you'd prefer that approach!

yumosx: > 2. **If loop exists but NOT running** (e.g., crashed or stopped externally): `run_coroutine_threadsafe()` would hang forever waiting for a result that never comes - the fallback uses `asyncio.run()` instead

Yes you are right, This seems to be a bug in the `run_sync` function. We should check whether this loop is running.

cc @RichardTang-Aden @TimothyZhang7 @bryanadenhq

Sourabsb: Thanks for the feedback and bringing this to the maintainers! 

You're right that this could be fixed at the root level in `_run_async()`. I understand both approaches:

**Option A (Current PR):** Handle the `is_running()` check specifically in `disconnect()` where we need precise control over the shutdown sequence.

**Option B (What you suggested):** Add the `is_running()` check to `_run_async()` itself so the fix applies everywhere it's used, and simplify `disconnect()`.

I'm open to either approach - let me know what the maintainers prefer! If they want Option B, I can add that small improvement to `_run_async()` and update the `disconnect()` method accordingly.

yumosx: I prefer option B, as it can reduce repeated code. 😀 
we can open a separate PR for this.

Sourabsb: Perfect! I'll create a separate PR to improve _run_async() with the is_running() check to reduce repeated code. This PR can stay as-is, and we'll follow up with that improvement.

TimothyZhang7:
@Sourabsb please fix lint issues

Sourabsb:
@TimothyZhang7, I've fixed all the lint issues.

yumosx:
just a small suggestion from my side.

We can first merge https://github.com/adenhq/hive/pull/638 and then update the code to directly use `run_sync`. Wouldn't that be better? The current approach is causing a lot of duplication.

RichardTang-Aden:
Fix is correct. Cleanup ordering mirrors inverse of init, async cleanup runs before loop shutdown, and error handling is solid. LGTM.

PR #444, opened 2026-01-25
fix(runtime): execution stream memory leak

fermano:
## Description

ExecutionStream never removes per-execution entries from _active_executions and _completion_events. In long-running services or high-throughput runs, memory grows without bound and can eventually OOM.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)

## Related Issues

Fixes #440

## Changes Made
Implement bounded retention for execution results while.

- Implemented bounded retention for execution results
- Cleaned in-flight state immediately to prevent memory growth without losing short-term query ability

## Testing

Describe the tests you ran to verify your changes:

- [x] Unit tests pass (`cd core && pytest tests/`)
- [ ] Lint passes (`cd core && ruff check .`)
- [ ] Manual testing performed

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Screenshots (if applicable)

Add screenshots to demonstrate UI changes.

PR #549, opened 2026-01-26
fix(core): critical memory leak in ExecutionStream task management

kawacukennedy:
This PR addresses the ExecutionStream memory leak identified in #549.

**Issue:**
Async tasks created for executions were not properly cleaned up in exception scenarios. This caused tasks to remain in the `_execution_tasks` dictionary indefinitely, leading to unbounded memory growth in long-running agents.

**Changes:**
- Updated cleanup logic to catch all exceptions, not just `CancelledError`
- Removed tasks from `_execution_tasks` after completion or failure
- Ensures no orphaned tasks remain, preventing memory leaks

**Files Modified:**
- `core/framework/runtime/execution_stream.py`

**Testing:**
- Verified cleanup manually
- Ran full `pytest` suite in `core/` and `tools/`

**Impact:**
- Prevents memory leaks for long-running agents
- Improves reliability and stability in production deployments
- High-value fix for core framework

Please review and advise if further improvements or tests are needed.

### DIFF
--- a/core/framework/runtime/agent_runtime.py
+++ b/core/framework/runtime/agent_runtime.py
@@ -221,6 +221,8 @@ class AgentRuntime:
         async with self._lock:
             # Stop all streams
             for stream in self._streams.values():
+                # Ensure cleanup of any orphaned tasks before stopping
+                await stream.cleanup_orphaned_tasks()
                 await stream.stop()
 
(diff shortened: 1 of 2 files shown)


bryanadenhq:
resolved by #444

PR #553, opened 2026-01-26
Fix/mock mode llm node

AnmollCodes:
## Objective
Fixes #539

The `Online Research Agent` example (and potentially other agents) fails immediately when running in `--mock` mode.

**Root cause:**
During mock execution, the runtime correctly sets `llm=None`, but the core [LLMNode] implementation treated a missing LLM as a fatal error ("LLM not available"). This caused the runtime to crash instantly before the graph could proceed.

## Solution
This PR introduces a native **Mock Strategy** inside the core [LLMNode](cci:2://file:///c:/Users/vansh/Downloads/hive-main/hive-main/core/framework/graph/node.py:414:0-896:31) implementation ([core/framework/graph/node.py].

Instead of failing when `ctx.llm` is `None`, the node now:

1.  **Detects Mock Mode**: Automatically identifies the missing LLM context and switches into `MOCK MODE`.
2.  **Improves Transparency**: Logs a clear warning (`Running Node ... in MOCK MODE`) so users understand why mock outputs are being produced.
3.  **Synthesizes Safe Placeholders**: Generates predictable dummy values for all configured `output_keys` (e.g. `[MOCK] Value for summary`).
    *   *Why?* Downstream nodes often depend on these keys existing in shared memory. Synthesizing them prevents cascading failures.
4.  **Prevents Side-Effects**: Memory writes in mock execution are wrapped defensively to ensure mock mode never triggers unexpected crashes.

## Impact
- **Fixes #539** completely.
- **Universal Fix**: `--mock` usage now works for *any* agent built on this framework.
- **Zero Config**: Agent implementations require no changes; mock execution works automatically when `llm` is None.

### DIFF
--- a/core/framework/graph/node.py
+++ b/core/framework/graph/node.py
@@ -448,7 +448,29 @@ class LLMNode(NodeProtocol):
         import time
 
         if ctx.llm is None:
-            return NodeResult(success=False, error="LLM not available")
+            # MOCK MODE: Generate dummy outputs if LLM is missing
+            logger.warning(f"      ⚠ Running Node '{ctx.node_spec.name}' in MOCK MODE (No LLM)")
+            
+            output = {}
+            mock_content = f"[MOCK] Response for {ctx.node_spec.name}"
+            
+            # Generate mock data for output keys
+            if ctx.node_spec.output_keys:
+                for key in ctx.node_spec.output_keys:
+                    output[key] = f"[MOCK] Value for {key}"
+                    try:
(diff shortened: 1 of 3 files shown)


AnmollCodes:
I verified the exception hierarchy. EmptyFileError and others are caught before the generic PdfReadError in the try/except block, so the order is correct. I also cleaned up the unused variable and added logging for mock mode based on feedback. Ready for merge!

Issue #561, opened 2026-01-26
[Bug]: Tool calls use inconsistent workspace_id/agent_id/session_id - files not found across steps


Hundao:
## Describe the Bug

When an agent executes multiple tool calls across different steps, the LLM generates different `workspace_id`, `agent_id`, and `session_id` values for each call. This causes files created in one step to be unfindable in subsequent steps due to sandbox path isolation (`get_secure_path`).

## To Reproduce

1. Create an agent with multiple steps that use tools requiring sandbox parameters
2. Step 1: Use `execute_command_tool` to create a file - LLM uses IDs like `workspace_id_placeholder/agent_id_placeholder/session_id_placeholder`
3. Step 2: Use another tool to read the file - LLM uses different IDs like `my_workspace/my_agent/my_session`
4. File not found error occurs because paths resolve to different sandbox directories

## Expected Behavior

Step 1: execute_command_tool(command="curl -o data.csv ...", workspace_id="workspace_id_placeholder", agent_id="agent_id_placeholder", session_id="session_id_placeholder")
File saved to: ~/.hive/workdir/workspaces/workspace_id_placeholder/agent_id_placeholder/session_id_placeholder/data.csv

Step 2: view_file(path="data.csv", workspace_id="workspace", agent_id="agent", session_id="session")
Looking in: ~/.hive/workdir/workspaces/workspace/agent/session/data.csv

Result: {"error": "File not found: data.csv"}

## Root Cause

The framework does not provide consistent session context to tool calls. The LLM has no way to know what IDs to use, so it makes up values each time.

## Expected Behavior

All tool calls within the same agent execution should use consistent `workspace_id`, `agent_id`, and `session_id` values.

## Proposed Solutions

1. **Framework auto-injection (Recommended)** - `AgentRuntime` should automatically inject session context into tool parameters before execution
2. **Tool parameter defaults** - Make these parameters optional with runtime-provided defaults
3. **Execution context passing** - Pass session IDs through `NodeContext` and inject at tool execution time

## Affected Components

- Any tool using `get_secure_path()` for sandbox isolation
- `execute_command_tool`, `view_file`, `write_to_file`, `csv_*` tools, etc.

## Environment

- Discovered during agent integration testing
- Affects all agents using sandboxed file operations

## Logs

Relevant log output:

```
▶ Step 1: Download Data (llm_tool_use)
    🔧 Tool call: execute_command_tool(
        command=curl -o happiness_report_2024.csv "https://...",
[...]

PR #563, opened 2026-01-26
fix: MCP integration example references non-existent paths

himanshu748:
## Summary
- Fixes the MCP integration example which referenced `exports/task-planner` that doesn't exist
- Adds a `create_example_agent()` helper function that dynamically creates minimal agents for demos
- Uses temporary directories with proper cleanup for all examples
- Removes dependency on pre-existing agent exports

## Test Plan
- [ ] Run `python core/examples/mcp_integration_example.py` to verify examples work
- [ ] Verify temporary directories are properly cleaned up after each example

Fixes #544

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,4 +1,4 @@
-Product Roadmap
+# Product Roadmap
 
 Aden Agent Framework aims to help developers build outcome oriented, self-adaptive agents. Please find our roadmap here
 
--- a/core/examples/mcp_integration_example.py
+++ b/core/examples/mcp_integration_example.py
@@ -6,45 +6,143 @@ This example demonstrates how to:
 1. Register MCP servers programmatically
 2. Use MCP tools in agents
 3. Load MCP servers from configuration files
+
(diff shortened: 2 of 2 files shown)


himanshu748:
## Note on CI Lint Failure

The lint check is failing due to **pre-existing errors** in other files (not in this PR):
- `examples/manual_agent.py` - unused import, trailing whitespace, f-string issues
- `framework/__init__.py` - import sorting

My changes to `examples/mcp_integration_example.py` are lint-clean. The pre-existing lint issues can be verified by the existence of [PR attempting to fix all 399 ruff errors](https://github.com/adenhq/hive/pull/561).

This PR addresses the issue by:
1. Creating a `create_minimal_agent()` helper that generates `agent.json` directly
2. Using the correct agent.json format (matching the `export_graph` output format)
3. Using temporary directories with proper cleanup for all examples
4. Removing the broken `WorkflowBuilder` usage (the class doesn't exist - correct class is `GraphBuilder`)

PR #602, opened 2026-01-26
fix(llm): correct type annotation from lowercase `callable` to `Callable`

Kira714:
## Summary

Fixes #599

The `callable` keyword in Python is a builtin function (to check if something is callable), NOT a type annotation. For type hints, we need `Callable` from the `typing` module.

## Changes

Updated all three LLM provider files:
- `core/framework/llm/provider.py`
- `core/framework/llm/anthropic.py`  
- `core/framework/llm/litellm.py`

**Before:**
```python
tool_executor: callable,
```

**After:**
```python
tool_executor: Callable[[ToolUse], ToolResult],
```

This also adds proper type information about the function signature - it takes a `ToolUse` and returns a `ToolResult`.

## Test plan

- [x] Verified with `grep` that no lowercase `callable` type hints remain
- [x] Imports added: `from typing import Callable` and `ToolResult` where needed

---

### DIFF
--- a/core/framework/llm/anthropic.py
+++ b/core/framework/llm/anthropic.py
@@ -1,9 +1,9 @@
 """Anthropic Claude LLM provider - backward compatible wrapper around LiteLLM."""
 
 import os
-from typing import Any
+from typing import Any, Callable
 
-from framework.llm.provider import LLMProvider, LLMResponse, Tool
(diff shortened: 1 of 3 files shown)


bryanadenhq (approved):
thanks for contributing, it looks good to me

PR #607, opened 2026-01-26
fix(storage): add type validation for cache entries

Kira714:
## Summary

Fixes #592

Added type validation before accessing `CacheEntry` methods to prevent `AttributeError` when cache entries are corrupted or invalid.

## Changes

Updated three locations in `core/framework/storage/concurrent.py`:

1. **load_run()**: Added `isinstance(entry, CacheEntry)` check
2. **load_summary()**: Added same validation
3. **get_cache_stats()**: Now counts invalid entries and handles non-CacheEntry values

Invalid entries are now:
- Removed from cache when detected
- Logged as warnings for debugging
- Tracked separately in cache stats

## Test plan

- [x] All cache access points now validate entry type
- [x] Invalid entries are cleaned up automatically
- [x] Stats now include `invalid_entries` count

---

### DIFF
--- a/core/framework/storage/concurrent.py
+++ b/core/framework/storage/concurrent.py
@@ -163,8 +163,12 @@ class ConcurrentStorage:
         # Check cache
         if use_cache and cache_key in self._cache:
             entry = self._cache[cache_key]
-            if not entry.is_expired(self._cache_ttl):
+            if isinstance(entry, CacheEntry) and not entry.is_expired(self._cache_ttl):
                 return entry.value
+            elif not isinstance(entry, CacheEntry):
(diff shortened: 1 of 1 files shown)


Hundao:
Closing this along with #592. The defensive isinstance checks aren't needed here — _cache is a private attribute with a clear type annotation (dict[str, CacheEntry]), and every write path (lines 140, 177, 197 in concurrent.py) only ever inserts CacheEntry instances. Adding type checks for internal state that's fully controlled by the class would set an inconsistent precedent in the codebase.

Thanks for the contribution and the thorough test plan!

PR #608, opened 2026-01-26
fix(runtime): use safe dictionary access in trigger_and_wait()

Kira714:
## Summary
- Replace direct dictionary access `self._streams[entry_point_id]` with `.get()` method
- Add explicit ValueError when stream is not found
- Prevents potential KeyError if entry point is removed between trigger() and accessing _streams

## Test plan
- [x] Syntax verified
- [ ] Existing tests should pass (no behavior change for valid cases)
- [ ] Test with invalid entry_point_id should raise ValueError instead of KeyError

Fixes #589


bryanadenhq (approved):
thanks for contributing!

PR #610, opened 2026-01-26
fix(stream): avoid private Semaphore._value attribute access

Kira714:
## Summary
- Replace access to `self._semaphore._value` with calculation from running execution count
- Private attributes are implementation details that may change between Python versions

## Changes
```python
# Before (fragile)
"available_slots": self._semaphore._value,

# After (robust)
running_count = statuses.get("running", 0)
available_slots = self.entry_spec.max_concurrent - running_count
```

## Test plan
- [x] Syntax verified
- [ ] Existing tests should pass
- [ ] Stats output remains equivalent for normal operation

Fixes #609


bryanadenhq (approved):
looks good, thanks for the contribution!

PR #619, opened 2026-01-26
Add deterministic workflow graph validation with tests and canonical example

Neeraj04-CY:
### What this PR adds

This PR introduces a deterministic pre-execution validation layer for workflow graphs to catch structural issues early and fail fast with actionable errors.

**Key additions:**
- A new `core/validation` module with explicit validation errors and results
- Graph-level validation invoked before execution
- Unit tests covering validation behavior
- A canonical example workflow demonstrating correct structure

### Why this matters

Without deterministic validation, malformed graphs can fail late during execution in non-obvious ways.
This change makes workflow construction safer, debuggable, and easier to extend.

### Scope

- No breaking changes to public APIs
- Validation runs once per execution
- Covered by tests

Happy to adjust naming, structure, or scope if you’d prefer a different integration point.


Neeraj04-CY:
Thanks for reviewing!

This change is intentionally scoped to pre-execution validation only.
Happy to adjust naming, structure, or the integration point if you’d prefer
validation to live elsewhere in the execution lifecycle.

Let me know if you’d like follow-ups such as:
- validation metrics / observability hooks
- richer error surfaces for tooling
- additional canonical examples

Neeraj04-CY:
Thanks for the detailed review — this is super helpful.

I’m going through the comments now and will push a follow-up commit that:
- aligns imports and packaging with existing `framework.*` conventions
- clarifies validator semantics (return vs raise)
- fixes `allow_cycles` precedence to match intended behavior
- tightens condition validation to better reflect runtime semantics

I’ll update the PR shortly once these are addressed.

PR #654, opened 2026-01-26
fix: respect active virtualenv on Windows during python setup

### LINKED ISSUE #650: [Bug]: Windows: setup-python.sh ignores active virtualenv and fails with multiple Python installs
## Describe the Bug

On Windows (Git Bash), `scripts/setup-python.sh` ignores the active virtual environment and always
uses `python3`, which resolves to the system Python instead of the venv Python. This causes setup
to fail with misleading errors such as "pip is not installed", even when pip exists in the venv.

## To Reproduce
1. On Windows, install multiple Python versions (e.g. 3.8, 3.11, 3.12)
2. Create and activate a virtualenv using Python 3.11
3. Run `bash scripts/setup-python.sh`
4. Script detects system Python (e.g. 3.12) instead of venv Python and fails with "pip is not installed"

## Expected Behavior
When a virtualenv is active, the setup script should use the venv's Python interpreter
and not fall back to system Python.

## Screenshots

N/A

## Environment
- OS: Windows 11
- Shell: Git Bash (MINGW64)
- Python versions installed: 3.8, 3.11, 3.12
- Virtualenv Python: 3.11.9

## Configuration

N/A

```yaml

```

## Logs

Relevant log output:

```
Detected Python: 3.12
Error: pip is not installed
Please install pip for Python 3.12
```

## Additional Context

Root cause appears to be hardcoded preference for `python3` in setup-python.sh.
On Windows, `python3` resolves to system Python even when a virtualenv is active.
I’m happy to submit a PR to address this.

AdityaGupta2804:
Fixes #650

### Summary
On Windows (Git Bash), `setup-python.sh` always preferred `python3`, which resolves to system
Python even when a virtualenv is active. This caused setup to fail with
"pip is not installed" errors.

### Changes
- Prefer the active virtualenv's `python` when `VIRTUAL_ENV` is set
- Preserve existing behavior for non-Windows environments

### Result
- Setup succeeds when a venv is active
- Avoids conflicts with multiple Python installations on Windows

### DIFF
--- a/scripts/setup-python.sh
+++ b/scripts/setup-python.sh
@@ -32,10 +32,14 @@ if ! command -v python &> /dev/null && ! command -v python3 &> /dev/null; then
     exit 1
 fi
 
-# Use python3 if available, otherwise python
-PYTHON_CMD="python3"
-if ! command -v python3 &> /dev/null; then
+# Prefer active virtualenv Python if present
+if [ -n "$VIRTUAL_ENV" ]; then
     PYTHON_CMD="python"
+else
+    PYTHON_CMD="python3"
+    if ! command -v python3 &> /dev/null; then
+        PYTHON_CMD="python"
+    fi
 fi
 
 # Check Python version


RichardTang-Aden:

PR #655, opened 2026-01-26
Avoid private semaphore access in ExecutionStream.get_stats

aiSynergy37:
## Summary\n- Compute available slots from tracked running count instead of accessing syncio.Semaphore._value\n- Keeps stats aligned with public data and avoids reliance on private internals\n\n## Testing\n- Not run (logic-only change)\n\nCloses #609


bryanadenhq:
resolved in #610

PR #689, opened 2026-01-26
fix(security): normalize unix-style paths on windows to prevent sandbox escape (fixes #499)

arshadshaik0000:
### Description
This PR addresses Issue #499.

On Windows, `get_secure_path` was rejecting valid Unix-style absolute paths (e.g., `/etc/passwd`) as being outside the sandbox, causing test failures in `test_security.py`.

### Changes
- Updated `get_secure_path` in `tools/src/aden_tools/tools/file_system_toolkits/security.py`.
- Added logic to strip both `os.sep` (backslash) and `/` (forward slash) when resolving paths to ensure consistent behavior across platforms.
- Added a `try...except ValueError` block around `os.path.commonpath` to handle Windows-specific edge cases (e.g., paths on different drives).

### Testing
- Verified that the logic correctly normalizes paths and resolves them within the session directory.
- This fix ensures cross-platform compatibility for path security checks.

### DIFF
--- a/tools/src/aden_tools/tools/file_system_toolkits/security.py
+++ b/tools/src/aden_tools/tools/file_system_toolkits/security.py
@@ -15,13 +15,19 @@ def get_secure_path(path: str, workspace_id: str, agent_id: str, session_id: str
     # Resolve absolute path
     if os.path.isabs(path):
         # Treat absolute paths as relative to the session root if they start with /
-        rel_path = path.lstrip(os.sep)
+        # FIX: Strip both native separator AND forward slash to handle Unix-style paths on Windows
+        rel_path = path.lstrip(os.sep + '/')
         final_path = os.path.abspath(os.path.join(session_dir, rel_path))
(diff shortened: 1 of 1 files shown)


bryanadenhq (changes requested):
This implementation us relying solely on isabs() which returns False for Unix-style paths on Windows. A better implementation may be to explicitly check path.startswith(("/", "\\")). Thanks for your interest in our project!

MekalaKaveri18:
@bryanadenhq if that better implemtation isn't done yet, could u please assign it to me.

arshadshaik0000:
Thanks for the review!

I’ve opened a new PR (#1204) that fully addresses the feedback about not relying solely on `os.path.isabs()` and instead explicitly handling Unix-style absolute paths on Windows using `path.startswith(("/", "\\"))`.

That PR contains the corrected implementation and cleanup, so I’ll close this one to avoid duplication.

PR #694, opened 2026-01-26
fix(stream): avoid Semaphore._value access in get_stats (#609)

dadeteye1:
Fix Summary\n• get_stats() derives available_slots from max_concurrent minus the running count tracked in statuses instead of reading the private asyncio.Semaphore._value.\n• This matches issue #609’s suggested fix by relying only on public state, keeping compatibility across Python versions.\n• The result stays non-negative to prevent regressions if Python’s Semaphore internals change.\n• All other stream metadata is unchanged while concurrency reporting becomes safer and more explicit.\n\nTesting\n• Not run (not requested).\n\nCloses #609


bryanadenhq:
resolved in #610

PR #740, opened 2026-01-26
Fix async task memory leak in ExecutionStream (#510)

ac-mmi:
Fixes a memory leak in `ExecutionStream `where async execution tasks were not removed from internal tracking dicts after completion, failure, or cancellation.

This ensures `_execution_tasks `and `_active_executions `are always cleaned up in `finally`, preventing unbounded memory growth in long-running agents.

Resolves: [#510](https://github.com/adenhq/hive/issues/510)


bryanadenhq:
resolved by #444

PR #742, opened 2026-01-26
Fix testing CLI documentation to match actual arguments

shubhankarvyas:
Fixes #717 

Updates testing CLI documentation examples to match the actual
command arguments accepted by the CLI, preventing copy-paste errors
for contributors.


austin931114 (commented):
close for outdated doc

PR #757, opened 2026-01-26
Feature/hitl lifecycle

DIBYA9:
feat(core): add Lifecycle APIs and Human Approval Tool

- Implements pause/resume in ExecutionStream and AgentRuntime
- Adds HumanApprovalTool for HITL workflows
- Adds comprehensive lifecycle test suite
- Fixes self-cancellation deadlock in pause logic


bryanadenhq:
Hi @DIBYA9, this is also currently not in our roadmap, but we do really appreciate the effort. Additionally, for future contributions, please do not include agent runtime logs or summaries. Thanks!

Issue #801, opened 2026-01-26
[Bug]: Unhandled Tool Execution Exceptions Can Crash Agent Runtime


thecodeblood:
### Description

**Summary**
The `complete_with_tools` method in the LiteLLM provider lacks proper exception handling around tool execution. While JSON parsing errors are caught, exceptions raised during actual tool execution are not handled, causing the entire LLM generation process to crash and potentially lose conversation context.

**Affected File**
- `core/framework/llm/litellm.py` - method `complete_with_tools`

**Current Behavior**
```python
# Current: Only JSONDecodeError is caught
result = tool_executor(tool_use)  # ← Exceptions here propagate and crash the runtime
```

When a tool raises an exception:
1. The entire LLM generation process crashes
2. Conversation context may be lost
3. The agent cannot recover or retry

**Expected Behavior**
Tool failures should be gracefully handled and communicated back to the LLM, allowing it to:
- Retry with different parameters
- Use an alternative tool
- Inform the user about the failure

**Impact**
- ❌ Single tool failure crashes entire agent session
- ❌ Loss of conversation context and state
- ❌ Poor user experience with no recovery options
- ⚠️ **Security concern**: Uncaught exceptions may leak sensitive stack trace information

**Proposed Solution**
Wrap tool execution in a comprehensive try/except block:

```python
try:
    result = tool_executor(tool_use)
except Exception as e:
    error_message = {
        "tool_name": tool_use.name,
        "error_type": type(e).__name__,
        "error_message": str(e),
        "status": "failed"
    }
    # Return error to LLM for graceful handling
    result = ToolResult(
        success=False,
        error=error_message
    )
    logger.error(f"Tool execution failed: {tool_use.name} - {e}")
```

**Steps to Reproduce**
1. Create a custom tool that intentionally raises an exception
2. Configure an agent to use this tool
3. Trigger the tool via LLM interaction
4. Observe: Runtime crashes instead of graceful error handling

**Additional Context**
This issue is related to the broader tool execution isolation concern noted in the security review. Implementing proper exception handling is the first step toward more robust tool execution.

**Labels**
`bug`, `priority:critical`, `component:llm`, `reliability`, `security`

zen-karia:
I would like to work on this!

m7mdhka:
@zen-karia Sorry, I worked on it already, Sorry for that.

Kerbeccs:
well i can handel it please assisgn me this issue

Hundao:
[...]

PR #809, opened 2026-01-26
fix(llm): handle tool execution exceptions in LiteLLMProvider

m7mdhka:
## Summary

Fixes a critical bug where exceptions raised during tool execution in `LiteLLMProvider.complete_with_tools()` would crash the entire LLM generation loop, causing loss of conversation context and preventing graceful error recovery.

## Problem

The `complete_with_tools()` method only caught `JSONDecodeError` when parsing tool arguments, but did not handle exceptions raised during actual tool execution. When a tool raised an exception:

- ❌ The entire LLM generation process crashed
- ❌ Conversation context was lost
- ❌ The agent could not recover or retry
- ⚠️ Security concern: Uncaught exceptions could leak sensitive stack trace information

## Solution

- Wrapped `tool_executor(tool_use)` calls in comprehensive try/except blocks
- Convert tool exceptions into structured `ToolResult` errors with error details
- Normalize unexpected return types (dict/str) from custom executors into `ToolResult`
- Added error logging for tool failures
- Tool failures are now communicated back to the LLM, allowing it to:
  - Retry with different parameters
  - Use alternative tools
  - Inform the user about the failure

## Changes

- **`core/framework/llm/litellm.py`**: Added exception handling and result normalization in `complete_with_tools()`
- **`core/tests/test_litellm_provider.py`**: Added regression test `test_complete_with_tools_tool_exception_is_captured`
- **`CHANGELOG.md`**: Documented the fix

## Testing

All 23 tests pass, including the new regression test that verifies:
- Tool exceptions are caught (no crash)
- Errors are converted to `ToolResult` with `is_error=True`
- The LLM loop continues and receives the error message
- Conversation context is preserved

uv run pytest tests/test_litellm_provider.py -v
# 23 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - N/A
 
 ### Fixed
-- N/A
+- Fixed crash in `LiteLLMProvider.complete_with_tools()` when tool execution raises exceptions. Tool failures are now gracefully handled and returned as structured error results, allowing the LLM to retry or use alternative tools instead of crashing the entire agent session.
 
 ### Security
 - N/A
--- a/core/framework/llm/litellm.py
+++ b/core/framework/llm/litellm.py
@@ -8,14 +8,18 @@ See: https://docs.litellm.ai/docs/providers
 """
 
 import json
+import logging
+from collections.abc import Callable
 from typing import Any
 
 try:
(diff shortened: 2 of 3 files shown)


m7mdhka:
#801

Hundao:
The complete_with_tools method this PR modifies no longer exists on main. Tool execution has moved to a different architecture. Closing, thanks for the effort!

PR #812, opened 2026-01-26
docs: add Windows examples for PYTHONPATH-based agent commands

saitejareddi3636:
The README previously used a Unix-style PYTHONPATH command, which doesn’t work on Windows.
This PR adds PowerShell and cmd.exe equivalents while keeping macOS/Linux instructions unchanged.

Docs-only change.


austin931114 (commented):
close for outdated doc

PR #819, opened 2026-01-26
docs(readme): add virtual environment setup instructions

AmanS3109:
**What does this PR do?** 
This PR adds instructions to create and activate a Python virtual environment before running the setup script.

**Why is this change needed?** 
On modern Linux distributions (and some macOS setups) enforcing PEP 668, running pip globally causes an externally-managed-environment error. Adding the venv step ensures the setup-python.sh script runs successfully for all users.

**Related Issue**: Documentation fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,10 @@ Aden is a platform for building, deploying, operating, and adapting AI agents:
 git clone https://github.com/adenhq/hive.git
 cd hive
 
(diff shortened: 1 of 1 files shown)


austin931114:
Thanks for bringing this up.
The problem has already been fixed in a separate PR, so I’ll close this issue. Appreciate the report!

PR #641, opened 2026-01-26
fix/windows-unicode-encoding

ap-dev-github:
Fixes UnicodeEncodeError on Windows by explicitly specifying UTF-8 encoding in all JSON file operations.

 This matches JSON RFC 7159 standard and ensures cross-platform compatibility.

## Description
Fixed UnicodeEncodeError when writing JSON files containing Unicode characters (✓ ✗ →) on Windows systems by adding explicit UTF-8 encoding to all file operations.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)

## Related Issues

Fixes #412

## Files Affected
- backend.py
- test_storage.py

## Changes Made
- Add UTF-8 encoding to file operations for Windows compatibility (#412 )
  - `core/framework/storage/backend.py`
    - `save_run()` (Line 56, 62)
    - `load_run()` (Line 66)
    - `load_summary()` (Line 73)
    - `_get_index()` (Line 127)
    - `_add_to_index()` (Line 142)
    - `_remove_from_index()` (Line 155)
  - `core/framework/testing/test_storage.py`
    - `save_test()` (Line 68)
    - `load_test()` (Line 78)
    - `save_result()` (Line 183, 193)
    - `get_latest_result()` (Line 208)
    - `get_result_history()` (Line 220)
    - `_get_index()` (Line 230)
    - `_add_to_index()` (Line 241)
    - `_remove_from_index()` (Line 253)

## Testing
- [x] Unit tests pass (`cd core && pytest tests/`)
  - Result: 184/184 tests passing 
  
- [x] Lint passes for modified files
  - `core/framework/storage/backend.py`: No issues
  - `core/framework/testing/test_storage.py`: No issues
  
- [x] Manual testing performed
  - Tested on Windows 11
  - Verified UTF-8 encoding works correctly
  - Cross-platform verified (Windows, Linux, macOS)

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] Changes are minimal and focused (UTF-8 encoding specs only)
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally (184/184)
- [x] No breaking changes to API or behavior
- [x] UTF-8 encoding is applied consistently across both files

## Screenshots 
image

PR #830, opened 2026-01-26
fix: add utf-8 encoding to file operations for Windows compatibility

AryanKansagara:
﻿Fixes #827

## Problem

On Windows, the framework fails with `UnicodeEncodeError` when saving runs that contain Unicode characters ( and  used in decision summaries). This is because Windows defaults to `cp1252` encoding which cannot handle these characters.

## Solution

Added `encoding="utf-8"` to all 7 `open()` calls in `core/framework/storage/backend.py`.

## Testing

- All 184 tests pass on Windows after this fix (previously 19 were failing)
- No impact on Linux/macOS (UTF-8 is already the default)

### DIFF
--- a/core/framework/storage/backend.py
+++ b/core/framework/storage/backend.py
@@ -52,13 +52,13 @@ class FileStorage:
         """Save a run to storage."""
         # Save full run using Pydantic's model_dump_json
         run_path = self.base_path / "runs" / f"{run.id}.json"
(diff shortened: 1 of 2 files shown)


Hundao:
Closing as duplicate — #641 covers the same fix and also addresses test_storage.py. Thanks for the contribution!

PR #1190, opened 2026-01-27
fix(setup): Fixes python and pip version detection and mismatch

mrunalnshah:
## Description
As the `setup-python.sh` was failing to detect pip correctly on Windows. I have added a cross-platform compatible robust code which detects the correct version of python and pip before proceeding from multiple python interpreter and selects the correct python. 

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [X] Documentation update
- [X] Refactoring (no functional changes)

## Related Issues

Fixes #476 

## Changes Made

- Ensured python version is found across all available python interpreters including python3, python, and py -3, and made it robust for easy interpreter add-on.
- Ensured that pip is found for the respective python interpreter.
- Generalized some variables like PYTHON_VERSION for flexiblity.
- Added a split to PYTHON_VERSION into Major and Minor version to create a robust code.
- Added clear documentation throughout the code .

## Testing

- [X] Manual testing performed on Windows and Linux for 3.12, and 3.9 python interpreter.

## Checklist

- [X] My code follows the project's style guidelines
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings

- Python 3.12
image

- Python 3.9
image


scripts/setup-python.sh
```diff
@@ -19,49 +19,56 @@ NC='\033[0m' # No Color
...
+    fi
+done
 
-# Use python3 if available, otherwise python
-PYTHON_CMD="python3"
-if ! command -v python3 &> /dev/null; then
-    PYTHON_CMD="python"
-fi
+# Check for python (Exit if not found)
+if [ -z $PYTHON_CMD ]; then
```
Hundao: `PYTHON_CMD` is initialized to "python3" before the loop, so [ -z $PYTHON_CMD ] will never be true — if no suitable python is found, it silently falls through with the default.

scripts/setup-python.sh
```diff
@@ -19,49 +19,56 @@ NC='\033[0m' # No Color
...
+    echo -e "${RED}Error: Python $PYTHON_MAJOR_VERSION.$PYTHON_MINOR_VERSION+ not found.${NC}"
+    echo "Please install Python $PYTHON_MAJOR_VERSION.$PYTHON_MINOR_VERSION+ from https://python.org"
+    exit 1
+fi  
 
-# Check Python version
+# Display Python version
 PYTHON_VERSION=$($PYTHON_CMD -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
```
Hundao: PYTHON_VERSION is defined as "3.11" at the top, variable name collision.

scripts/setup-python.sh
```diff
@@ -19,49 +19,56 @@ NC='\033[0m' # No Color
...
-    echo ""
-fi
-
 echo -e "${GREEN}✓${NC} Python version check passed"
 echo ""
 
-# Check for pip
-if ! $PYTHON_CMD -m pip --version &> /dev/null; then
+# Check for pip (Exit if not found)
+if ! eval $PYTHON_CMD -m pip --version &> /dev/null; then
```
Hundao: `$cmd` is used with eval here but without eval above — should be consistent for "py -3" to work.

scripts/setup-python.sh
```diff
@@ -19,49 +19,56 @@ NC='\033[0m' # No Color
...
 echo ""
 
-# Check for Python
-if ! command -v python &> /dev/null && ! command -v python3 &> /dev/null; then
-    echo -e "${RED}Error: Python is not installed.${NC}"
-    echo "Please install Python 3.11+ from https://python.org"
-    exit 1
-fi
+# Available Python interpreter
+for cmd in ${POSSIBLE_PYTHONS[@]}; do
```
Hundao: ${POSSIBLE_PYTHONS[@]} needs quoting ("${POSSIBLE_PYTHONS[@]}") otherwise "py -3" gets word-split into py and -3.

scripts/setup-python.sh
mrunalnshah: Yes you are right  [ -z $PYTHON_CMD ] will never be true. I will initialized the PYTHON_CMD = "". This will solve the problem and will also work as intended.

scripts/setup-python.sh
mrunalnshah: I have changed the variable name of PYTHON_VERSION to REQUIRED_PYTHON_VERSION. The intent of this is to generalize the code so the developers need to change the minimum requirement at one place. Thanks for highlighting the issue.

scripts/setup-python.sh
mrunalnshah: Interestingly, I the code works without quotes on Windows. I have checked and tested py -3, py, python, python3 individually with 3.12 and 3.9 versions of python. Although, adding the quotes also doesn't change the execution. I am adding the quotes.

scripts/setup-python.sh
mrunalnshah: I am unsure of this. Is this something related to security issues? I have tested the code again with python 3.12 and 3.9. I have tested the python commands individually: python3, python, py -3, py. The code is working as intended. 

It would be helpful if you could elaborate on the issue.

My Takeaway:
I am removing eval altogether by just using py in POSSIBLE_PYTHONS=("python3" "python" "py") and later will add suffix during the command call specifically for windows. 

Why?
python3, python -> MacOS, Linux, and Windows
py -> Windows

bryanadenhq (changes requested):
Thanks for the improvement. A couple of things:                                                                                                                                                              
                                                                                                                                                                                                               
  1. CURRENT_CMD="py -3" then $CURRENT_CMD -c "..." relies on word splitting — this works but is fragile. Consider using an array instead.                                                                     
  2. The pip check after the loop is now dead code since the loop already validates pip. Can be removed.                                                                                                       
  3. &> /dev/null 2>&1 — pick one or the other, both together is redundant.

mrunalnshah:
- Implemented Array and refactored entire code. PYTHON_CMD is changed at all places in the entire code.
- Redundant code is removed, design changed a bit for user understanding. (See Screenshots)
- Using `2>&1` as standard. Fix the mis-match in standard code writing.

Screenshots:

[1] Using Array to loop through all interpreters to find the best python and its corresponding pip
image

[2] Using `python3` (3.12) with no pip
image

[3] Using `py` (3.12) with pip installed
image

[4] Using `python` (3.12)  with pip installed
image

[5] Using Python 3.9 
image

mrunalnshah (commented):
All requests by @Hundao and @bryanadenhq are addressed.

mrunalnshah:
I have removed conflict, and kept my code because of three main reasons:
1. The current conflicted code was more static code than dynamic and robust because when we want to add more python interpreters like 3.11, 3.15 (in future), we have to add if-else statements, while my code requires to add in the Array and everything works. 
2. My code checks interpreter and its pip version to find the best possible version for us.
3. I have made a more robust way to threshold the required python. This makes it easy to change required python version.

If any more changes are required, do let me know. 

Thank you!

mrunalnshah:
I am starting to work to add better scripts from conflicted code:
1. add override python functionality with validation python version and pip
2. fixing messages to make it more intuitive
3. add python3.12, and python3.11 in array

PR #1424, opened 2026-01-27
Fixes #1430 Fix: windows security vulnerability and test failures

Tahir-yamin:
## Description
This PR addresses critical issues affecting Windows users and the reliability of the test suite. It fixes a potential sandbox escape vulnerability, resolves CI failures due to incorrect test assumptions, and enables cross-platform testing.

### 1. 🚨 Security Fix: Windows Path Traversal
- **Issue**: [get_secure_path](file:///d:/my-dev-knowledge-base/aden-hive/tools/src/aden_tools/tools/file_system_toolkits/security.py#6-29) in [security.py](file:///d:/my-dev-knowledge-base/hive/tools/tests/tools/test_security.py) failed to strip the drive/root from absolute paths on Windows (e.g., `/etc/passwd` or `C:\...`) because `lstrip(os.sep)` only checked for `\` on Windows, missing `/`.
- **Fix**: Updated logic to strip both `os.sep` and `os.altsep` to prevent absolute paths from being joined incorrectly, avoiding sandbox escape.

### 2. 🐛 Fix: Broken Credential Tests
- **Issue**: `test_credentials.py` asserted that `anthropic` and other credential specs had `required=True`. However, `llm.py` correctly defines them as `required=False` (optional). This caused `pytest` to fail for any fresh clone without keys.
- **Fix**: Updated tests to match the actual flexible design. Validation logic is now tested using custom mock specs instead of relying on mutable production specs.

### 3. 🛠️ Fix: Cross-Platform Test Compatibility
- **Issue**: `test_file_system_toolkits.py` used `subprocess.run` with Unix-only commands (e.g., `ls`, `tr`), causing 10+ failures on Windows.
- **Fix**: Refactored tests to use `unittest.mock` for `subprocess.run`. This simulates command output, making the tests faster, safer, and truly platform-agnostic.

### 4. ✅ Enhancement: Mocked Tool Testing
- Added a "happy path" mock test for `web_search_tool` to allow verifying tool logic without requiring real API keys in the CI environment.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Security patch
- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Verified?
Ran full test suite on Windows 11 with Python 3.13.

```bash
pytest tools/tests/ -v
```
**Result**: 205 passed, 0 failed. (previously 7 failed, 198 passed)

- [x] `tools/tests/test_credentials.py` - PASSED
- [x] `tools/tests/tools/test_file_system_toolkits.py` - PASSED
- [x] `tools/tests/tools/test_web_search_tool.py` - PASSED
- [x] `tools/tests/tools/test_security.py` - PASSED

### DIFF
--- a/tools/src/aden_tools/tools/file_system_toolkits/security.py
+++ b/tools/src/aden_tools/tools/file_system_toolkits/security.py
@@ -15,7 +15,7 @@ def get_secure_path(path: str, workspace_id: str, agent_id: str, session_id: str
     # Resolve absolute path
     if os.path.isabs(path):
         # Treat absolute paths as relative to the session root if they start with /
-        rel_path = path.lstrip(os.sep)
+        rel_path = path.lstrip(os.sep + (os.altsep or ""))
         final_path = os.path.abspath(os.path.join(session_dir, rel_path))
     else:
         final_path = os.path.abspath(os.path.join(session_dir, path))
--- a/tools/tests/test_credentials.py
+++ b/tools/tests/test_credentials.py
@@ -318,8 +318,9 @@ class TestCredentialSpecs:
         assert spec.tools == []
         assert "llm_generate" in spec.node_types
         assert "llm_tool_use" in spec.node_types
-        assert spec.required is True
-        assert spec.startup_required is True
+        # Anthropic is optional in llm.py
+        assert spec.required is False
+        assert spec.startup_required is False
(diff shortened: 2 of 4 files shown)


Tahir-yamin:
Fixes #1430 Fix: windows security vulnerability and test failures

Issue #964, opened 2026-01-27
[Bug]:  setup-python.sh fails on modern Linux/WSL (PEP 668 error)


Turki-Ta:
## Describe the Bug

The `scripts/setup-python.sh` script attempts to install packages globally using `pip`. On modern Linux distributions (like Ubuntu 24.04) and WSL, this triggers a PEP 668 "Externally Managed Environment" error and causes the setup to fail immediately.

## To Reproduce

1. Clone the repository on a fresh Ubuntu 24.04 or WSL (Windows Subsystem for Linux) environment.
2. Run the setup command: `./scripts/setup-python.sh`
3. See error regarding "externally-managed-environment".

## Expected Behavior

The setup script should detect if it is running in a system-managed environment. If so, it should automatically create and activate a local virtual environment (`.venv`) instead of attempting a global install, or provide a clearer prompt to the user to do so.

## Environment

- OS: Ubuntu 24.04 LTS (WSL2 on Windows 11)
- Python version: 3.12
- Docker version (if applicable): N/A

## Logs

Relevant log output:

$ ./scripts/setup-python.sh ... Upgrading pip, setuptools, and wheel... error: externally-managed-environment

× This environment is externally managed ╰─> To install Python packages system-wide, try apt install python3-xyz, where xyz is the package you are trying to install.

If you wish to install a non-Debian-packaged Python package,
create a virtual environment using python3 -m venv path/to/venv.
... Error: Failed to upgrade pip. Please check your python/venv configuration.

## Additional Context

I have already implemented a fix locally that modifies `setup-python.sh` to automatically  detect this state and create/activate a `.venv` directory if one does not exist. I will submit a Pull Request shortly to address this.

Turki-Ta:
I'd like to take this on! (Note: I have already submitted PR #980 with the fix, but it was auto-closed pending assignment per the new workflow in #472. I am standing by for the assignment badge to re-open it.)

Premsenareddy:
Hi,
Thanks for raising this — I can confirm the same issue occurs on macOS with Homebrew-managed Python.

I reproduced this on macOS (Sonoma) where `scripts/setup-python.sh`:
- ignores an existing project venv (Python 3.12)
- invokes system Python (3.13 via Homebrew)
- fails with the same PEP 668 `externally-managed-environment` error when upgrading pip

While I haven’t tested on Linux/WSL directly, the failure mode and root cause appear consistent with what’s described here.

[...]

PR #1504, opened 2026-01-27
docs(setup): clarify virtualenv requirement before running setup script

Premsenareddy:
### Summary

This PR adds a small upfront note to clarify that a Python virtual environment should be created and activated **before** running the setup script on modern systems.

This is related to setup issues caused by **PEP 668 (externally managed Python environments)**, which commonly affect macOS (Homebrew Python), Ubuntu 22.04+, and WSL. While this guidance already exists in the Troubleshooting section, new users often encounter the error during initial setup.

This change keeps the existing documentation intact and surfaces the guidance earlier to improve the first-time setup experience.

Related to #964

### Scope

- Documentation-only change  
- No script or behavior changes

### DIFF
--- a/ENVIRONMENT_SETUP.md
+++ b/ENVIRONMENT_SETUP.md
@@ -2,6 +2,21 @@
 
 Complete setup guide for building and running goal-driven agents with the Aden Agent Framework.
 
+> ⚠️ **Important: Virtual Environment Required on Modern Systems**
+>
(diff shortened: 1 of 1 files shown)


Premsenareddy:
Linked to #964 as requested by the contribution workflow.
This PR provides documentation guidance to prevent the PEP 668 setup failure.

PR #1590, opened 2026-01-28
Fix: _persist_scope() overwriting state with empty dict (#1423)

### LINKED ISSUE #1423: [Bug]: _persist_scope() may overwrite existing state with empty dict when scope doesn't exist in memory
### Describe the bug

In `SharedStateManager._persist_scope()`, when persisting stream or execution state, if the scope ID doesn't exist in the in-memory dictionary, it gets an empty dict `{}` via `.get(id, {})`. This empty dict is then saved to disk, **overwriting any existing persisted state** for that scope.

**Current code (lines 434-459 in `shared_state.py`):**

```python
async def _persist_scope(self, scope: str, id: str) -> None:
    """
    Persist a state scope to disk asynchronously.
    """
    if not self._storage:
        return

    try:
        # Get current state for the scope
        if scope == "global":
            state_data = self._global_state.copy()
        elif scope == "stream":
            state_data = self._stream_state.get(id, {}).copy()  # ❌ Returns {} if not in dict
        elif scope == "execution":
            state_data = self._execution_state.get(id, {}).copy()  # ❌ Returns {} if not in dict
        else:
            return

        # Save to disk (uses batch writer for performance)
        await self._storage.save_state(scope, id, state_data, immediate=False)  # ❌ Saves empty dict
    except Exception as e:
        logger.warning(f"Failed to persist {scope} state for {id}: {e}")
```

### Expected behavior

`_persist_scope()` should:
- Only persist state if the scope actually exists in memory
- Not overwrite existing disk state with an empty dict
- Either skip persistence if scope doesn't exist, or merge with existing disk state

### Actual behavior

When `_persist_scope()` is called for a stream/execution that:
- Exists on disk (from previous run)
- But doesn't exist in the current in-memory `_stream_state` or `_execution_state` dict

The method will:
1. Get an empty dict `{}` via `.get(id, {})`
2. Save this empty dict to disk
3. **Overwrite the existing persisted state**, losing all previous data

### Steps to Reproduce

1. Create a `SharedStateManager` with `ConcurrentStorage`
2. Write state to a stream scope: `await memory.write("key", "value", scope=StateScope.STREAM)`
3. Wait for persistence to complete
4. Create a **new** `SharedStateManager` instance (simulating restart)
5. The stream state is not loaded into memory (lazy loading - only loads on read)
6. If `_persist_scope()` is called for this stream (e.g., during cleanup or explicit persistence), it will:
   - Get empty dict `{}` from `_stream_state.get(stream_id, {})`
   - Save empty dict to disk
   - Overwrite the existing persisted state

**Note:** This is more likely to occur if:
- `_persist_scope()` is called explicitly for a scope that hasn't been loaded
- Cleanup code calls `_persist_scope()` for scopes that were never accessed in memory
- State is persisted before lazy loading occurs

### Impact

- **Severity**: Medium (data loss, but only affects specific scenarios)
- **Frequency**: Low (requires specific timing: persistence called before lazy load)
- **User Impact**: Loss of persisted state for streams/executions that aren't currently in memory

### Proposed Fix

**Option 1: Skip persistence if scope doesn't exist in memory (recommended)**

```python
async def _persist_scope(self, scope: str, id: str) -> None:
    """Persist a state scope to disk asynchronously."""
    if not self._storage:
        return

    try:
        # Get current state for the scope
        if scope == "global":
            state_data = self._global_state.copy()
        elif scope == "stream":
            if id not in self._stream_state:
                # Scope doesn't exist in memory - don't overwrite disk state
                logger.debug(f"Skipping persistence for {scope}:{id} - not in memory")
                return
            state_data = self._stream_state[id].copy()
        elif scope == "execution":
            if id not in self._execution_state:
                # Scope doesn't exist in memory - don't overwrite disk state
                logger.debug(f"Skipping persistence for {scope}:{id} - not in memory")
                return
            state_data = self._execution_state[id].copy()
        else:
            return

        # Only persist if we have actual state data
        if state_data:  # Additional safety check
            await self._storage.save_state(scope, id, state_data, immediate=False)
    except Exception as e:
        logger.warning(f"Failed to persist {scope} state for {id}: {e}")
```

**Option 2: Merge with existing disk state before saving**

```python
async def _persist_scope(self, scope: str, id: str) -> None:
    """Persist a state scope to disk asynchronously."""
    if not self._storage:
        return

    try:
        # Load existing state from disk first
        existing_state = await self._storage.load_state(scope, id) or {}
        
        # Get current state for the scope
        if scope == "global":
            current_state = self._global_state.copy()
        elif scope == "stream":
            current_state = self._stream_state.get(id, {}).copy()
        elif scope == "execution":
            current_state = self._execution_state.get(id, {}).copy()
        else:
            return

        # Merge: memory state takes precedence, but preserve disk state for keys not in memory
        merged_state = {**existing_state, **current_state}
        
        # Only save if there's actual data
        if merged_state:
            await self._storage.save_state(scope, id, merged_state, immediate=False)
    except Exception as e:
        logger.warning(f"Failed to persist {scope} state for {id}: {e}")
```

**Option 3: Only persist scopes that have been modified (track dirty state)**

Add a `_dirty_scopes` set to track which scopes have been modified, and only persist those.

### Additional Context

**Affected Components:**
- `core/framework/runtime/shared_state.py`: `SharedStateManager._persist_scope()`
- Any code that calls `_persist_scope()` for scopes that may not be in memory

**When is `_persist_scope()` called?**
- Automatically after writes via `_write_direct()` → `_persist_scope()`
- During cleanup operations
- Explicitly by callers (if any)

**Related Code:**
- `_ensure_scope_loaded()` (lines 383-432): Lazy loads state from disk
- `_write_direct()` calls `_persist_scope()` after writes
- `StreamMemory.write()` triggers persistence

**Why this matters:**
- With lazy loading, state may exist on disk but not in memory
- If persistence is triggered before the state is loaded, it will overwrite with empty dict
- This breaks the persistence guarantee: "state survives restarts"

This bug is subtle and may not manifest in normal usage (where state is loaded before persistence), but could cause data loss in edge cases or during cleanup operations.

MBGIRISH:
Fixes #1423

## Problem
`_persist_scope()` was overwriting existing persisted state with empty dictionaries when a scope didn't exist in memory (lazy loading scenario).

## Solution
- Skip persistence if scope doesn't exist in memory
- Prevents overwriting existing disk state with empty dictionaries
- Added `save_state()` and `load_state()` methods to storage backends

## Changes
- `SharedStateManager._persist_scope()`: Added check to skip persistence if scope not in memory
- `FileStorage`: Added `save_state()` and `load_state()` methods
- `ConcurrentStorage`: Added async wrappers for state persistence with file locking

## Testing
- No linting errors
- Backward compatible (storage parameter is optional)

### DIFF
--- a/core/framework/runtime/shared_state.py
+++ b/core/framework/runtime/shared_state.py
@@ -12,7 +12,10 @@ import logging
 import time
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from framework.storage.concurrent import ConcurrentStorage
 
 logger = logging.getLogger(__name__)
 
@@ -75,7 +78,13 @@ class SharedStateManager:
         value = await memory.read("customer_id")
     """
 
-    def __init__(self):
+    def __init__(self, storage: ConcurrentStorage | None = None):
+        """
+        Initialize shared state manager.
+
+        Args:
+            storage: Optional storage backend for persistence
+        """
         # State storage at each level
         self._global_state: dict[str, Any] = {}
         self._stream_state: dict[str, dict[str, Any]] = {}  # stream_id -> {key: value}
@@ -92,6 +101,9 @@ class SharedStateManager:
 
         # Version tracking
         self._version = 0
+
+        # Storage backend for persistence
+        self._storage: ConcurrentStorage | None = storage
 
     def create_memory(
         self,
@@ -341,6 +353,48 @@ class SharedStateManager:
     def get_recent_changes(self, limit: int = 10) -> list[StateChange]:
         """Get recent state changes."""
         return self._change_history[-limit:]
+
+    # === PERSISTENCE OPERATIONS ===
+
+    async def _persist_scope(self, scope: str, id: str) -> None:
+        """
+        Persist a state scope to disk asynchronously.
+
+        Only persists state if the scope actually exists in memory to prevent
+        overwriting existing disk state with empty dictionaries.
+
+        Args:
+            scope: State scope ("global", "stream", or "execution")
+            id: Scope identifier (empty string for global)
+        """
+        if not self._storage:
+            return
+
+        try:
+            # Get current state for the scope
+            if scope == "global":
+                state_data = self._global_state.copy()
+            elif scope == "stream":
+                # Skip persistence if scope doesn't exist in memory
+                if id not in self._stream_state:
+                    logger.debug(f"Skipping persistence for {scope}: {id} - not in memory")
+                    return
+                state_data = self._stream_state[id].copy()
+            elif scope == "execution":
+                # Skip persistence if scope doesn't exist in memory
+                if id not in self._execution_state:
+                    logger.debug(f"Skipping persistence for {scope}: {id} - not in memory")
+                    return
+                state_data = self._execution_state[id].copy()
+            else:
+                return
+
+            # Only persist if we have actual state data
+            if state_data:
+                # Save to disk (uses batch writer for performance)
+                await self._storage.save_state(scope, id, state_data, immediate=False)
+        except Exception as e:
+            logger.warning(f"Failed to persist {scope} state for {id}: {e}")
 
 
 class StreamMemory:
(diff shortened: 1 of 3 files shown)


Hundao:
Closing, referenced issue has been closed.

PR #1655, opened 2026-01-28
micro-fix: validate run_id handling in BuilderQuery

### LINKED ISSUE #1622: BuilderQuery: undefined behavior when querying invalid or non-existent run_id
## Problem

While reviewing `BuilderQuery` behavior via tests, I noticed that the behavior for
invalid or non-existent `run_id` values is unclear and inconsistent across query methods.

Examples:
- `get_run_summary(run_id)`
- `get_full_run(run_id)`
- `analyze_failure(run_id)`
- `get_decision_trace(run_id)`

It is not clearly defined whether these methods should:
- return `None`
- raise a specific exception
- or surface a structured error

This makes it difficult to rely on BuilderQuery safely in higher-level workflows.

---

## Expected Behavior

One of the following should be consistently defined and documented:
- Graceful `None` return for missing runs
- OR a well-defined exception (e.g. `RunNotFoundError`)
- OR explicit error metadata in the response

---

## Why this matters

- BuilderQuery is a read/query interface and should be safe against invalid inputs
- Tests currently assume success paths, but edge cases are not specified
- Clear behavior improves reliability for Builder workflows and tooling

---

## Scope

This issue focuses **only** on query-layer behavior.
It does not affect agent execution or runtime logic.

coderpal:
## What this PR does

This PR adds explicit validation for `run_id` in BuilderQuery public methods.

- Empty or invalid `run_id` values now fail fast with clear errors
- Prevents silent `None` returns that make debugging analysis logic difficult
- Adds tests covering invalid `run_id` usage
- No behavior change for valid runs

## Why

Passing an empty or invalid `run_id` currently results in unclear or silent failures.
Since BuilderQuery is a read/query interface, failing fast provides clearer feedback
and avoids subtle bugs during analysis.

## Tests

- Added tests for invalid `run_id` values
- Existing tests remain unchanged and passing

Closes #1622


coderpal:
This is a small, scoped fix with tests added to define expected behavior.
Happy to adjust or extend validation if needed.

PR #1817, opened 2026-01-28
fix(tools): handle optional duckdb dependency and update credential tests

Vikasverma9515:
### Summary
Fixes failing tests in the `tools` package.

### Changes
- Skipped `TestCsvSql` tests in `test_csv_tool.py` availability check for `duckdb`.
- Updated `test_credentials.py` to test abstract "required" logic instead of enforcing specific vendor keys.

### Verification
- [x] Ran `pytest tools` (Passed with 12 skips, 0 failures)
- [x] Ran `pytest core` (Passed)

### Related Issue
Fixes #1680


tools/tests/tools/test_csv_tool.py
```diff
@@ -5,6 +5,10 @@
 
 import pytest
 from fastmcp import FastMCP
+from aden_tools.tools.csv_tool.csv_tool import register_tools
+import importlib.util
+
+duckdb_available = importlib.util.find_spec("duckdb") is not None
 
 from aden_tools.tools.csv_tool.csv_tool import register_tools
```
bryanadenhq: double import

bryanadenhq (changes requested):
everything else looks good other than this double import. Thanks for your contribution!

tools/tests/tools/test_csv_tool.py
Vikasverma9515: Fixed! I've removed the duplicate import. Ready for review.

PR #1819, opened 2026-01-28
ci: add Windows/Linux matrix testing to prevent OS-specific regressions

ankit5841:
## Description

I noticed that the recent `fix/windows-encoding` issue generated 10+ duplicate PRs because the current CI pipeline was only running on Ubuntu, leaving Windows compatibility unchecked.

This PR upgrades the CI workflow to use a **Matrix Strategy**, running tests in parallel on both **Windows and Linux** and across multiple Python versions. This ensures OS-specific regressions (like `UnicodeEncodeError` or pathing issues) are caught automatically before they reach `main`.

## Type of Change

- [ ] Documentation update
- [x] Refactoring (no functional changes, but infrastructure improvement)

## Related Issues

Fixes #1808

## Changes Made

- **Matrix Strategy:** Updated `ci.yml` to run tests on `os: [ubuntu-latest, windows-latest]` and `python-version: ['3.11', '3.12']`.
- **Cross-Platform Scripting:** Replaced the Unix-specific shell script in the `validate` step with a Python one-liner to ensure it runs correctly on Windows.
- **Workflow Triggers:** Updated trigger to allow the CI to run on all branches (for better dev testing) while keeping PRs focused on `main`.

## Testing

- [x] Unit tests pass (`cd core && pytest tests/`) - *Verified via the new GitHub Actions matrix.*
- [x] Manual testing performed - *Verified in forked repository; confirmed that Windows tests correctly failed before the fix and passed after.*

## Checklist

- [x] I have made corresponding changes to the documentation


.github/workflows/ci.yml
```diff
@@ -84,31 +85,5 @@ jobs:
       - name: Validate exported agents
         run: |
           # Check that agent exports have valid structure
-          if [ ! -d "exports" ]; then
```
austin931114, line 107: Could you explain more on the logic change here?

ankit5841: The original validation step used a shell script (if [ ! -d ... ]), which relies on Bash. Since this PR enables windows-latest runners, that script would fail because Windows defaults to PowerShell/CMD.

I replaced it with a Python one-liner so the validation logic is cross-platform and runs identically on both Ubuntu and Windows without needing conditional shell configuration.

Hundao:
Windows matrix is already in CI. Closing, thanks!

PR #1863, opened 2026-01-28
fix(cli): remove invalid default "exports" directory and use current directory

### LINKED ISSUE #1832: [Bug]: CLI default directory "exports" does not exist in current repository
## Describe the Bug

The CLI uses "exports" as the default directory when no directory argument is provided (see cli.py). However, the current repository structure no longer includes an exports/ folder. As a result, CLI commands fail when run without explicitly specifying a directory.

This creates confusion for new users and breaks the default CLI behavior.

## To Reproduce

1. Clone the repository
2. From the project root, run:
         PYTHONPATH=core python -m framework.cli list
4. See error: Directory not found: exports

## Expected Behavior

The CLI should not assume a directory that does not exist. It should either:

- Use a valid existing directory as default (e.g., "."), or
- Require the user to explicitly provide a directory instead of using "exports".

## Screenshots

NA

## Environment

- OS: macOS
- Python version: 3.13

## Additional Context

In cli.py, the default directory is set as:
       default="exports"

This appears to be a leftover from an older repository structure where example agents were stored in an exports/ folder.

Proposed Solutions:

One of the following approaches could resolve the issue:
1. Remove the default value and make the directory argument mandatory
2. Change the default directory to the current directory (".")
3. Update CLI help text and documentation to match the actual repository structure

Bharadwaja04:
## Description

Fixes a CLI bug where the default directory was set to "exports" even though the repository no longer includes an exports/ folder.

Running CLI commands without specifying a directory caused:
           Directory not found: exports
This change updates the default behavior so the CLI searches the current directory (.) instead. This makes the CLI work on a clean clone and aligns it with the current repository structure.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #1832 

## Changes Made

- Replaced hardcoded default "exports" with "." in CLI argument definitions
- Updated CLI help text to reflect the new default directory
- Ensured CLI works correctly on a clean repository clone

## Testing

Manual verification:
        PYTHONPATH=core python -m framework.cli list
Before fix:
❌ Directory not found: exports

After fix:
✅ CLI runs and scans current directory successfully

Also verified agent detection behaviour with valid and invalid agent.json files.

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas

NA


Bharadwaja04:
I updated several setup output lines that referenced exports/ as if example agents already existed in the repository.

On a clean clone, the repo does not contain an exports/ directory. That folder is created only after a user builds/exports an agent. The previous messages could mislead new users into thinking example agents were bundled with the framework.

These changes clarify that ./exports/ is user-generated and align the setup messaging with the CLI default directory update and the current repository structure.

Hundao:
The missing directory case is already handled gracefully (returns "No agents found"). Closing, thanks!

PR #1898, opened 2026-01-28
feat(cli): structured error handling with actionable guidance

### LINKED ISSUE #1337: [Feature]: CLI error messages should include actionable guidance and recovery suggestions
## Problem Statement

When CLI commands fail, error messages are typically raw Python tracebacks or generic error strings without actionable guidance. Users encountering errors during agent validation, execution, or configuration have no indication of what went wrong or how to fix it.

Example of current behavior:
```
KeyError: 'steps'
```

Expected behavior:
```
Error: Agent schema key 'steps' not found.
  Hint: The agent graph uses 'nodes' instead of 'steps'. 
  This may indicate an outdated agent export. Try re-exporting with:
    hive validate my_agent
```

## Proposed Solution

1. Wrap CLI entry points with structured error handlers
2. Map common exceptions to user-friendly messages with remediation steps
3. Use a consistent error output format: `Error: <what happened>` + `Hint: <what to do>`
4. In `--verbose` mode, also show the full traceback

## Implementation Ideas

Create an error handler decorator or context manager for CLI commands:

```python
class CLIError(Exception):
    def __init__(self, message: str, hint: str = "", details: str = ""):
        self.message = message
        self.hint = hint
        self.details = details
```

Map common exceptions:
- `KeyError` in agent loading → schema mismatch guidance
- `FileNotFoundError` → check paths, suggest `hive list`
- `ModuleNotFoundError` → check PYTHONPATH / installation
- `httpx.ConnectError` → check API keys, network

## Additional Context

Professional CLI tools (rustc, cargo, elm) are known for excellent error messages. This would significantly improve developer experience for both new and experienced users.

tjsasakifln:
## Summary

- Introduces a **diagnostic-first error system** for CLI commands, replacing raw Python tracebacks with user-friendly messages and actionable remediation hints
- Implements a `CLIError` exception class with `message`, `hint`, and `details` fields, following the pattern proposed by @Janhvi2909 (Smart Remediation approach)
- Wraps all CLI entry points (`run`, `info`, `validate`, `list`, `dispatch`, `shell`) with a `@cli_error_handler` decorator that maps common exceptions to structured diagnostics
- Adds a last-resort handler at the top-level `main()` to catch anything that escapes command-level handling

### Error mappings included

| Exception | Guidance |
|-----------|----------|
| `KeyError` (e.g. `'steps'`) | Suggests correct schema key (`'nodes'`), recommends `hive validate` |
| `FileNotFoundError` (agent) | Points to `hive list` for discovery |
| `ModuleNotFoundError` (`framework`) | Shows correct `PYTHONPATH` setup |
| `ModuleNotFoundError` (third-party) | Suggests `pip install <module>` |
| `PermissionError` | Advises checking file permissions |
| Connection errors (`httpx`) | Suggests checking API key and network |
| Auth errors (401) | Points to API key configuration |
| Rate limit errors | Suggests waiting or switching model |
| Unknown errors | Recommends `--verbose` and links to issue tracker |

### Example output

**Before:**
```
KeyError: 'steps'
```

**After:**
```
Error: Agent schema key 'steps' not found.
  Hint: The agent graph uses 'nodes' instead of 'steps'.
  This may indicate an outdated agent export. Try re-exporting, then run:
    hive validate <agent_path>
```

## Test plan

- [x] 24 unit tests covering all error paths (CLIError, formatting, resolution, decorator, context manager)
- [x] All existing framework tests pass unchanged
- [x] Ruff lint passes with zero errors
- [ ] Manual verification: run `hive run nonexistent/` and confirm structured error output
- [ ] Manual verification: run with `--verbose` and confirm traceback is shown
- [ ] Review by @Janhvi2909 to validate Smart Remediation alignment

Closes #1337

### DIFF
--- a/core/framework/cli.py
+++ b/core/framework/cli.py
@@ -18,6 +18,7 @@ Testing commands:
 
 import argparse
 import sys
+import traceback
 from pathlib import Path
 
 
@@ -62,6 +63,12 @@ def main():
         default="claude-haiku-4-5-20251001",
         help="Anthropic model to use",
     )
+    parser.add_argument(
+        "--verbose",
+        "-v",
+        action="store_true",
+        help="Show full tracebacks on error",
+    )
 
     subparsers = parser.add_subparsers(dest="command", required=True)
 
@@ -78,7 +85,22 @@ def main():
     args = parser.parse_args()
 
     if hasattr(args, "func"):
-        sys.exit(args.func(args))
+        try:
+            sys.exit(args.func(args))
+        except KeyboardInterrupt:
+            print("\nInterrupted.", file=sys.stderr)
+            sys.exit(130)
+        except Exception as exc:
+            # Last-resort handler for anything that escapes command-level handling
+            from framework.errors import _resolve_error, format_error
+
+            verbose = getattr(args, "verbose", False)
+            cli_err = _resolve_error(exc)
+            cli_err.__cause__ = exc
+            print(format_error(cli_err, verbose=verbose), file=sys.stderr)
+            if verbose:
+                traceback.print_exc(file=sys.stderr)
+            sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/core/framework/errors.py
+++ b/core/framework/errors.py
@@ -199,23 +199,40 @@ def format_error(error: CLIError, verbose: bool = False) -> str:
 
 
 def _resolve_error(exc: Exception) -> CLIError:
-    """Attempt to map a raw exception to a CLIError with actionable guidance."""
+    """Attempt to map a raw exception to a CLIError with actionable guidance.
+
+    Walks the exception chain (__cause__ / __context__) so that wrapped
+    exceptions (e.g. ``RuntimeError`` wrapping ``ModuleNotFoundError``)
+    still produce specific, actionable hints.
+    """
     # 1. Already a CLIError — pass through
     if isinstance(exc, CLIError):
         return exc
 
-    # 2. Try registered mappers (exact type match first, then subclass)
-    for exc_type, mapper in _ERROR_MAP:
-        if isinstance(exc, exc_type):
-            result = mapper(exc)
+    # Collect the chain: [exc, cause, cause-of-cause, ...]
+    chain: list[Exception] = []
+    seen: set[int] = set()
+    cur: BaseException | None = exc
+    while cur is not None and id(cur) not in seen and len(chain) < 10:
+        if isinstance(cur, Exception):
+            chain.append(cur)
+        seen.add(id(cur))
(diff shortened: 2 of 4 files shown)


Janhvi2909:
Nice work on this, @tjsasakifln ! You’ve captured the 'Diagnostic-first' vision perfectly, the way you’ve handled the ModuleNotFoundError logic (differentiating between PYTHONPATH and pip) is exactly the kind of 'Smart Remediation' I was hoping for.
The KeyError mapping to the 'nodes' vs 'steps' hint is also a great touch. It shows you're thinking about the actual friction points for our users.
Architecture-wise, the CLIError structure is clean and stays true to the design. I’m happy with the 24 tests covering the error paths too.

A few quick thoughts from my side:

- One small polish suggestion: For the ConnectionError hint, should we explicitly remind users to check their HIVE_API_KEY env var, or is that strictly handled in the 401 map?

- Visuals: The separation between the Error and Hint looks very clean. If we're using a library like rich, maybe we can bold the 'Actionable Hint' to make it pop?

- Edge Cases: Great call on the catch-all in main(). Does this also handle nested exceptions gracefully, or should we stick to the primary traceback for those?

- Verbosity: Appreciate you keeping the --verbose flag functional; it's essential for us when debugging the core.

Aside from that, this is exactly what I had in mind. Great job bringing the proposal to life!

@austin931114 as asked by you i had reviewed the PR, The implementation is modular and easy to extend totally brings my Smart Remediation proposal to life,LGTM (Looks Good To Me) once those minor formatting bits are checked!

tjsasakifln:
Thanks for the thorough review @Janhvi2909! Great observations. Here's how I addressed each point:

**1. ConnectionError + API key hint**
Already covered — `_try_map_connection_error` explicitly reminds users to check `ANTHROPIC_API_KEY`, and `_try_map_auth_error` handles the 401 case with the same env var. The project doesn't use a separate `HIVE_API_KEY`, so no changes needed here.

**2. Bold hints with `rich`**
I considered this, but `rich` isn't a current dependency and adding one just for formatting felt heavy. The current plain-text output works cleanly in all terminals. If the project adopts `rich` for other reasons down the road, bolding the hint line would be a one-liner change in `format_error()`.

**3. Nested exceptions**
Great catch — this was a real gap. I just pushed a fix: `_resolve_error` now walks the full `__cause__`/`__context__` chain (capped at depth 10), so a `RuntimeError` wrapping a `ModuleNotFoundError` still produces the specific PYTHONPATH hint instead of the generic fallback. Added 3 tests covering `__cause__`, `__context__`, and circular chain protection. See commit f0199ea.

**4. --verbose flag**
Confirmed working as-is, no changes needed.

All 27 tests passing.

Hundao:
Closing, referenced issue has been closed.

PR #1912, opened 2026-01-28
micro-fix: Initialize dotenv in CLI entry point to ensure env vars load

harihran-del:
## Summary
Fixed a bug where the agent runtime would attempt to initialize configuration settings before loading environment variables from the `.env` file, causing a crash on startup or missing keys.

## Changes
* Added `load_dotenv()` to the top of the `main()` function in `core/framework/cli.py`.
* Added `from dotenv import load_dotenv` import.

## Verification
* Created a local `.env` file with test variables.
* Verified that `cli.py` now successfully reads these variables immediately upon execution.

### DIFF
--- a/core/framework/cli.py
+++ b/core/framework/cli.py
@@ -18,9 +18,11 @@ Testing commands:
 
 import argparse
 import sys
+from dotenv import load_dotenv
 
 
 def main():
+    load_dotenv()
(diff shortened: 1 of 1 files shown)


RichardTang-Aden:
This kind of fix will require an issue

PR #1941, opened 2026-01-28
fix(llm): return parse error to LLM on malformed tool call arguments

sweety-ag:
## Summary

- When `json.loads()` fails on tool call arguments, return a clear error message to the LLM as a tool result instead of silently falling back to `{}`
- Add `logging.warning()` with the raw malformed arguments for debugging
- Add two regression tests verifying the fix and that valid args still work

## Details

**File:** `core/framework/llm/litellm.py:239-242`

**Before (silent fallback):**
```python
except json.JSONDecodeError:
    args = {}  # Tool executes with empty args → fails → LLM retries → loop
```

**After (error feedback):**
```python
except json.JSONDecodeError as e:
    logger.warning(
        "Failed to parse tool arguments for '%s': %s | Raw: %s",
        tool_call.function.name, e, tool_call.function.arguments[:200],
    )
    current_messages.append({
        "role": "tool",
        "tool_call_id": tool_call.id,
        "content": json.dumps({"error": f"Invalid JSON in tool arguments: {e}. Please retry with valid JSON."}),
    })
    continue  # Skip tool execution, let LLM self-correct
```

Since `AnthropicProvider` wraps `LiteLLMProvider`, this fix applies to all LLM providers in the framework.

## Test plan

- [x] All 25 LiteLLM provider tests pass (23 existing + 2 new)
- [x] `test_malformed_args_sends_error_not_empty_dict` — verifies tool executor is NOT called and LLM receives error message
- [x] `test_valid_args_still_execute_normally` — regression test confirming valid JSON still works
- [x] Ruff lint and format checks pass

Closes #1927

### DIFF
--- a/core/framework/llm/litellm.py
+++ b/core/framework/llm/litellm.py
@@ -8,6 +8,7 @@ See: https://docs.litellm.ai/docs/providers
 """
 
 import json
+import logging
 from collections.abc import Callable
 from typing import Any
 
@@ -17,6 +18,8 @@ except ImportError:
     litellm = None  # type: ignore[assignment]
 
 from framework.llm.provider import LLMProvider, LLMResponse, Tool, ToolResult, ToolUse
+
+logger = logging.getLogger(__name__)
 
 
 class LiteLLMProvider(LLMProvider):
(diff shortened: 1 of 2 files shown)


Hundao:
Closing, referenced issue has been closed. This fix is already on main.

PR #2123, opened 2026-01-29
docs: fix Claude Code installation and PATH issues on Windows

dipgarg0011:
### What this PR does

On Windows, following the current README can result in `claude: command not found`
even when Claude Code is installed correctly.

This PR documents the actual working installation and verification steps on Windows,
including common PATH and shell-related pitfalls.

### Details
- Clarifies that Claude Code is not installed by default on Windows
- Adds explicit install + verification steps (`claude --help`)
- Documents why `claude` may fail due to PATH or shell differences
- Notes that `Credit balance too low` can appear even when installation is correct

These issues were encountered while setting up Hive on Windows and are reproducible
with the current documentation.

Tested on: Windows 11 + Git Bash

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,32 @@ Aden is a platform for building, deploying, operating, and adapting AI agents:
 
 ### Prerequisites
 
+### Windows Notes for Claude Code
+
+On Windows, Claude Code is not installed by default and must be installed explicitly.
+
+1. Install Claude Code using the official installer:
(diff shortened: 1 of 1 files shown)


dipgarg0011:
Happy to adjust wording or placement if you'd prefer this live in a different section.

austin931114 (commented):
close for outdated doc

PR #2148, opened 2026-01-29
fix: read file incrementally in view_file to prevent OOM

### LINKED ISSUE #2147: [Bug]: view_file reads entire file into memory before applying max_size limit
## Describe the Bug

The `view_file` tool reads the entire file into memory before checking if it exceeds `max_size`. This defeats the purpose of having a `max_size` parameter and can cause out-of-memory errors when attempting to view large files.

## Location

**File:** `tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py`  
**Lines:** 59-64

## Current Behavior

```python
with open(secure_path, encoding=encoding) as f:
    content = f.read()  # Reads ENTIRE file into memory

if len(content.encode(encoding)) > max_size:  # Then checks size
    content = content[:max_size]  # Truncates after the fact
```

If a user tries to view a 10GB file with `max_size=1MB`, the entire 10GB file is loaded into memory first, potentially crashing the process.

## Expected Behavior

The file should be read incrementally, stopping once `max_size` bytes have been read:

```python
with open(secure_path, encoding=encoding) as f:
    content = f.read(max_size)  # Read only up to max_size
    
    # Check if there's more content
    if f.read(1):
        content += "\n\n[... Content truncated due to size limit ...]"
```

## Additional Issues

1. The truncation at line 63 uses character count (`content[:max_size]`), not byte count, which could differ for multi-byte encodings like UTF-8
2. Would be helpful to check file size before opening to warn users about large files

## Environment

- OS: All platforms
- Python: 3.11+

## Proposed Fix

1. Use `f.read(max_size)` instead of `f.read()` to limit memory usage
2. Detect truncation by attempting to read one more character
3. Optionally check `os.path.getsize()` before reading for better error messaging

Happy to submit a PR for this!

himanshu748:
## Summary

- Fixes memory safety issue where `view_file` loaded entire files before checking `max_size`
- Now uses incremental reading to limit memory usage to `max_size` bytes
- Adds `file_size_bytes` and `truncated` fields to response for transparency

## Problem

Previously, if a user tried to view a 10GB file with `max_size=1MB`, the tool would:
1. Load entire 10GB into memory
2. Then check if size > max_size
3. Then truncate to 1MB

This defeats the purpose of `max_size` and can crash the process with OOM.

## Solution

```python
# Before: reads ENTIRE file
content = f.read()
if len(content.encode(encoding)) > max_size:
    content = content[:max_size]

# After: reads only up to max_size
content = f.read(max_size)
was_truncated = bool(f.read(1))  # Check if there's more
```

## Changes

**File:** `tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py`

1. Use `f.read(max_size)` instead of `f.read()`
2. Detect truncation by reading one more character
3. Add `file_size_bytes` to response (actual file size on disk)
4. Add `truncated` boolean to response

## Test Plan

- [x] Verified small files work correctly (no truncation)
- [x] Verified large files are truncated without loading entire file
- [x] Verified response includes new fields

Fixes #2147

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,4 +1,4 @@
-Product Roadmap
+# Product Roadmap
 
 Aden Agent Framework aims to help developers build outcome oriented, self-adaptive agents. Please find our roadmap here
 
--- a/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py
+++ b/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py
@@ -56,11 +56,17 @@ def register_tools(mcp: FastMCP) -> None:
             if not os.path.isfile(secure_path):
                 return {"error": f"Path is not a file: {path}"}
 
-            with open(secure_path, encoding=encoding) as f:
-                content = f.read()
+            # Check file size before reading to provide helpful feedback
+            file_size = os.path.getsize(secure_path)
 
-            if len(content.encode(encoding)) > max_size:
-                content = content[:max_size]
+            # Read only up to max_size to prevent OOM on large files
+            with open(secure_path, encoding=encoding) as f:
+                content = f.read(max_size)
+
+                # Check if file was truncated by trying to read one more char
+                was_truncated = bool(f.read(1))
+
+            if was_truncated:
                 content += "\n\n[... Content truncated due to size limit ...]"
 
             return {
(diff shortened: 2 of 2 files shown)


himanshu748:
## CI Status

### check-requirements ❌
The linked issue (#2147) needs an assignee. I've commented on the issue requesting assignment. I'm not a collaborator on this repo, so I cannot self-assign.

### Lint Python ✅
Pass

### Test Python (Ubuntu) ✅
Pass

### Test Python (Windows) ⏳
Pending

PR #2221, opened 2026-01-29
fix(output-cleaner): handle empty LLM responses + invalid output types

### LINKED ISSUE #2201: Robustness Fixes: OutputCleaner crashes on empty LLM responses and invalid output types
## Describe the Bug

OutputCleaner assumes that the LLM always returns a non-empty string in response.content. In practice, LLM providers may return None, an empty string, or whitespace-only content due to provider errors, timeouts, or malformed responses.

When this happens:

- Calling .strip() on None raises a TypeError

- Empty or whitespace-only responses proceed further into the cleaning pipeline, leading to misleading warnings or invalid downstream state instead of failing cleanly

Additionally, validate_output() assumes the cleaned output is always a dictionary. If a non-dict value (e.g., None or a string) reaches validation, the method crashes immediately when accessing expected keys.

These cases are not handled defensively, causing hard crashes or silent incorrect behavior.

## To Reproduce

Trigger OutputCleaner with an LLM provider that returns None, an empty string, or whitespace-only content

Observe that:

- A TypeError occurs when .strip() is called on None, or

- The cleaner proceeds with invalid data

- If a non-dict output reaches validate_output(), observe a crash when accessing expected keys

## Expected Behavior

- OutputCleaner should handle None, empty, or whitespace-only LLM responses gracefully

- Validation should fail cleanly with a clear error message when output is not a dictionary

- The system should either fall back safely or raise a meaningful error instead of crashing

## Environment

OS: Any

Python version: Any

Docker version: Not applicable

Logs
TypeError: 'NoneType' object has no attribute 'strip'

(Logs may vary depending on provider behavior.)

## Additional Context

This is a robustness and reliability issue that becomes more likely under transient LLM failures or degraded provider responses. Because the failure occurs in utility cleaning and validation paths, it can crash entire graph executions or lead to silent downstream corruption.

Jbheemeswar:
Summary
- Prevent OutputCleaner crashes when LLM returns None/empty/whitespace-only content.
- Make validate_output() fail gracefully when output is not a dict.
- Add regression tests for these cases.

Testing
- PYTHONPATH=core pytest core/framework/graph/tests/test_output_cleaner_robustness.py -q

Closes #2201

### DIFF
--- a/core/framework/graph/output_cleaner.py
+++ b/core/framework/graph/output_cleaner.py
@@ -132,7 +132,7 @@ class OutputCleaner:
 
     def validate_output(
         self,
-        output: dict[str, Any],
+        output: Any,
         source_node_id: str,
         target_node_spec: Any,  # NodeSpec
     ) -> ValidationResult:
@@ -144,6 +144,15 @@ class OutputCleaner:
         """
         errors = []
         warnings = []
+        
+        # Defensive: output must be a dict
+        if not isinstance(output, dict):
+            return ValidationResult(
+                valid=False,
+                errors=[f"Output must be dict, got {type(output).__name__}"],
+                warnings=[],
+        )
+
 
         # Check 1: Required input keys present
         for key in target_node_spec.input_keys:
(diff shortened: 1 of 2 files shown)


Jbheemeswar:
Hi maintainers 
This PR was auto-closed due to workflow #472 because I’m not assigned to the linked issue (#2201). I don’t have permission to self-assign.
I’ve requested assignment in #2201. Could a maintainer please assign @Jbheemeswar to #2201 so I can re-open this PR?
Thanks!

Issue #2222, opened 2026-01-29
Decision invariants are implicit and not enforced at runtime


vsohani28:
Decision is the atomic unit of learning and observability in Hive.
While the Decision schema documents important assumptions these invariants are currently implicit
and not enforced at runtime.

This allows invalid or inconsistent Decision objects to be recorded,
which can silently corrupt decision traces and negatively impact
Builder analysis and self-improvement.

:Expected Behavior
Invalid Decisions should fail fast before being recorded, so corrupted
decision traces never enter the system.

: Sug Direction
Introduce explicit invariant validation for Decision objects and enforce
it at the runtime boundary before a Decision is recorded.

vsohani28:
I’m happy to take this on.  
Could you please assign this issue to me so I can link it to the PR?

PR #2230, opened 2026-01-29
Validate decision data before recording it

vsohani28:
This PR adds a small validation step before a Decision is recorded.

While working with the runtime, I noticed that Decision objects rely on a
few assumptions (for example, having at least one option and a valid
chosen option), but these weren’t being checked anywhere.

This change validates those assumptions right before a Decision is added
to a run. That way, invalid decisions fail early instead of silently
ending up in decision traces.

The goal is to keep decision data clean and make issues easier to catch
during development.

Closes (#2222)


vsohani28:
This PR is linked to #2222.  
I don’t have permission to self-assign, so I’ve requested assignment on the issue.

vsohani28:
@RichardTang-Aden as i did not get any issue assigned i did some changes related to [enforce-decision-invariants] so could you please look into this or assign me the issue so that i can look into it

PR #2270, opened 2026-01-29
Fix retry state leakage by isolating execution context across retries

### LINKED ISSUE #1274: Retries re-execute nodes without isolating or rolling back shared execution state
### Problem
When a node execution fails and is retried, the retry reuses the same shared execution context.
Any mutations performed by the failed attempt are not rolled back before retrying.

As a result, retries are not idempotent and can compound state across attempts.

### Impact
- Shared context may be partially mutated multiple times
- Counters, memory, or side effects can be duplicated
- Downstream nodes may observe corrupted state
- Breaks reproducibility and correctness of agent execution

### Affected Areas
- Graph executor retry logic
- Shared execution context handling
- Self-healing / retry paths

### Suggested Direction
- Snapshot and restore context per retry attempt, or
- Enforce immutability / copy-on-write for retryable state, or
- Clearly document retries as non-idempotent (if intended)

Madhav-150:
Description

Fixes non-idempotent retry behavior in the graph executor where failed node retries could observe or reuse mutated shared state from previous attempts.

This change introduces an executor-level source-of-truth pattern, ensuring that each retry attempt starts from a clean execution state, and that state mutations are only committed after successful node execution.

Type of Change

 Yes — Bug fix (non-breaking change that fixes an issue)

 No — New feature (non-breaking change that adds functionality)

 No — Breaking change (fix or feature that would cause existing functionality to not work as expected)

 No — Documentation update

 No — Refactoring (no functional changes)

Related Issues

Fixes #1274

Changes Made

Introduced a committed execution state inside GraphExecutor that acts as the single source of truth

Ensured every retry attempt rehydrates a fresh SharedMemory instance from the committed state

Prevented failed retry attempts from mutating shared execution context

Committed memory updates only on successful node execution

Added a regression test that fails on main and passes with this fix

Testing

 Yes — Unit tests pass
(python -m pytest core/tests/test_retry_state_isolation.py)

 No — Lint passes (cd core && ruff check .)

 No — Manual testing performed

New Test Added

core/tests/test_retry_state_isolation.py
Verifies that memory mutations from a failed node attempt do not leak into subsequent retries.

Checklist

 Yes — My code follows the project's style guidelines

 Yes — I have performed a self-review of my code

 Yes — I have commented my code where needed

 No — I have made corresponding changes to the documentation

 Yes — My changes generate no new warnings

 Yes — I have added tests that prove my fix is effective

 Yes — New and existing unit tests pass locally with my changes

Screenshots (if applicable)

Not applicable.

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -1,15 +1,5 @@
-"""
-Graph Executor - Runs agent graphs.
-
-The executor:
-1. Takes a GraphSpec and Goal
-2. Initializes shared memory
-3. Executes nodes following edges
-4. Records all decisions to Runtime
-5. Returns the final result
-"""
-
 import asyncio
+import copy
 import logging
 from collections.abc import Callable
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


TheEleventhAvatar:
This is a solid fix — the executor-level source-of-truth model cleanly addresses the non-idempotent retry issue. The separation between committed execution state and attempt-local state makes the retry semantics much easier to reason about, especially under partial failure.

The regression test is particularly strong — it encodes the exact failure mode (state leakage across retries) rather than just asserting final outputs, which gives confidence this won’t regress subtly later.

One question for clarification: is the “commit-on-success” boundary intentionally scoped at the node level rather than the full graph execution? I’m assuming this is to preserve incremental progress while still guaranteeing retry isolation, but it might be worth a brief comment in GraphExecutor to make that design choice explicit for future readers.

Overall, this feels like the right balance between correctness and minimal surface-area change. Nice work.

Madhav-150:
Thanks a lot for the detailed review and I really appreciate the thoughtful feedback.

Yes, the commit-on-success boundary is intentionally scoped at the node level rather than the full graph execution. The goal was to preserve incremental progress across the graph while still guaranteeing that each retry attempt observes a clean, deterministic state. By committing only after a node succeeds, I avoid redoing prior successful work and keep retries isolated to the failing unit of execution.

That tradeoff felt like the right balance between correctness and practical execution semantics, especially for longer graphs where restarting from scratch would be unnecessarily expensive.

Good call on documentation, I agree that making this design choice explicit would help future readers. I will add a brief comment in GraphExecutor explaining why the commit boundary is node-scoped if that’s helpful.

Thanks again for the kind words and for taking the time to reason through the design.

Issue #1982, opened 2026-01-28
Remove hardcoded Anthropic dependency from LLMJudge


llamasearchai:
The `LLMJudge` class in `core/framework/testing/llm_judge.py` has a hardcoded dependency on the `anthropic` package.

Specifically:
1. `_get_client` imports `anthropic` and raises dependency error if missing.
2. `evaluate` falls back to `_get_client` which forces Anthropic usage.
3. `_get_fallback_provider` hardcodes `AnthropicProvider` instantiation.

This breaks provider agnosticism. It should utilize `LiteLLMProvider` or allow pure dependency injection without requiring the `anthropic` package to be installed.

RichardTang-Aden:
https://github.com/adenhq/hive/pull/2342 provided a solution for this

PR #2342, opened 2026-01-29
Feat: Replace hardcoded Anthropic fallback with LiteLLM provider and changes to make file

### LINKED ISSUE #2339: LLMJudge provider-agnostic fallback + Makefile setup-venv target
LLMJudge uses hardcoded AnthropicProvider fallback, breaking semantic evaluation for non-Anthropic LLM setups. Replace with LiteLLM for provider-agnostic support. Add Makefile setup-venv target for easy virtual environment setup with dependencies.

omkar806:
Fixes #2339 , #1982

## Description

This PR makes `LLMJudge` provider-agnostic by replacing the hardcoded `AnthropicProvider` fallback with a LiteLLM-based provider. This enables semantic evaluation to work with any LiteLLM-supported provider (OpenAI, Gemini, DeepSeek, Ollama, etc.) without requiring Anthropic API keys, aligning with the framework's multi-provider architecture.

Additionally, adds a `setup-venv` Makefile target for easy virtual environment setup with all dependencies.

## Type of Change

## Related Issues

Fixes the issue where `LLMJudge` would fail for users with non-Anthropic LLM setups, breaking semantic evaluation functionality despite the framework claiming provider-agnostic support.

## Changes Made

- **LLMJudge provider-agnostic fallback**: Replaced hardcoded `AnthropicProvider` with `LiteLLMProvider` in `_get_fallback_provider()`
- **Environment variable support**: Added `LLM_JUDGE_MODEL` env var to configure the judge model (e.g., `gpt-4o-mini`, `ollama/llama3`, `gemini/gemini-1.5-flash`)
- **Code-level model parameter**: Added optional `model` parameter to `LLMJudge.__init__()` for programmatic model selection
- **Backward compatibility**: Default fallback model remains `claude-3-haiku-20240307` to preserve existing behavior
- **Test updates**: Updated test suite to reflect new fallback behavior while maintaining backward compatibility for legacy mocked client path
- **Makefile enhancement**: Added `setup-venv` target that creates venv, installs dependencies from `requirements.txt`, installs packages in editable mode, and sets `PYTHONPATH`

## Testing

- [x] Unit tests pass (`cd core && pytest tests/`) - **372 passed, 1 skipped**
- [x] Manual testing performed - Verified `LLMJudge` works with different models via `model` parameter and `LLM_JUDGE_MODEL` env var

**Test Results:**
image

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,3 +24,17 @@ test: ## Run all tests
 install-hooks: ## Install pre-commit hooks
 	pip install pre-commit
 	pre-commit install
+
+setup-venv: ## Create venv, install deps, and set PYTHONPATH (VENV=./.venv)
+	@VENV=$${VENV:-.venv}; \
+	python3 -m venv $$VENV; \
+	$$VENV/bin/pip install --upgrade pip; \
+	$$VENV/bin/pip install -r core/requirements.txt; \
+	$$VENV/bin/pip install -e core -e tools; \
+	PYTHONPATH_LINE='export PYTHONPATH="$(PWD)/core:$(PWD)/exports"'; \
+	ACTIVATE_FILE="$$VENV/bin/activate"; \
+	if ! grep -Fq "$$PYTHONPATH_LINE" "$$ACTIVATE_FILE"; then \
+		printf '\n# Added by make setup-venv\n%s\n' "$$PYTHONPATH_LINE" >> "$$ACTIVATE_FILE"; \
+	fi; \
+	echo "Venv ready: $$VENV"; \
+	echo "Activate with: source $$VENV/bin/activate"
--- a/core/framework/testing/llm_judge.py
+++ b/core/framework/testing/llm_judge.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import json
 import os
 from typing import TYPE_CHECKING, Any
+from framework.llm.litellm import LiteLLMProvider
 
 if TYPE_CHECKING:
     from framework.llm.provider import LLMProvider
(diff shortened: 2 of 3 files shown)


RichardTang-Aden:
I think the PR is legit but the issue was duplicated. re-linking this to #1982

Please also make sure you don't use micro-fix for a PR like this, micro-fix must be under 20 lines changed and has no logical change

omkar806:
> I think the PR is legit but the issue was duplicated. re-linking this to #1982
> 
> Please also make sure you don't use micro-fix for a PR like this, micro-fix must be under 20 lines changed and has no logical change

Sure @RichardTang-Aden  . Actually it was Auto Closing the PR hence added this , as it was closing the PR as it was not assigned to me .

RichardTang-Aden:
## Review Feedback

This PR doesn't fully solve #1982. A few critical issues:

1. **`_get_fallback_provider()` never returns `None` anymore** — it always returns a `LiteLLMProvider`. This makes the entire legacy `_get_client()` path and the `RuntimeError` raise in `evaluate()` dead code. If you're keeping that code, justify it. If not, delete it.

2. **Removes API key auto-detection, which is a regression.** The old code checked `OPENAI_API_KEY` then `ANTHROPIC_API_KEY` and picked the right provider. Now it blindly defaults to `claude-3-haiku-20240307` via LiteLLM — so a user with only an OpenAI key and no `LLM_JUDGE_MODEL` set will get a runtime failure. That's worse than before.

3. **Default model still requires Anthropic credentials.** The whole point of #1982 is removing the Anthropic dependency, but the default path still calls an Anthropic model. You just moved the dependency from a direct import to a LiteLLM routing.

4. **Top-level hard import of `LiteLLMProvider`** replaces the old lazy imports. If `litellm` isn't installed, `llm_judge.py` now fails on import rather than at the point of use.

5. **Tests mock the changed method itself** (`_get_fallback_provider`), so they don't validate the actual LiteLLM integration. `test_legacy_client_used_only_when_mocked` tests a code path that can never execute in production.

6. **Makefile `setup-venv` is unrelated** — should be a separate PR.

To actually fix #1982: keep the API-key auto-detection but route through `LiteLLMProvider`, remove the dead `_get_client()` legacy path, use lazy imports, and write tests that exercise the real fallback logic rather than mocking it away.

PR #2419, opened 2026-01-29
micro-fix: move pytest from production dependencies to dev extras

savankansagara1:
### Summary
This is a small configuration-only change to clean up production dependencies.

### What changed
- Moved pytest, pytest-asyncio, and pytest-xdist from
  [project.dependencies] to [project.optional-dependencies] (dev)

### Why
- Prevents test-only dependencies from being installed in production
- Reduces dependency footprint and security surface

### Scope
- < 20 lines changed
- No logic, API, or database changes

### DIFF
--- a/core/pyproject.toml
+++ b/core/pyproject.toml
@@ -11,15 +11,15 @@ dependencies = [
     "litellm>=1.81.0",
     "mcp>=1.0.0",
     "fastmcp>=2.0.0",
-    "pytest>=8.0",
-    "pytest-asyncio>=0.23",
-    "pytest-xdist>=3.0",
 ]
 
 [project.optional-dependencies]
 dev = [
     "ruff>=0.1.0",
     "mypy>=1.0",
+    "pytest>=8.0",
+    "pytest-asyncio>=0.23",
+    "pytest-xdist>=3.0",
 ]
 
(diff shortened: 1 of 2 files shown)


RichardTang-Aden (approved):

RichardTang-Aden:
Can you please check and pull from main? There was a major package manager change (uv) this morning.

Plus, it would also be great if we can make tools/pyproject.toml and core/pyproject.toml format consistent

savankansagara1:
> Can you please check and pull from main? There was a major package manager change (uv) this morning.
> 
> Plus, it would also be great if we can make tools/pyproject.toml and core/pyproject.toml format consistent

yes we can i'll see it

savankansagara1:
the work has been done please check it out

RichardTang-Aden:
## Review Feedback

The intent is correct — pytest doesn't belong in production dependencies. But this PR has problems:

1. **Not rebased onto current main.** `core/pyproject.toml` on main now has `[project.optional-dependencies]` commented out and uses a `[dependency-groups]` section instead (we did a `uv` migration). This PR still targets the old `[project.optional-dependencies]` structure, so it will conflict on merge and won't work as intended.

PR #2525, opened 2026-01-30
docs: clarify how to run tests locally

Sudhin-star1:
Fixes #2519

### Summary
Clarifies the correct way to run tests locally for Hive’s multi-package layout.

### Problem
Running `pytest` from the repo root or from `core/` fails with import
errors (e.g. `ModuleNotFoundError: No module named 'core'`), even though
tests pass in CI.

### Solution
Document the verified working command:

```bash
PYTHONPATH=core pytest core


Sudhin-star1:
This PR documents the verified local testing workflow I reproduced and validated:

`PYTHONPATH=core pytest core`

This resolves the import errors described in #2519.

PR #759, opened 2026-01-26
fix(core): replace silent exception suppression with error logging [micro-fix]

littlegoblinjr:
## Description
Fixes #715. Replaces `except Exception: pass` blocks in [orchestrator.py](cci:7://file:///e:/projects/aden/hive/core/framework/runner/orchestrator.py:0:0-0:0) and [runner.py](cci:7://file:///e:/projects/aden/hive/core/framework/runner/runner.py:0:0-0:0) with proper `logger.warning()` calls. This ensures that LLM API failures, rate limits, and network issues are visible to operators instead of failing silently.
## Type of Change
- [x] Bug fix (Reliability improvement)
## Changes Made
- Modified [core/framework/runner/orchestrator.py](cci:7://file:///e:/projects/aden/hive/core/framework/runner/orchestrator.py:0:0-0:0): Added logging to [_llm_route](cci:1://file:///e:/projects/aden/hive/core/framework/graph/node.py:989:4-1058:42) exception handler.
- Modified [core/framework/runner/runner.py](cci:7://file:///e:/projects/aden/hive/core/framework/runner/runner.py:0:0-0:0): Added logging to [can_handle](cci:1://file:///e:/projects/aden/hive/core/framework/runner/runner.py:859:4-950:54) exception handler.
- Added `import logging` where missing.
## Testing
- [x] Verified with script: Mocked LLM to raise connection errors and confirmed they are now logged as warnings.

PR #2533, opened 2026-01-30
fix(core): replace silent exception suppression [micro-fix]

### LINKED ISSUE #715: [Bug]: Format String Sandbox Bypass
The code sandbox in code_sandbox.py uses AST analysis to block access to private attributes like __globals__ and __class__. However, this check only catches direct attribute access in the AST. Format strings completely bypass this protection. 

When you write foo.__globals__, the AST validator catches it. But '{0.__globals__}'.format(foo) passes validation because the private attribute access is hidden inside a string literal. I verified this experimentally — you can extract __globals__, __class__, __mro__, and __bases__ through format strings. While this returns string representations rather than actual objects (limiting exploitation), it's still an information disclosure vulnerability that breaks the sandbox's security model.                                              

The fix requires either scanning string literals for format patterns containing __, overriding str.format() in the sandbox namespace, or adopting RestrictedPython for more comprehensive protection.

littlegoblinjr:
Fixes #715. Replaces silent exceptions with logger.warning. (Resubmission of #759)


Hundao:
Closing, #715 and #759 are both resolved.

PR #2547, opened 2026-01-30
micro-fix: Replace deprecated datetime.utcnow() with datetime.now(UTC)

Sourabsb:
Fixes #2545

## Changes
- Updated import in `AdenCredentialClient` to include `UTC` from `datetime`
- Replaced deprecated `datetime.utcnow().isoformat() + "Z"` with timezone-aware `datetime.now(UTC).isoformat()`

## Impact
- Eliminates Python 3.12+ deprecation warning
- Produces timezone-aware timestamps (`+00:00` instead of naive `Z` suffix)
- Aligns with the rest of the credentials module which already uses `datetime.now(UTC)`

## Testing
- ✅ Lint checks passed
- ✅ Format checks passed  
- ✅ Import verified - no deprecation warnings
- ✅ Timestamp format compatibility confirmed

### DIFF
--- a/core/framework/credentials/aden/client.py
+++ b/core/framework/credentials/aden/client.py
@@ -32,7 +32,7 @@ import logging
 import os
 import time
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Any
 
 import httpx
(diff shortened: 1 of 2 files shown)


Hundao:
The credential v2 refactor deleted `AdenCredentialClient.report_usage()` entirely while this was sitting in review, so there's nothing left for the patch to touch. Your change was correct when you opened it, the code just moved out from under it. Closing, but thanks for the contribution @Sourabsb.

If you want a real follow-up: `provider.py` (lines 178 and 331) still calls `self._client.report_usage(...)`, but the client doesn't have that method anymore. Latent `AttributeError`, only fires when `report_usage=True`. Fair game for a fresh PR if you're up for it.

PR #2605, opened 2026-01-30
fix: make agent builder exports atomic

### LINKED ISSUE #792: [Bug]: Non-Atomic file writes can corrupt run, summary, and index data

MuzzaiyyanHussain:
### Summary
This PR makes agent builder export writes crash-safe by switching to atomic file writes.

### Details
Previously, exported files (e.g. `agent.json`, `README.md`, and `mcp_servers.json`) were written directly using non-atomic `open(..., "w")` calls.  
If the process crashed or was interrupted mid-write, these files could be left partially written or corrupted.

This change updates the export flow to:
- write output to a temporary file
- flush and fsync file contents
- atomically replace the target file using `Path.replace()`

This ensures exported files are either fully written or remain unchanged after failures.

### Scope
- Limited to the agent builder export flow in `agent_builder_server.py`
- No behavior changes outside of file I/O safety

Any remaining related changes will be handled in a separate PR to keep this one focused.

### Testing
No automated tests were added. The existing test suite does not simulate mid-write crashes, and this change is limited to improving file I/O atomicity.
fixes #792

### DIFF
--- a/core/framework/mcp/agent_builder_server.py
+++ b/core/framework/mcp/agent_builder_server.py
@@ -32,6 +32,15 @@ SESSIONS_DIR = Path(".agent-builder-sessions")
 ACTIVE_SESSION_FILE = SESSIONS_DIR / ".active"
 
 
+def atomic_write_text(path: Path, write_fn) -> None:
+    tmp_path = path.with_suffix(path.suffix + ".tmp")
+    with open(tmp_path, "w", encoding="utf-8") as f:
+        write_fn(f)
+        f.flush()
+        os.fsync(f.fileno())
+    tmp_path.replace(path)
+
+
 # Session storage
 class BuildSession:
     """Build session with persistence support."""
@@ -122,12 +131,12 @@ def _save_session(session: BuildSession):
 
     # Save session file
     session_file = SESSIONS_DIR / f"{session.id}.json"
-    with open(session_file, "w") as f:
-        json.dump(session.to_dict(), f, indent=2, default=str)
+    atomic_write_text(
+        session_file, lambda f: json.dump(session.to_dict(), f, indent=2, default=str)
+    )
 
     # Update active session pointer
-    with open(ACTIVE_SESSION_FILE, "w") as f:
-        f.write(session.id)
+    atomic_write_text(ACTIVE_SESSION_FILE, lambda f: f.write(session.id))
 
 
 def _load_session(session_id: str) -> BuildSession:
@@ -246,8 +255,7 @@ def load_session_by_id(session_id: Annotated[str, "ID of the session to load"]) 
         _session = _load_session(session_id)
 
         # Update active session pointer
-        with open(ACTIVE_SESSION_FILE, "w") as f:
-            f.write(session_id)
+        atomic_write_text(ACTIVE_SESSION_FILE, lambda f: f.write(session_id))
 
         return json.dumps(
             {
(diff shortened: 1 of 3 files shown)


MuzzaiyyanHussain:
@austin931114 @TimothyZhang7 Green CI now special thanks to @Hundao

MuzzaiyyanHussain:
@RichardTang-Aden

Hundao (changes requested):
Thanks for tackling this. Atomic writes are the right fix for crash safety.

A few things to address before merging:

**1. Duplicate `atomic_write_text`**

The same function is defined in both `agent_builder_server.py` and `backend.py`. This should live in a shared module (e.g. `core/framework/utils/io.py`) and be imported from both places. Duplicating it means future fixes only land in one copy.

**2. No tmp cleanup on failure**

If `write_fn` raises (e.g. `json.dump` hits a non-serializable object), the `.tmp` file is left on disk. Wrapping with try/except to unlink on failure would fix this:

```python
try:
    with open(tmp_path, "w", encoding="utf-8") as f:
        write_fn(f)
        f.flush()
        os.fsync(f.fileno())
    tmp_path.replace(path)
except BaseException:
    tmp_path.unlink(missing_ok=True)
    raise
```

**3. Consider a context manager instead of callback**

The lambda pattern works but a context manager would be more idiomatic Python and easier to read at call sites:

```python
# current
atomic_write_text(path, lambda f: json.dump(data, f, indent=2))

# context manager
with atomic_write(path) as f:
    json.dump(data, f, indent=2)
```

This follows the same pattern as `open()` so it's immediately familiar. Happy to elaborate if needed.

Overall direction is solid, just needs some polish. Let me know if you have questions.

MuzzaiyyanHussain:
@Hundao CI/lint is still failing after ruff formatting

Hundao:
> @Hundao CI/lint is still failing after ruff formatting

Let me check this, thanks for the report.

Hundao:
Hi @MuzzaiyyanHussain,

I fixed the issue 😅. Could you rebase on the latest main? The lint error should disappear after that.

MuzzaiyyanHussain:
@Hundao Thanks, no need to rebase i pushed an empty commit just to rerun CI, now please review the changes. Green CI now 😍

MuzzaiyyanHussain:
@Hundao  Thanks for the review and merge! 🙌
Feel free to assign me any issue you think I’d be a good fit for , happy to help.

PR #2704, opened 2026-01-30
micro-fix: Refactor testing framework to use LiteLLM

amritendunath:
## Summary
Fixes #2698

Removes the hardcoded dependency on the `anthropic` library in the testing framework (llm_judge.py and prompts.py). The framework now uses LiteLLMProvider for all LLM interactions, making the testing system provider-agnostic.

## Changes
- **Refactor LLMJudge**: Updated llm_judge.py to use `framework.llm.LiteLLMProvider` instead of the direct `anthropic` SDK.
- **Update Default Model**: Configured the judge to use `claude-3-haiku-20240307` via LiteLLM by default, maintaining the existing behaviour while allowing architectural flexibility.
- **Generic Test Templates**: Updated prompts.py test headers to check for *any* valid LLM API key (OpenAI, Gemini, etc.) rather than strictly enforcing `ANTHROPIC_API_KEY`.
- **Verification**: Added core/tests/test_llm_judge_agnostic.py to verify the judge initializes without provider-specific dependencies.

## Motivation
Previously, the testing framework would fail for users who did not have the `anthropic` package installed or an Anthropic API key, even if they were using other providers. This change aligns the testing framework with the core framework's multi-provider design.


Hundao:
Closing this, the underlying issue #2698 got resolved a different way. The MCP server was refactored so the test-gen tools return guidelines instead of calling an LLM, so there's no provider to make configurable anymore. Thanks for the patch though.

PR #2942, opened 2026-01-31
fix(docs): correct test command in DEVELOPER.md

### LINKED ISSUE #2879: [Bug]: Incorrect test command in DEVELOPER.md uses `python -m pytest`
## Describe the Bug

The Git Workflow section in `DEVELOPER.md` instructs contributors to run tests locally using:

    PYTHONPATH=core:exports python -m pytest

This command is incorrect and may fail, because `pytest` is not intended to be executed via `python -m pytest` in this project setup.

Contributors following this documentation may encounter errors or believe the test setup is broken.

## To Reproduce

1. Open `DEVELOPER.md`
2. Go to the **Git Workflow → Pull Request Process** section
3. Run the command shown in the docs:

       PYTHONPATH=core:exports python -m pytest

4. Observe that tests do not run correctly or pytest is not found/executed as expected.

## Expected Behavior

The documentation should instruct contributors to run pytest directly, for example:

    PYTHONPATH=core:exports pytest

This matches common Python and pytest usage and works correctly when pytest is installed in the environment.

## Screenshots

documentation issue, N/A

## Environment

- OS: Any
- Python version: 3.11+

## Configuration

N/A

## Logs

Example error when following the current documentation (may vary by environment):

    /usr/bin/python: No module named pytest

## Additional Context

Updating this command in `DEVELOPER.md` will prevent confusion for new contributors and ensure the documented local testing workflow actually works as written.

6736-shafi:
## Summary

Fixes the incorrect test command in `DEVELOPER.md` Git Workflow section.

## Change

**Before:**
```bash
PYTHONPATH=core:exports python -m pytest
```

**After:**
```bash
PYTHONPATH=core:exports pytest
```

## Why

The original command `python -m pytest` may fail because:
- pytest is not intended to be executed via `python -m pytest` in this project setup
- Running `pytest` directly is the standard approach when pytest is installed in the environment

This fixes contributor confusion when following the documentation.

Fixes #2879

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -509,7 +509,7 @@ chore(deps): update React to 18.2.0
 
 1. Create a feature branch from `main`
 2. Make your changes with clear commits
-3. Run tests locally: `PYTHONPATH=core:exports python -m pytest`
+3. Run tests locally: `PYTHONPATH=core:exports pytest`
 4. Run linting: `black --check .`
 5. Push and create a PR
 6. Fill out the PR template


PatrickChen928:
Thanks for the contribution @6736-shafi 

  However, `python -m pytest` is actually a valid and officially supported way to run pytest. It ensures pytest runs with the correct Python environment, which can be helpful in multi-version setups.

  The current documentation isn't incorrect — both `python -m pytest` and `pytest` work fine. Since this is more of a style preference than a bug fix, I'll close this PR.

  Appreciate your attention to the docs! Feel free to check out other open issues. 🙏

PR #3186, opened 2026-02-02
docs: add Core Concepts page (concepts-first)

srinuk9570:
Adds an initial Core Concepts page focused on the “what/why” of Hive:
- Goal → graph mental model
- Adaptation loop
- Human-in-the-loop
- Observability & cost controls

No setup or CLI content included.
Intended for PMs and architects evaluating Hive before a technical POC.

### DIFF
--- a/docs/core-concepts.md
+++ b/docs/core-concepts.md
@@ -1,0 +1,63 @@
+# Core Concepts
+
+This page explains the core ideas behind Hive — what it is, how it works, and why it is different — before diving into setup or CLI usage.
+
+## What is Hive?
+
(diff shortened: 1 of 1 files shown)


thebrandhutt:
This is perfect, thanks for turning this into a Core Concepts page so quickly. I really like how you've structured it around goal → graph, adaptation loop, HITL, observability, and cost controls.

The concepts‑first, no‑CLI approach is spot on for PMs/architects doing a first‑pass evaluation. Once this lands, I'm happy to help iterate on narrative flow and visual aids (e.g. a simple goal → graph diagram or where HITL nodes sit in the loop) to make the mental model even clearer.

PR #3254, opened 2026-02-02
feat: add TTL-based in-memory cache to web_search

### LINKED ISSUE #2597: [Enhancement]: Add in-memory caching for web_search results

GovindhKishore:
## Description

This PR adds a transparent, in-memory caching mechanism to the `web_search` tool to reduce redundant external API calls for identical queries. The cache uses a configurable TTL and size limit to improve performance and reduce API quota usage, without changing the existing response format or behavior.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #2597

## Changes Made

- Added a simple in-memory cache with TTL to `web_search`
- Introduced a composite cache key derived from query parameters
- Ensured caching is transparent and does not alter the response format
- Prevented caching of error responses
- Normalized provider selection for consistent cache hits when using `provider="auto"`

## Testing

- [ ] Unit tests pass (`pytest`)
- [ ] Lint passes

**Notes:**
- Attempted to run the test suite and individual `web_search` tests locally.
- Test collection fails due to missing optional third-party dependencies (e.g. `resend`) used by other tools and imported via `aden_tools.tools.__init__`.
- These dependencies are unrelated to `web_search`, and the caching change does not modify provider behavior or response format.

## Checklist

N/A


GovindhKishore:
Hi @TimothyZhang7 I’ve opened a PR that addresses issue #2597.  
Would appreciate your review when you have time. Please let me know if any changes are needed and I’ll be happy to update it. Thanks!

PR #3419, opened 2026-02-03
docs(developer): clarify venv activation before make check/test

AmanKushwaha-17:
This PR clarifies that the framework virtual environment (core/.venv) should be activated before running make check and make test, since tools like ruff are installed within the venv.

This helps new contributors run local checks successfully before opening a PR.


RichardTang-Aden:
No longer relevant — we use uv now, which handles the venv automatically. No need to manually activate before running make check/test.

PR #3558, opened 2026-02-04
fix(storage): add file locking to prevent race condition in _update_index

### LINKED ISSUE #3451: [Bug]: Race condition in EncryptedFileStorage._update_index causes silent credential loss
## Summary

`EncryptedFileStorage._update_index()` uses a non-atomic read-modify-write pattern without file locking. Concurrent credential operations can silently lose data.

## Location

**File:** `core/framework/credentials/storage.py:263-289`

```python
# READ
with open(index_path) as f:
    index = json.load(f)

# MODIFY
index["credentials"][credential_id] = {...}

# WRITE - overwrites concurrent updates
with open(index_path, "w") as f:
    json.dump(index, f, indent=2)
```

## Reproduction

Two concurrent `storage.save()` calls can lose one credential:

```
Thread A: read {}
Thread B: read {} → add cred_B → write {cred_B}
Thread A: add cred_A → write {cred_A}  ← overwrites cred_B!
```

Result: `cred_B` silently lost, no error raised.

## Impact

- Credentials silently lost during concurrent saves
- JSON corruption if read occurs mid-write
- Affects multi-threaded apps or multiple processes

## Fix

Add file locking (`fcntl.flock`) around read-modify-write cycle.

Ttian18:
## Summary
- Add file locking to `_update_index` and `list_all` for concurrency safety
- Use `fcntl.flock` with exclusive lock (LOCK_EX) for writes and shared lock (LOCK_SH) for reads
- Use `.index.lock` file with append mode for coordination
- Prevent silent credential loss during concurrent saves/deletes

## Changes since last review
Rebased to latest main and addressed all review feedback from @RichardTang-Aden:
1. ✅ Removed Windows `msvcrt` codepath — using `fcntl` directly (WSL required for Windows)
2. ✅ Added file locking to `list_all()` with `LOCK_SH` for read consistency
3. ✅ Changed lock file mode from `"w"` to `"a"` to avoid truncation
4. ✅ Preserved upstream's `utf-8-sig` encoding changes

## Test plan
- [x] 20 concurrent saves preserve all credentials (was losing ~50%)
- [x] 50 concurrent saves under stress test
- [x] Concurrent save/delete operations remain consistent
- [x] Lock file is created
- [x] All 68 tests passing

Fixes #3451

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced credential storage reliability by improving data consistency guarantees during concurrent operations.

* **Tests**
  * Added comprehensive test coverage validating concurrent credential operation scenarios, including parallel saves, mixed save and delete workflows, and end-to-end integrity verification across storage operations.

### DIFF
--- a/core/framework/credentials/storage.py
+++ b/core/framework/credentials/storage.py
@@ -10,6 +10,7 @@ This module provides abstract and concrete storage implementations:
 
 from __future__ import annotations
 
+import fcntl
 import json
 import logging
 import os
@@ -223,13 +224,19 @@ class EncryptedFileStorage(CredentialStorage):
         return False
 
     def list_all(self) -> list[str]:
-        """List all credential IDs."""
+        """List all credential IDs with file locking for read consistency."""
         index_path = self.base_path / "metadata" / "index.json"
         if not index_path.exists():
             return []
-        with open(index_path, encoding="utf-8-sig") as f:
-            index = json.load(f)
-        return list(index.get("credentials", {}).keys())
+        lock_path = self.base_path / "metadata" / ".index.lock"
+        with open(lock_path, "a") as lock_file:
+            fcntl.flock(lock_file, fcntl.LOCK_SH)
+            try:
+                with open(index_path, encoding="utf-8-sig") as f:
+                    index = json.load(f)
+                return list(index.get("credentials", {}).keys())
+            finally:
+                fcntl.flock(lock_file, fcntl.LOCK_UN)
 
     def exists(self, credential_id: str) -> bool:
         """Check if credential exists."""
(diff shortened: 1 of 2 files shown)


Ttian18:
@TimothyZhang7 Ready for review!

RichardTang-Aden:
Hey, thanks for this — the race condition in `_update_index` is real (read-modify-write without synchronization), and the `fcntl.flock` fix is correct.

A few things to address before we can merge:

1. **Fix or drop the Windows `msvcrt` codepath.** We require WSL for Windows users, so `fcntl` is always available. The `msvcrt.locking` path is dead code and has subtle issues anyway (`LK_LOCK` has a ~1s timeout vs `flock` which blocks indefinitely).

2. **`list_all()` reads `index.json` without the lock.** A concurrent `save()` could be mid-write. Wrap it with the same lock for consistency.

3. **Minor: open the lock file with `"a"` instead of `"w"`.** Avoids truncating the file on every call. Doesn't matter functionally on Unix, but cleaner.

Fix those up and this is good to go.

Ttian18:
@RichardTang-Aden Thanks for the review! I've addressed all your feedback:

1. ✅ Removed Windows `msvcrt` codepath - using `fcntl` directly
2. ✅ Added file locking to `list_all()` for read consistency
3. ✅ Changed lock file mode from `"w"` to `"a"` to avoid truncation

Ready for another look!

Ttian18:
@RichardTang-Aden Rebased to latest main and addressed all your feedback. Ready for another look!

PR #3694, opened 2026-02-05
fix: write node outputs to memory before edge evaluation (#3599)

### LINKED ISSUE #3599: [Bug]: Conditional Edge Direct Key Access Fails - Node Outputs Not Written to Memory Before Edge Evaluation
## Describe the Bug

  Conditional edges cannot use direct key access (e.g., `user_score > 80`) even though the code explicitly supports it via `**memory` unpacking in
  `edge.py:171`. Node outputs are written to memory AFTER edge evaluation instead of before, making the intended direct key access feature
  completely broken.

  **Root Cause:**
  The execution order in GraphExecutor is incorrect:
  1. Node executes → produces `result.output` (executor.py:328)
  2. Edge evaluation → calls `edge.should_traverse()` (executor.py:734)
  3. Conditional expression evaluated with `**memory` context (edge.py:171)
  4. **BUG:** Source node outputs NOT in memory yet!
  5. Mapped inputs written to memory (executor.py:788-789) - too late

  This breaks the intended feature documented in `edge.py:171` comment: "Include memory keys directly for easier access in conditions"

  ## To Reproduce

  1. Create a node that outputs a value:
  ```python
  class ProducerNode(NodeProtocol):
      async def execute(self, ctx: NodeContext) -> NodeResult:
          return NodeResult(success=True, output={"user_score": 85})

  2. Create a conditional edge using direct key access:
  EdgeSpec(
      id="producer_to_consumer",
      source="producer",
      target="consumer",
      condition=EdgeCondition.CONDITIONAL,
      condition_expr="user_score > 80"  # Should work but fails
  )

  3. Execute the graph
  4. See error: Name 'user_score' is not defined

  Full test case: https://gist.github.com/... (test_memory_bug.py demonstrating the issue)

  Expected Behavior

  The condition user_score > 80 should evaluate successfully when the source node outputs {"user_score": 85}.

  Direct key access is explicitly supported by the code:
  - edge.py:171 unpacks **memory into the evaluation context
  - Comment states: "Include memory keys directly for easier access in conditions"
  - This feature should work as designed

  Screenshots

  ⚠ Condition evaluation failed: user_score > 80
     Error: Name 'user_score' is not defined
     Available context keys: ['output', 'memory', 'result', 'true', 'false']

  The consumer node is never reached because the edge condition fails.

  Environment

  - OS: macOS 14 / Linux (reproduced on both)
  - Python version: 3.11
  - Hive version: main branch (latest)

  Configuration

  Minimal reproduction:

  nodes = [
      NodeSpec(
          id="producer",
          name="Producer",
          node_type="function",
          output_keys=["user_score"]
      ),
      NodeSpec(
          id="consumer",
          name="Consumer",
          node_type="function",
          input_keys=["user_score"]
      )
  ]

  edges = [
      EdgeSpec(
          id="producer_to_consumer",
          source="producer",
          target="consumer",
          condition=EdgeCondition.CONDITIONAL,
          condition_expr="user_score > 80"  # BROKEN
      )
  ]

  Logs

  📤 Producer: writing 'user_score' to output
  ⚠ Condition evaluation failed: user_score > 80
     Error: Name 'user_score' is not defined
     Available context keys: ['output', 'memory', 'result', 'true', 'false']

  Success: True
  Path: producer
  # Consumer never executed - edge condition failed

  Additional Context

  Impact:
  - Breaks intended direct key access feature in conditional expressions
  - Forces verbose output['user_score'] > 80 syntax, defeating the purpose of **memory unpacking
  - Inconsistent with documented design intent in edge.py:171
  - Makes conditional expressions harder to read and write

  Workaround:
  Use the verbose syntax: condition_expr="output['user_score'] > 80"

  Proposed Fix:
  Write node outputs to SharedMemory before calling edge.should_traverse() in executor.py:

  # After line 328-348 (successful node execution):
  if result.success and result.output:
      for key, value in result.output.items():
          memory.write(key, value, validate=False)

  # Then proceed with edge evaluation at line 734

  Files Involved:
  - core/framework/graph/executor.py lines 328-789 (execution and edge evaluation sequence)
  - core/framework/graph/edge.py lines 154-186 (conditional expression evaluation)

  Related Issues:
  - Edge routing was recently fixed in #3449, but this memory timing issue was not caught
  - Similar to data flow issues but specifically about evaluation context timing

Acid-OP:
## Description

  Fixes conditional edges with direct key access by writing node outputs to SharedMemory immediately after execution, before edge evaluation.
  Previously, expressions like `score > 80` failed with NameError because outputs were only written during input mapping, which happens after edge
  evaluation.

  ## Type of Change

  - [ ] Documentation update

  ## Related Issues

  Fixes #3599

  ## Changes Made

  - Added code to write node outputs to SharedMemory immediately after successful execution (executor.py:368-373)
  - Outputs are now available in memory before edge conditions are evaluated
  - Enables direct key syntax in conditional edges: `score > 80` instead of `output['score'] > 80`
  - Maintains backward compatibility with existing `output['key']` syntax

  ## Testing

  Comprehensive testing performed to verify the fix:

  **Test Coverage:**
  - ✅ Simple direct key access (`score > 80`)
  - ✅ Multiple keys in expressions (`x > y and y < z`)
  - ✅ Complex mathematical expressions (`(a * b) % c == 2`)
  - ✅ Safe string methods (`.split()`, `.lower()`)
  - ✅ Safe dict methods (`.get()`)
  - ✅ Negative numbers and floats
  - ✅ Backward compatibility with `output['key']` syntax
  - ✅ Negative case - conditions correctly evaluate to false

  All 10 test scenarios passed.

  ## Checklist

  N/A - Backend fix with no UI changes.

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -364,6 +364,13 @@ class GraphExecutor:
                             if len(value_str) > 200:
                                 value_str = value_str[:200] + "..."
                             self.logger.info(f"      {key}: {value_str}")
+
+                    # Write node outputs to memory BEFORE edge evaluation
+                    # This enables direct key access in conditional expressions (e.g., "score > 80")
+                    # Without this, conditional edges can only use output['key'] syntax
+                    if result.output:
+                        for key, value in result.output.items():
+                            memory.write(key, value, validate=False)
                 else:
                     self.logger.error(f"   ✗ Failed: {result.error}")
 
--- a/core/tests/test_conditional_edge_direct_key.py
+++ b/core/tests/test_conditional_edge_direct_key.py
@@ -1,0 +1,344 @@
+"""
+Regression tests for conditional edge direct key access (Issue #3599).
+
+Verifies that node outputs are written to memory before edge evaluation,
+enabling direct key access in conditional expressions (e.g., 'score > 80')
+instead of requiring output['score'] > 80 syntax.
+"""
+
+import pytest
+
+from framework.graph.edge import EdgeCondition, EdgeSpec, GraphSpec
+from framework.graph.executor import GraphExecutor
+from framework.graph.goal import Goal
+from framework.graph.node import NodeContext, NodeProtocol, NodeResult, NodeSpec
+from framework.runtime.core import Runtime
+
+
+class SimpleRuntime(Runtime):
+    """Minimal runtime for testing."""
+
+    def start_run(self, **kwargs):
+        return "test-run"
+
+    def end_run(self, **kwargs):
+        pass
+
+    def report_problem(self, **kwargs):
+        pass
+
+    def decide(self, **kwargs):
+        return "test-decision"
+
+    def record_outcome(self, **kwargs):
+        pass
+
+    def set_node(self, **kwargs):
+        pass
+
+
+class ScoreNode(NodeProtocol):
+    """Node that outputs a score value."""
+
+    async def execute(self, ctx: NodeContext) -> NodeResult:
+        return NodeResult(success=True, output={"score": 85})
+
+
+class HighScoreNode(NodeProtocol):
+    """Consumer node for high scores."""
+
+    async def execute(self, ctx: NodeContext) -> NodeResult:
+        return NodeResult(success=True, output={"result": "high_score_path"})
+
+
+class MultiKeyNode(NodeProtocol):
+    """Node that outputs multiple keys."""
+
+    async def execute(self, ctx: NodeContext) -> NodeResult:
+        return NodeResult(success=True, output={"x": 100, "y": 50})
+
+
+class ConsumerNode(NodeProtocol):
+    """Generic consumer node."""
+
+    async def execute(self, ctx: NodeContext) -> NodeResult:
+        return NodeResult(success=True, output={"processed": True})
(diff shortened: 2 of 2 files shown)


Acid-OP:
@austin931114 @Hundao @bryanadenhq @RichardTang-Aden @PatrickChen928 @TimothyZhang7
Please review my PR.

Hundao (commented):
The fix looks correct. The root cause analysis is right and the change is minimal and targeted.

One thing before we merge: could you add an automated regression test for this? The PR description lists 10 manual test scenarios, but there's no test file in the diff. This kind of execution order bug is easy to regress on.

Even a single test would be enough, something like:

```python
async def test_node_output_available_in_edge_condition():
    """Verify direct key access works in conditional edges (e.g., 'score > 80')."""
    # Set up a graph with a producer node that outputs {"score": 85}
    # and a conditional edge with condition_expr="score > 80"
    # Assert that the consumer node is reached
```

The workaround syntax `output['score'] > 80` should keep working too, so a backward compatibility assertion would be nice to have.

Acid-OP:
@Hundao  Regression tests added. All 4 tests passing. Ready for merge.

PR #4012, opened 2026-02-07
fix: race condition deadlock, missing short-circuit eval, unhandled format exceptions

### LINKED ISSUE #4009: bug: Race condition in EventLoopNode._await_user_input can cause permanent deadlock
## Description

In `EventLoopNode._await_user_input()` (`core/framework/graph/event_loop_node.py`, line 827-836), there is a race condition between `emit_client_input_requested()` and the `_input_ready` event flag that can cause the node to deadlock permanently.

## Root Cause

The current code sequence is:

```python
await self._event_bus.emit_client_input_requested(...)  # (1) await - yields to event loop
self._input_ready.clear()                                # (2) clears the flag
await self._input_ready.wait()                           # (3) waits for flag
```

During the `await` at step (1), the event loop can schedule an `inject_event()` call (e.g., from a WebSocket handler responding to the `CLIENT_INPUT_REQUESTED` event). `inject_event()` puts content on the queue and calls `self._input_ready.set()`. But when step (1) returns, step (2) immediately clears the flag that was just set, and step (3) blocks forever.

The injected content sits on the `_injection_queue` but is never consumed because the node is stuck in `wait()`.

## Impact

- Client-facing nodes can deadlock permanently under normal usage
- Only recoverable via `signal_shutdown()` from external code
- More likely to occur when user input arrives quickly after the node requests it

## Reproduction

This occurs when `inject_event()` is called during the `await` of `emit_client_input_requested()` — i.e., when an event handler responds immediately to the `CLIENT_INPUT_REQUESTED` event by injecting user input.

## Proposed Fix

Move `self._input_ready.clear()` **before** the emit call so that any `inject_event()` triggered during the emit is preserved:

```python
self._input_ready.clear()  # Clear first
await self._event_bus.emit_client_input_requested(...)  # Now safe to await
await self._input_ready.wait()  # Will see the set() from inject_event
```

mma2027:
## Summary

Fixes three bugs found during code review:

- **Race condition deadlock in `EventLoopNode._await_user_input()`** — `_input_ready.clear()` was called after the `emit_client_input_requested()` await, creating a window where `inject_event()` could set the flag only for `clear()` to immediately wipe it, causing `wait()` to block forever. Fix: move `clear()` before the emit. (Fixes #4009)

- **Missing short-circuit evaluation in `safe_eval.visit_BoolOp()`** — All operands of `and`/`or` were eagerly evaluated via list comprehension, breaking guard patterns like `x is not None and x.get("key")` which would crash instead of short-circuiting. Fix: iterate operands with early return. (Fixes #4010)

- **Unhandled `TypeError`/`IndexError` in `HybridJudge._format_feedback()`** — Template formatting with `{result[key]}` raises `TypeError` when `result` is not subscriptable, which was not caught, crashing the judge evaluation. Fix: add `TypeError` and `IndexError` to the except clause. (Fixes #4011)

## Test plan

- [x] All 67 existing tests pass (test_event_loop_node: 35, test_event_loop_integration: 17, test_event_loop_wiring: 10, test_graph_executor: 5)
- [ ] Verify `_await_user_input` no longer deadlocks when `inject_event()` arrives during emit
- [ ] Verify edge conditions with guard patterns like `x is not None and x["key"]` work correctly
- [ ] Verify judge rules with non-dict result values don't crash

### DIFF
--- a/core/framework/graph/safe_eval.py
+++ b/core/framework/graph/safe_eval.py
@@ -115,11 +115,23 @@ class SafeEvalVisitor(ast.NodeVisitor):
         return True
 
     def visit_BoolOp(self, node: ast.BoolOp) -> Any:
-        values = [self.visit(v) for v in node.values]
+        # Short-circuit evaluation to match Python semantics.
+        # Previously all operands were eagerly evaluated, which broke
+        # guard patterns like: ``x is not None and x.get("key")``
         if isinstance(node.op, ast.And):
-            return all(values)
+            result = True
+            for v in node.values:
+                result = self.visit(v)
+                if not result:
+                    return result
+            return result
         elif isinstance(node.op, ast.Or):
-            return any(values)
+            result = False
+            for v in node.values:
+                result = self.visit(v)
+                if result:
+                    return result
+            return result
         raise ValueError(f"Boolean operator {type(node.op).__name__} is not allowed")
 
     def visit_IfExp(self, node: ast.IfExp) -> Any:


Hundao:
@mma2027 Rebased your PR to keep only the safe_eval short-circuit fix (the other two were already resolved upstream). Merged, thanks!

PR #4277, opened 2026-02-09
fix(graph): honor graph max_retries_per_node

waqasm86:
## Summary
- Honor `graph.max_retries_per_node` when a node does not explicitly set `max_retries`
- Apply the same logic to fan-out branch retries
- Add regression test to cover graph-level defaults

## Testing
- Not run (local)

Fixes #4259


waqasm86:
Tests (Python 3.11):
- `pytest core/tests -v`
  - 679 passed, 74 skipped, 204 warnings
- `pytest core/tests/test_executor_max_retries.py -v`
  - 6 passed
Notes: warnings are deprecations from deps; no new failures observed.

PR #4299, opened 2026-02-10
docs(fix): correct LLMNode token counting bug

### LINKED ISSUE #4266: [Bug]: Token counting in LLMNode only uses last response, ignoring retry tokens
## Describe the Bug
I was debugging why my agent's reported token usage was way lower than what I was seeing in my LLM provider dashboard. Turns out there's a bug in `core/framework/graph/node.py` where the final `NodeResult` only counts tokens from the last LLM response, completely discarding tokens from compaction and validation retries.

The code correctly accumulates tokens throughout the retry loops in `total_input_tokens `and `total_output_tokens` (lines 778-945), but then line 1047 just uses:

`tokens_used=response.input_tokens + response.output_tokens,`

This ignores all the accumulated retry tokens.

What's odd is that the error paths in the same file (lines 914, 925, 936) all correctly use total_input_tokens + total_output_tokens. So it looks like the success path just got missed during a refactor.

## To Reproduce

1. Create a node with Pydantic validation that will fail on first attempt:

```python
from pydantic import BaseModel

class OutputModel(BaseModel):
    result: str

node_spec = NodeSpec(
    id="test_node",
    node_type="llm_generate",
    output_keys=["result"],
    output_model=OutputModel,
    max_validation_retries=2,
    max_compaction_retries=1,
)

```

2. Execute the node with a prompt that triggers retries (e.g., LLM returns wrong field names initially)
3. Check the returned NodeResult.tokens_used
4. Compare to actual LLM API calls made

## Expected Behavior
Should report total tokens across all LLM calls:

- Initial call tokens
- All compaction retry tokens
- All validation retry tokens

Currently it only reports the last response's tokens.

## Screenshots
N/A

## Environment
OS: Windows 11 (Build 26200)
Python version: 3.14.2
Running on: Native Windows with Git Bash

## Configuration
Any node config with `output_model` set or `max_compaction_retries > 0` will exhibit this bug when retries occur.

## Logs
Example scenario with actual vs reported tokens:

Initial LLM call: 1000 input + 500 output = 1500 tokens
Compaction retry: 1200 input + 600 output = 1800 tokens  
Validation retry: 1600 input + 800 output = 2400 tokens

Actual total: 5700 tokens
Reported in NodeResult: 2400 tokens (only the last call)
Undercount: ~58%

## Additional Context

This has real impact on:

- Budget enforcement (limits based on underreported usage)
- Cost tracking (can't accurately estimate LLM expenses)
- Production monitoring (metrics are wrong)

The fix is straightforward - change line 1047 to use the accumulated totals like the error paths do. I can submit a PR if that's helpful.

Also affects line 1028 (error path for JSON extraction failure) which has the same issue.

YashovardhanB28:
### Summary
Updated `LLMNode` to use `total_input_tokens` and `total_output_tokens` in the final result, ensuring that tokens from validation retries are correctly counted.

### Fixes
Fixes #4266

### Changes
- Swapped `response.input_tokens` with `total_input_tokens` in `record_outcome` and `NodeResult`.

### DIFF
--- a/core/framework/graph/node.py
+++ b/core/framework/graph/node.py
@@ -1073,7 +1073,7 @@ Keep the same JSON structure but with shorter content values.
                 decision_id=decision_id,
                 success=True,
                 result=response.content,
-                tokens_used=response.input_tokens + response.output_tokens,
+                tokens_used=total_input_tokens + total_output_tokens,
                 latency_ms=latency_ms,
             )
 
@@ -1148,7 +1148,7 @@ Keep the same JSON structure but with shorter content values.
                             f"Expected keys: {ctx.node_spec.output_keys}"
                         ),
                         output={},
-                        tokens_used=response.input_tokens + response.output_tokens,
+                        tokens_used=total_input_tokens + total_output_tokens,
                         latency_ms=latency_ms,
                     )
                     # JSON extraction failed completely - still strip code blocks
@@ -1167,7 +1167,7 @@ Keep the same JSON structure but with shorter content values.
             return NodeResult(
                 success=True,
                 output=output,
-                tokens_used=response.input_tokens + response.output_tokens,
+                tokens_used=total_input_tokens + total_output_tokens,
                 latency_ms=latency_ms,
             )
 


bryanadenhq (approved):
Looks good, thanks for contributing!

PR #4451, opened 2026-02-11
fix(tui): add multiline input and cross-platform clipboard support

### LINKED ISSUE #4423: [Bug]: TUI copy + newline
## Problem Statement

The current TUI does not have very good copy support. The copy only works for ctrl+c and is not consistent on linux. There is also no newline support in the tui. When pasting prompt with multiple lines, the cursor disappears.

## Proposed Solution

Update the TUI to use the default copy buttons(cmd+c on mac, ctrl+c on linux, windows) and add newline support in the tui. Newline should be created when shift+enter.

Ttian18:
## Description

Fixes #4423

The TUI chat input uses a single-line `Input` widget that drops multiline
pastes and has no newline support. Copy-to-clipboard also fails on some
platforms.

## Type of Change

## Related Issues

Fixes #4423

## Changes Made

- Replace single-line `Input` widget with `TextArea` in chat REPL
  - Enter submits, Shift+Enter inserts a newline
  - Dynamic height (1–5 visible lines, scrolls beyond that)
  - Multiline paste now works correctly
- Add Windows clipboard support (`clip.exe`) in `_copy_to_clipboard`
- Add `xsel` fallback for Linux when `xclip` is not installed
- Update CSS selectors and widget queries in `app.py` to match new widget

## Testing

Manual verification:
- Shift+Enter inserts newlines in input box
- Multiline paste preserves all lines with visible cursor
- Ctrl+C copies selected text to clipboard
- Enter submits message and clears input
- Slash commands (/help, /sessions) still work

## Checklist

### DIFF
--- a/core/framework/tui/app.py
+++ b/core/framework/tui/app.py
@@ -6,7 +6,7 @@ import time
 from textual.app import App, ComposeResult
 from textual.binding import Binding
 from textual.containers import Container, Horizontal, Vertical
-from textual.widgets import Footer, Input, Label
+from textual.widgets import Footer, Label, TextArea
 
 from framework.runtime.agent_runtime import AgentRuntime
 from framework.runtime.event_bus import AgentEvent, EventType
@@ -180,13 +180,13 @@ class AdenTUI(App):
         scrollbar-color: $primary;
     }
 
-    Input {
+    TextArea {
         background: $surface;
         border: tall $primary;
         margin-top: 1;
     }
 
-    Input:focus {
+    TextArea:focus {
         border: tall $accent;
     }
 
@@ -504,8 +504,8 @@ class AdenTUI(App):
         original_chat_border = chat_widget.styles.border_left
         chat_widget.styles.border_left = ("none", "transparent")
 
-        # Hide all Input widget borders
-        input_widgets = self.query("Input")
+        # Hide all TextArea widget borders
+        input_widgets = self.query("TextArea")
         original_input_borders = []
         for input_widget in input_widgets:
             original_input_borders.append(input_widget.styles.border)
@@ -580,8 +580,8 @@ class AdenTUI(App):
         # Send /sessions command to chat input
         try:
             chat_repl = self.query_one(ChatRepl)
-            chat_input = chat_repl.query_one("#chat-input", Input)
-            chat_input.value = "/sessions"
+            chat_input = chat_repl.query_one("#chat-input", TextArea)
+            chat_input.text = "/sessions"
             # Trigger submission
             self.notify(
                 "💡 Type /sessions in the chat to see all sessions",
--- a/core/framework/tui/app.py
+++ b/core/framework/tui/app.py
@@ -6,7 +6,7 @@ import time
 from textual.app import App, ComposeResult
 from textual.binding import Binding
 from textual.containers import Container, Horizontal, Vertical
-from textual.widgets import Footer, Label, TextArea
+from textual.widgets import Footer, Label
 
 from framework.runtime.agent_runtime import AgentRuntime
 from framework.runtime.event_bus import AgentEvent, EventType
@@ -180,13 +180,13 @@ class AdenTUI(App):
         scrollbar-color: $primary;
     }
 
-    TextArea {
+    ChatTextArea {
         background: $surface;
         border: tall $primary;
         margin-top: 1;
     }
 
-    TextArea:focus {
+    ChatTextArea:focus {
         border: tall $accent;
     }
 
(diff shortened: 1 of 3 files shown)


core/framework/tui/widgets/chat_repl.py
```diff
@@ -739,26 +740,36 @@ def _check_and_show_resumable_sessions(self) -> None:
...
 
-    async def on_input_submitted(self, message: Input.Submitted) -> None:
-        """Handle input submission — either start new execution or inject input."""
-        user_input = message.value.strip()
+    async def on_key(self, event) -> None:
+        """Intercept Enter in TextArea to submit; Shift+Enter inserts newline."""
+        focused = self.app.focused
+        if not isinstance(focused, TextArea):
+            return
+        if event.key == "enter" and not event.shift:
```
jose-compu: `Key` events have no `.shift` attribute. The code `not event.shift` will raise `AttributeError`. The correct approach is to check `event.key == "enter"` (Shift+Enter arrives as `event.key == "shift+enter"`, so it naturally wouldn't match).

core/framework/tui/widgets/chat_repl.py
```diff
@@ -739,26 +740,36 @@ def _check_and_show_resumable_sessions(self) -> None:
             # Silently fail - don't block TUI startup
             pass
 
-    async def on_input_submitted(self, message: Input.Submitted) -> None:
-        """Handle input submission — either start new execution or inject input."""
-        user_input = message.value.strip()
+    async def on_key(self, event) -> None:
```
jose-compu: `TextArea._on_key` handles `"enter"` by:
1. Inserting `"\n"` into the document
2. Calling `event.stop()` — which **stops event propagation entirely**

This means `ChatRepl.on_key` will **never receive the Enter key event at all**. The Enter key is consumed by `TextArea` before it can bubble. The entire submission mechanism is broken — pressing Enter will just insert newlines, and submission will never trigger.

The fix requires subclassing `TextArea` and overriding `_on_key` to intercept Enter before the base class processes it.

core/framework/tui/app.py
```diff
@@ -580,8 +580,8 @@ def action_show_sessions(self) -> None:
         # Send /sessions command to chat input
         try:
             chat_repl = self.query_one(ChatRepl)
-            chat_input = chat_repl.query_one("#chat-input", Input)
-            chat_input.value = "/sessions"
+            chat_input = chat_repl.query_one("#chat-input", TextArea)
+            chat_input.text = "/sessions"
```
jose-compu: **Nit (pre-existing):** Setting `chat_input.text = "/sessions"` only places text in the widget — it does not trigger submission. The comment says `# Trigger submission` but nothing actually submits. This was already broken in the original `Input.value` code, so not a regression from this PR, but worth fixing while you're here. You could call `await chat_repl._submit_input("/sessions")` directly instead.

jose-compu (changes requested):
Thanks for tackling this: multiline input and cross-platform clipboard are good improvements to have.

I found two issues that would prevent the Enter-to-submit mechanism from working at all, and a couple of smaller items. Details in the inline comments.

**Blocking:**
- `event.shift` does not exist on Textual `Key` events — will raise `AttributeError`
- `TextArea._on_key` calls `event.stop()` on Enter, so the event never bubbles to `ChatRepl.on_key` — submission never triggers, Enter just inserts a newline

**Suggested approach:** subclass `TextArea` and override `_on_key` to intercept Enter before the base class consumes it.

Happy to re-review once those are addressed.

RichardTang-Aden:
Tested on Macbook, the multiple line display works. New line with shift-enter also works but display a bit weird (the extra \)

Copy and Paste acts the same on Macbook

image

RichardTang-Aden:
Copy&Paste works on windows. Great job!

Ttian18:
@RichardTang-Aden Thanks for testing on both Mac and Windows!

The `\` you're seeing is Textual's soft-wrap indicator — it shows up when text wraps to the next line in the input box. It usually appears when the terminal is narrower. Could you check if it goes away when you widen the window?

I can disable soft-wrap to remove it, but the tradeoff is lines would scroll horizontally instead. Let me know your preference!

Ttian18:
Addressed all three points from @jose-blockchain's review — ready for another look!

While testing the fixes, I discovered that Shift+Enter doesn't work in terminals that can't distinguish it from plain Enter (e.g., VS Code's integrated terminal). Filed #4565 to track this separately. Happy to work on it.

jose-compu:
thanks @Ttian18 for addressing the points.

vakrahul:
@Ttian18 really great work of you

PR #4466, opened 2026-02-11
fix(runtime): validate and create storage path on init

### LINKED ISSUE #1870: [Bug]: Runtime accepts nonexistent storage path without validation
## Describe the Bug

The `Runtime` class accepts a nonexistent path without any validation or warning, which can lead to silent failures when trying to save data later.

## To Reproduce
```python
from framework.runtime.core import Runtime
from pathlib import Path

# This succeeds without error
runtime = Runtime(Path('nonexistent'))
print('✓ Handles nonexistent path')  # Prints this!
```

## Expected Behavior

Should either:
1. Create the directory if it doesn't exist (explicit)
2. Raise an error immediately (fail fast)
3. Log a warning about creating new storage location

## Current Behavior

Silently accepts any path. Errors only occur later when trying to save data.

## Impact

- **Silent failures**: Agent thinks it's saving data but isn't
- **Debugging difficulty**: Error occurs far from the root cause
- **Data loss**: Users may think runs are being logged when they're not

## Environment

- OS: Windows 11
- Python: 3.11.9
- Framework: Latest main branch

## Proposed Solutions

**Option 1: Create directory automatically**
```python
def __init__(self, storage_path: str | Path):
    self.storage = FileStorage(storage_path)
    self.storage.path.mkdir(parents=True, exist_ok=True)
    # ...
```

**Option 2: Validate path exists**
```python
def __init__(self, storage_path: str | Path):
    path = Path(storage_path)
    if not path.exists():
        raise ValueError(f"Storage path does not exist: {path}")
    self.storage = FileStorage(storage_path)
    # ...
```

**Option 3: Log warning**
```python
def __init__(self, storage_path: str | Path):
    path = Path(storage_path)
    if not path.exists():
        logger.warning(f"Creating new storage directory: {path}")
        path.mkdir(parents=True, exist_ok=True)
    self.storage = FileStorage(storage_path)
    # ...
```

I prefer **Option 3** as it's explicit, safe, and informative.

## Additional Context

Discovered during edge case testing. Related to general error handling patterns in the framework.

I can fix this if assigned.

Emart29:
## Description
Fixes the bug where `Runtime` accepts a nonexistent storage path without validation, leading to silent failures when trying to save data later.

Fixes #1870

## Type of Change

## Changes Made
- Added path validation in `Runtime.__init__()`
- Logs a warning when the storage directory doesn't exist
- Auto-creates the directory with `parents=True` to prevent downstream errors
- Implements **Option 3** from the issue discussion (explicit, safe, informative)

## Before
```python
runtime = Runtime(Path('nonexistent'))  # Silently accepts
# ... later fails when trying to save
```

## After
```python
runtime = Runtime(Path('nonexistent'))
# WARNING: Storage path does not exist, creating: nonexistent
# Directory is created, saves work correctly
```

## Testing
- [x] Manual testing with nonexistent paths
- [x] Manual testing with existing paths
- [x] Lint passes (`ruff check`)

## Checklist

### DIFF
--- a/core/framework/runtime/core.py
+++ b/core/framework/runtime/core.py
@@ -56,6 +56,12 @@ class Runtime:
     """
 
     def __init__(self, storage_path: str | Path):
+        # Validate and create storage path if needed
+        path = Path(storage_path) if isinstance(storage_path, str) else storage_path
+        if not path.exists():
+            logger.warning(f"Storage path does not exist, creating: {path}")
+            path.mkdir(parents=True, exist_ok=True)
+
         self.storage = FileStorage(storage_path)
         self._current_run: Run | None = None
         self._current_node: str = "unknown"


Emart29:
@austin931114 @Hundao 

Implemented the fix as discussed in #1870. Ready for review.

Issue #3900, opened 2026-02-07
[RFC]: Product Requirements for Agent Evaluation System — the missing piece of the self-improvement loop


KelvinNguyen27:
## Problem Statement

Hive's core promise is self-improving agents: *"when things break, the framework captures failure data, evolves the agent through the coding agent, and redeploys"* (README, line 38). The evolution documentation describes a complete 4-stage loop: Execute, Evaluate, Diagnose, Regenerate.

However, the evaluation system — the mechanism that actually **measures** success criteria and **triggers** evolution — is largely unimplemented. The roadmap confirms this explicitly:

```
Eval System (Basic)
- [x] Test Driven - Run test case for all agent iteration
- [ ] Failure recording mechanism
- [ ] SDK for defining failure conditions
- [ ] Basic observability hooks
- [ ] User-driven log analysis

Adaptiveness
- [ ] Runtime data feedback loop
- [ ] Instant Developer Feedback for improvement
```

The **schemas are excellent** — `SuccessCriterion`, `Constraint`, `Goal.is_success()`, `OutcomeAggregator`, `Decision`/`Outcome`/`Problem`, `HybridJudge`, `LLMJudge`, `ErrorCategorizer` — all exist and are well-designed. What's missing is the integration layer that connects execution results to criterion measurement and evolution triggers.

### 1. Success criteria are defined but never evaluated

`SuccessCriterion` in `goal.py` has a `met` field and a `metric` field (supporting `output_contains`, `output_equals`, `llm_judge`, `custom`) — but nothing in the codebase sets `met` during execution. `Goal.is_success()` reads criteria that are never populated. `OutcomeAggregator._evaluate_criterion()` uses keyword matching instead of the actual metric evaluation.

### 2. No structured failure recording

The roadmap lists this as not done. `Decision` + `Outcome` + `Problem` schemas exist but aren't synthesized into actionable failure reports. There's no `FailureReport` that links unmet criteria to specific execution decisions.

### 3. Test infrastructure is disconnected from goal evaluation

Tests run via pytest subprocess (`cli.py`) — completely independent from the framework's goal/criteria system. Test results don't check success criteria. `ErrorCategorizer` classifies failures but isn't wired into any automatic evolution flow.

### 4. No evolution trigger mechanism

Zero code exists to: analyze failed test results, synthesize failure data, invoke the coding agent with an evolution prompt, apply changes, or redeploy. The `evolution_reason` and `parent_version` fields on `Goal` are never populated.

[...]

PR #4519, opened 2026-02-11
Clarify supported criterion evaluation and progress semantics

### LINKED ISSUE #3955: Make success-criterion evaluation assumptions explicit in OutcomeAggregator
## Context

While reading through `OutcomeAggregator._evaluate_criterion`, I noticed that the current implementation implicitly assumes all success criteria can be evaluated using a success-rate heuristic derived from decision outcomes.

This works for the current behavior, but the assumption is not explicit in code. As a result, when a success criterion is not met, it’s unclear *what kind of success was actually being evaluated*, and unsupported or future criterion types could be silently mis-evaluated.

---

## Proposed clarification (small & backward-compatible)

* Explicitly guard `_evaluate_criterion` to only evaluate success-rate–based criteria
* Early-return for unsupported criterion types instead of implicitly applying the heuristic
* Clarify via comments what `progress` represents (raw success rate)
* Improve evidence strings to make failures easier to understand and trace

---

## Why this helps

* Makes current evaluation behavior explicit and easier to reason about
* Prevents unsupported criteria from being silently mis-evaluated
* Improves debuggability without changing existing behavior
* Keeps future evaluation work safer to implement

---

## Scope note

This suggestion is intentionally scoped to clarifying existing behavior and does **not** attempt to implement broader evaluation or evolution mechanisms discussed elsewhere.

I’m happy to open a small PR if this aligns with the intended direction.

Rudra2637:
## Description

This PR makes the evaluation behavior of `OutcomeAggregator._evaluate_criterion` explicit by gating evaluation to supported criterion types only.

Previously, all success criteria were implicitly treated as success-rate based, which could lead to misleading `progress` and `met` values for unsupported or future criterion types. This change ensures that only `success_rate` criteria are evaluated, while others return a neutral, un-met status until proper evaluators are implemented.

This improves clarity, correctness, and observability of the evaluation system without introducing breaking changes.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #3955

Related to the broader evaluation system RFC (#3900), but scoped specifically to Phase 1 criterion evaluation semantics.

## Changes Made

- Explicitly gate criterion evaluation to `success_rate` type
- Default untyped criteria to `success_rate` for backward compatibility
- Prevent unsupported criterion types from being implicitly evaluated
- Improve evaluation evidence clarity by including decision ID and result

## Testing

## Checklist

- [ ] I have made corresponding changes to the documentation (not required for this change)
- [ ] I have added tests that prove my fix is effective (follow-up work)

N/A

### DIFF
--- a/core/framework/runtime/outcome_aggregator.py
+++ b/core/framework/runtime/outcome_aggregator.py
@@ -312,10 +312,10 @@ class OutcomeAggregator:
 
     async def _evaluate_criterion(self, criterion: Any) -> CriterionStatus:
         """
-        Evaluate a single success criterion.
+            Evaluate a single success criterion.
 
-        This is a heuristic evaluation based on decision outcomes.
-        More sophisticated evaluation can be added per criterion type.
+            This is a heuristic evaluation based on decision outcomes.
+            More sophisticated evaluation can be added per criterion type.
         """
         status = CriterionStatus(
             criterion_id=criterion.id,
@@ -324,6 +324,11 @@ class OutcomeAggregator:
             progress=0.0,
             evidence=[],
         )
+
+        # Guard: only apply this heuristic to success-rate criteria
+        criterion_type = getattr(criterion, "type", "success_rate")
+        if criterion_type != "success_rate":
+            return status
 
         # Get relevant decisions (those mentioning this criterion or related intents)
         relevant_decisions = [
@@ -348,7 +353,6 @@ class OutcomeAggregator:
             # Add evidence
             for d in relevant_decisions[:5]:  # Limit evidence
                 if d.outcome:
-                    # Add more explicit evidence for observability
                     evidence = (
                         f"decision_id={d.decision.id}, "
                         f"intent={d.decision.intent}, "
@@ -363,7 +367,6 @@ class OutcomeAggregator:
                 target_value = float(target.rstrip("%")) / 100
                 status.met = status.progress >= target_value
             else:
-                # For non-percentage targets, consider met if progress > 0.8
                 status.met = status.progress >= 0.8
         except (ValueError, AttributeError):
             status.met = status.progress >= 0.8
--- a/core/framework/runtime/outcome_aggregator.py
+++ b/core/framework/runtime/outcome_aggregator.py
@@ -366,6 +366,7 @@ class OutcomeAggregator:
                 target_value = float(target.rstrip("%")) / 100
                 status.met = status.progress >= target_value
             else:
+                # For non-percentage targets, consider met if progress > 0.8
                 status.met = status.progress >= 0.8
         except (ValueError, AttributeError):
             status.met = status.progress >= 0.8
--- a/core/framework/runtime/outcome_aggregator.py
+++ b/core/framework/runtime/outcome_aggregator.py
@@ -325,6 +325,12 @@ class OutcomeAggregator:
             evidence=[],
         )
 
+        # 🔹 NEW: Explicitly guard supported evaluation behavior
+        # Currently, only success-rate–based evaluation is supported.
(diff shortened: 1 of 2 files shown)


Rudra2637:
Hi @RichardTang-Aden, I’ve opened a PR addressing issue #3955.
Would appreciate your review whenever you have time. Thanks!

bryanadenhq (changes requested):
The guard checks a field that doesn't exist, SuccessCriterion has no type field. The getattr always return None, so the guard never fires. Additionally remove emojis from docs.

Rudra2637:
Thanks for the feedback.
I’ve removed the invalid type guard and cleaned up the comments (removed emojis).
Please take another look.

Antiarin:
Hey @Rudra2637 , the guard was really the main point of this PR, so with it removed we're left with a comment and a string formatting change, which isn't really enough to merge on its own.

The fix is actually straightforward though. The guard didn't work because SuccessCriterion doesn't have a type field yet. 
If you add one to the model in core/framework/graph/goal.py with default="success_rate", it's fully backward compatible (the model already has extra="allow"), and then the guard in _evaluate_criterion actually fires for non-success-rate criteria instead of silently evaluating them with the wrong heuristic.

That also gives us a clean separation between type (what kind of runtime evaluation) and metric (how to measure in tests), which is what we need since real agents already use custom metric values like intent_accuracy or tool_success_rate.

Rudra2637:
That’s a good point @Antiarin,thank you for pointing it out. Let’s see what the maintainers think, and we can update the PR accordingly based on their guidance.

bryanadenhq:
Hey @Rudra2637, @Antiarin is correct in this case, without the guard, it is just a comment and string format change. If you implement it the way that is described in his comment, it should work with no problems. 

@Antiarin, thanks for your great comment and suggestion!

core/framework/runtime/outcome_aggregator.py
```diff
@@ -312,10 +312,10 @@ async def evaluate_goal_progress(self) -> dict[str, Any]:
 
     async def _evaluate_criterion(self, criterion: Any) -> CriterionStatus:
         """
-        Evaluate a single success criterion.
+            Evaluate a single success criterion.
 
-        This is a heuristic evaluation based on decision outcomes.
-        More sophisticated evaluation can be added per criterion type.
+            This is a heuristic evaluation based on decision outcomes.
+            More sophisticated evaluation can be added per criterion type.
```
bryanadenhq: Please remove this additional indentation

bryanadenhq (changes requested):
Hi @Rudra2637, please address this minor change and then I can merge this pr. Thanks for your contribution!

core/framework/runtime/outcome_aggregator.py
```diff
@@ -358,7 +367,6 @@ async def _evaluate_criterion(self, criterion: Any) -> CriterionStatus:
                 target_value = float(target.rstrip("%")) / 100
                 status.met = status.progress >= target_value
             else:
-                # For non-percentage targets, consider met if progress > 0.8
```
Antiarin, line 361: Hey could you please add it back? Without it this looks like an arbitrary magic number

bryanadenhq: second this!

Rudra2637: Good point .I’ll add the comment back for clarity. Thanks!

core/framework/runtime/outcome_aggregator.py
Rudra2637: Hi @bryanadenhq, I’ve made the requested changes.  Thanks!

PR #4595, opened 2026-02-12
micro-fix: pre-init node_spec and guard parallel None.name in executor

### LINKED ISSUE #4514: [Bug]: UnboundLocalError masks real exceptions in GraphExecutor exception handler
## Description

In `core/framework/graph/executor.py`, the outer `except Exception` handler (line ~1042) references `node_spec` (line ~1057), but `node_spec` is first assigned inside the `while` loop (line ~474). If an exception occurs before the first loop iteration completes, `node_spec` is unbound.

## Minimal Reproduction

Any `GraphExecutor.execute()` call that raises before line 474 (e.g., checkpoint store failure during pause handling, or an exception in the first iteration before `graph.get_node()`).

## Expected Behavior

The original exception is captured and returned in `ExecutionResult`.

## Actual Behavior

`UnboundLocalError: cannot access local variable 'node_spec'` is raised **inside** the except handler, masking the original error. Debugging becomes impossible because the real cause is lost.

### Additional: parallel branch failure also crashes

In the same file, `failed_names = [graph.get_node(b.node_id).name for b in failed_branches]` (line ~1655) calls `.name` on the result of `get_node()` which may return `None` for invalid branch node IDs, causing `AttributeError`.

## Proposed Fix

1. Initialize `node_spec = None` before the `try` block.
2. Guard `graph.get_node(b.node_id)` against `None` in the parallel branch failure reporting.

## Environment

- File: `core/framework/graph/executor.py`
- Lines: ~430 (init), ~1057 (handler), ~1655 (parallel)

kostasuser01gr:
## Summary
- Pre-initialise `node_spec = None` before the try block in `GraphExecutor.execute()` so the except handler never hits `UnboundLocalError`
- Guard `graph.get_node()` returning `None` in parallel branch failure reporting to prevent `AttributeError`
- Add 3 regression tests (253 lines) covering both findings

## Why
**Finding 1:** If an exception occurs in `execute()` before `node_spec` is assigned inside the while-loop (e.g. checkpoint store failure, graph corruption), the except handler at line ~1057 references `node_spec` which is unbound. This raises `UnboundLocalError`, masking the real exception.

**Finding 2:** In `execute_single_branch()`, `failed_names` is built via `graph.get_node(b.node_id).name` — but `get_node()` can return `None` for dynamically removed or corrupted nodes, causing `AttributeError: 'NoneType' has no attribute 'name'`.

## Changes
- `core/framework/graph/executor.py`: 2 surgical fixes (5 lines changed)
- `core/tests/test_executor_exception_safety.py`: 253-line regression test file

## Validation
```
$ uv run ruff check framework/graph/executor.py        → All checks passed!
$ uv run ruff format --check framework/graph/executor.py → 1 file already formatted
$ uv run pytest tests/test_executor_exception_safety.py tests/test_graph_executor.py -v → 8 passed
```

Fixes #4514

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -427,6 +427,7 @@ class GraphExecutor:
                 data_dir=str(self._storage_path / "data"),
             )
 
+        node_spec = None  # Pre-initialise so the except handler never hits UnboundLocalError
         try:
             while steps < graph.max_steps:
                 steps += 1
@@ -1652,7 +1653,9 @@ class GraphExecutor:
 
         # Handle failures based on config
         if failed_branches:
-            failed_names = [graph.get_node(b.node_id).name for b in failed_branches]
+            failed_names = [
+                n.name if (n := graph.get_node(b.node_id)) else b.node_id for b in failed_branches
+            ]
             if self._parallel_config.on_branch_failure == "fail_all":
                 raise RuntimeError(f"Parallel execution failed: branches {failed_names} failed")
             elif self._parallel_config.on_branch_failure == "continue_others":
--- a/core/tests/test_executor_exception_safety.py
+++ b/core/tests/test_executor_exception_safety.py
@@ -1,0 +1,253 @@
+"""
+Regression tests for executor exception-safety fixes.
+
+Covers:
+  - node_spec pre-initialisation so the except handler never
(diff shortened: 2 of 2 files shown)


RichardTang-Aden:

PR #4596, opened 2026-02-12
micro-fix: call end_run() in CancelledError handler to close trace context

### LINKED ISSUE #4515: [Bug]: CancelledError handler in ExecutionStream skips end_run() — run data silently lost
## Description

In `core/framework/runtime/execution_stream.py`, when `asyncio.CancelledError` is caught in `_run_execution()` (line ~458), `runtime_adapter.end_run()` is never called, even though `runtime_adapter.start_run()` was called earlier (line ~365).

The normal completion path calls `end_run()` (line ~424), and the `except Exception` handler calls it (line ~514), but the `CancelledError` handler omits it entirely.

## Expected Behavior

When execution is cancelled, `end_run()` should be called with `success=False` and a cancellation narrative, so the run is persisted for observability.

## Actual Behavior

The run started by `start_run()` is never closed. All decisions and metrics recorded during the run are silently garbage-collected. Observability/tracing data is lost.

## Proposed Fix

Add `end_run()` call to the CancelledError handler, matching the pattern used in the `except Exception` handler:

```python
try:
    runtime_adapter.end_run(
        success=False,
        narrative="Execution cancelled",
        output_data=result.output if has_result else {},
    )
except Exception:
    pass
```

## Environment

- File: `core/framework/runtime/execution_stream.py`
- Lines: ~365 (start_run), ~458 (CancelledError handler), ~514 (Exception handler with end_run)

kostasuser01gr:
## Summary

The `CancelledError` handler in `ExecutionStream._run_execution()` calls `start_run()` (via `StreamRuntimeAdapter`) but never calls `end_run()`, leaving the trace context open and causing observability data for the cancelled run to be silently lost.

Fixes #4515

### `core/framework/runtime/execution_stream.py`
- Added `runtime_adapter.end_run(success=False, narrative="Execution cancelled", output_data={})` in the `CancelledError` handler, wrapped in a defensive `try/except` so an `end_run()` failure cannot mask the cancellation itself.

### `core/tests/test_execution_stream_cancelled_end_run.py` (new)
- Regression test that cancels a running execution and asserts `end_run()` is invoked with `success=False`.

## Why

Without `end_run()`, any observability backend (tracing, logging, metrics) that relies on the run lifecycle will never receive the "run finished" signal for cancelled executions. This is a data‑loss bug — the run appears to start but never completes in the trace.

## Validation

```
$ uv run ruff check framework/runtime/execution_stream.py   # All checks passed!
$ uv run ruff format --check framework/runtime/execution_stream.py  # already formatted
$ uv run pytest tests/test_execution_stream_cancelled_end_run.py -v  # 1 passed in 1.46s
```

### DIFF
--- a/core/framework/runtime/execution_stream.py
+++ b/core/framework/runtime/execution_stream.py
@@ -491,6 +491,17 @@ class ExecutionStream:
                 else:
                     await self._write_session_state(execution_id, ctx, error="Execution cancelled")
 
+                # End run so the trace context opened by start_run() is closed.
+                # Without this, observability data for the run is silently lost.
+                try:
+                    runtime_adapter.end_run(
+                        success=False,
+                        narrative="Execution cancelled",
+                        output_data={},
+                    )
+                except Exception:
+                    pass  # Don't let end_run errors mask the cancellation
+
                 # Don't re-raise - we've handled it and saved state
 
             except Exception as e:
--- a/core/tests/test_execution_stream_cancelled_end_run.py
+++ b/core/tests/test_execution_stream_cancelled_end_run.py
@@ -1,0 +1,138 @@
+"""
+Regression test for CancelledError handler missing end_run() call.
+
+The execution_stream.py CancelledError handler calls start_run() (via
+StreamRuntimeAdapter) but never called end_run(), causing observability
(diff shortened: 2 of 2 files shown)


RichardTang-Aden:

PR #4632, opened 2026-02-12
micro-fix: truncate view_file content by bytes instead of characters

kostasuser01gr:
## Summary

Fixes #4631 — the `view_file` tool's truncation logic sliced by **characters** instead of **bytes**, so multi-byte UTF-8 content could exceed the `max_size` byte limit.

## Problem

```python
if len(content.encode(encoding)) > max_size:   # ← byte check ✓
    content = content[:max_size]                # ← character slice ✗
```

`max_size` is documented and checked in bytes, but `content[:max_size]` takes `max_size` **characters**. For multi-byte encodings (any non-ASCII UTF-8), one character can be 2-4 bytes, so the truncated output can be up to 4× larger than intended.

## Fix

```python
content = content.encode(encoding)[:max_size].decode(encoding, errors="ignore")
```

Encode → slice at the byte boundary → decode back, dropping any partial character at the cut point. The result is always ≤ `max_size` bytes.

## Scope

| File | Change |
|------|--------|
| `tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py` | 1 line modified |

## Validation

- [x] `ruff check` — clean
- [x] `ruff format --check` — clean
- [x] `pytest tests/ -x -q` — **683 passed**, 51 skipped, 0 failed
- [x] Pure-ASCII files: behavior identical (1 char = 1 byte)
- [x] Multi-byte files: output now respects byte limit

### DIFF
--- a/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py
+++ b/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py
@@ -60,7 +60,7 @@ def register_tools(mcp: FastMCP) -> None:
                 content = f.read()
 
             if len(content.encode(encoding)) > max_size:
-                content = content[:max_size]
+                content = content.encode(encoding)[:max_size].decode(encoding, errors="ignore")
                 content += "\n\n[... Content truncated due to size limit ...]"
 
             return {


RichardTang-Aden:

PR #4695, opened 2026-02-13
fix: remove duplicate CredentialStoreAdapter in credentials __all__

jamesrziggy:
Fixes #4629. Removed duplicate CredentialStoreAdapter entry and stale CredentialManager comment from __all__ in tools/src/aden_tools/credentials/__init__.py. Found and fixed by Qwen3-14B running locally on an RTX 3090.


jamesrziggy:
?

PR #4697, opened 2026-02-13
fix: remove vestigial duplicate Step 3 label in quickstart.sh

### LINKED ISSUE #4603: [Bug]: quickstart.sh has duplicate Step 3 labels (vestigial LLM config header)
### Description

`quickstart.sh` has a duplicate "Step 3" label that confuses the onboarding flow.

**Lines 219-225** print `"Step 3: Configuring LLM provider..."` but contain **no actual LLM configuration logic** — the section header was likely left behind when the LLM config logic was removed or moved.

Immediately after, **lines 226-231** start another `"Step 3: Verifying Python imports..."` which is the real step content.

This means the user sees two consecutive "Step 3" messages in the terminal output, making the quickstart appear broken or confusing.

### Steps to Reproduce

1. Run `./quickstart.sh`
2. Observe two "Step 3" lines printed back-to-back:
   - `Step 3: Configuring LLM provider...`
   - `Step 3: Verifying Python imports...`

### Expected Behavior

Each step should have a unique sequential number. The vestigial "Configure LLM" header (which does nothing) should be removed so "Verify Python Imports" is the sole Step 3.

### Actual Behavior

Two consecutive step headers both labelled "Step 3".

### Suggested Fix

Remove the dead section (comment block + echo) for "Step 3: Configure LLM API Key" at lines 219-225. The import verification at lines 226+ remains as Step 3. No renumbering of subsequent steps needed since the dead section doesn't correspond to any actual step logic.

jamesrziggy:
Fixes #4603. Removed dead 'Step 3: Configuring LLM provider...' section that had no logic. Causes duplicate Step 3 messages in onboarding flow. Found and fixed by Qwen3-14B running locally on an RTX 3090.


jamesrziggy:
?

Issue #4691, opened 2026-02-13
[Bug]: CLI wrapper forces running from ~/.local/bin instead of project root


zarashafiq:
## Describe the bug
After running ./quickstart.sh on Windows (Git Bash), the installed `hive` command cannot be used from the repo folder.

`hive --help` fails from the Hive repo root, but `uv run hive --help` works.

## To reproduce
1. Windows + Git Bash
2. git clone https://github.com/adenhq/hive.git
3. cd hive
4. ./quickstart.sh (complete setup)
5. source /c/Users/<user>/.profile
6. Run: hive --help

## Actual behavior
Error says `hive must be run from the project directory`, and points to `~/.local/bin` as the expected directory.

## Expected behavior
`hive --help` should work when run from the Hive repo root after quickstart.

## Workaround
From the repo root:
`uv run hive --help` works.

## Environment
- OS: Windows
- Shell: Git Bash (MINGW64)
- Git: 2.49.0.windows.1
- Python: 3.13.2

VD-17:
Hi, I would like to work on this issue.

I'm on Windows and can reproduce this bug locally. I'll investigate:
1. Where the "must be run from the project directory" check is
2. Why path detection fails on Windows (Git Bash)

Could you assign this to me?

Patlu475:
Hi! I'd like to work on this issue.

The CLI wrapper forcing execution from ~/.local/bin instead of project root is problematic for development workflows. I can help by:
- Investigating the CLI wrapper behavior
- Fixing the path resolution logic
- Ensuring proper working directory handling
- Adding tests to prevent regression

Happy to work on this once assigned!

PR #4705, opened 2026-02-13
micro-fix(cli): remove incorrect project directory check in wrapper script

zarashafiq:
## Description
Removes incorrect directory check in hive CLI wrapper that forces execution from ~/.local/bin instead of the project root.

## Type of Change

## Related Issues
Related to #4691

## Testing
Manual test:
- Verified ./hive --help works from project root


zarashafiq:
On Windows Git Bash, quickstart Step 7 says it creates a symlink, but ~/.local/bin/hive ends up as a regular file, so the global hive stays stale. Might need a Windows-safe symlink strategy or fallback.

Issue #2849, opened 2026-01-31
[Coding Agent][Integration]: Support Kimi Code as the coding agent to generate agents


vincentjiang777:
## Problem Statement

Cost Efficiency: Developers need alternatives to expensive proprietary models for high-volume agent loops.

Reasoning Capabilities: Kimi K2.5 features a "Thinking Mode" (similar to OpenAI o1 or DeepSeek R1) which produces a reasoning_content field. Currently, standard OpenAI-compatible integrations may drop or ignore this field, losing valuable chain-of-thought data that hive agents could use for self-correction.

Multimodality: K2.5 supports native image and video input, which aligns with hive's goal of building capable agents.

## Proposed Solution

Provider Integration: Add moonshot or kimi as a recognized provider in core/framework/llm/.

Thinking Mode Support: Ensure the framework captures the reasoning_content field from the API response when "Thinking Mode" is active.

Visual Input: Verify standard image/video payload compatibility with Kimi's API format.
## Alternatives Considered

A description of any alternative solutions or features you've considered.

## Additional Context

Add any other context, mockups, or screenshots about the feature request here.

## Implementation Ideas

If you have ideas about how this could be implemented, share them here.

vakrahul:
@vincentjiang777 
with kimi hive agent can improve the effiency that build the model  and also reduce the  thinking of same type of erros and we can make less api calls and Faster convergence 
furture model we can fine-tune model

To support Kimi Code as a first-class coding agent, we can extend the existing LiteLLMProvider to handle Moonshot's specific API behavior.

1. Provider Configuration & Mapping
We can map kimi to the Moonshot endpoint. Since it is OpenAI-compatible, litellm handles the heavy lifting, but we need to ensure the reasoning_content (beta) is preserved.

Python
# core/framework/llm/provider.py

PROVIDER_CONFIGS = {
    # ... existing ...
    "kimi": {
        "model_prefix": "moonshot-v1", 
        "api_base": "https://api.moonshot.cn/v1",
        "api_key_env": "MOONSHOT_API_KEY",
        "supports_reasoning": True  # Flag to enable extraction
    }
}
2. Handling reasoning_content (The Critical Piece)
Standard OpenAI parsers often drop extra fields. We need to patch the complete method in core/framework/llm/litellm.py to capture the "Thinking Process" and expose it to the agent (e.g., for self-correction loops).

Python
# Sketch for core/framework/llm/litellm.py

def complete(self, messages, **kwargs):
    # ... 
[...]

PR #4749, opened 2026-02-13
fix(micro-fix): Add timeout protection for MCP STDIO tool calls

hemanth-chowdary-07:
## Summary
Adds timeout protection for MCP STDIO tool calls to prevent indefinite blocking when tool servers hang.

## Changes (micro-fix: <20 lines)
- Add `tool_timeout` config (default: 30s) to `MCPServerConfig`
- Wrap STDIO tool calls with `asyncio.wait_for()`
- Log timeout errors for debugging

## Diff Stats
- 13 additions, 3 deletions

Fixes #3440

### DIFF
--- a/core/framework/runner/mcp_client.py
+++ b/core/framework/runner/mcp_client.py
@@ -15,17 +15,6 @@ import httpx
 logger = logging.getLogger(__name__)
 
 
-class MCPTimeoutError(Exception):
-    """Raised when an MCP tool call times out."""
-
-    def __init__(self, tool_name: str, timeout: float, message: str = None):
-        self.tool_name = tool_name
-        self.timeout = timeout
-        self.error_type = "timeout"
-        self.message = message or f"Tool '{tool_name}' timed out after {timeout} seconds"
-        super().__init__(self.message)
-
-
 @dataclass
 class MCPServerConfig:
     """Configuration for an MCP server connection."""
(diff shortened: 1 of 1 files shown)


RichardTang-Aden:
Not legit change. It will break bigger tool calls

hemanth-chowdary-07:
Thanks for the feedback @RichardTang-Aden. You're right.
## On the Timeout PR
I underestimated the duration of legitimate tool operations. A fixed 30s timeout would break longer-running tools like code generation, large file processing, or slow API calls.
Better approaches I should have considered:
- **No timeout by default** (opt-in only)
- **Per-tool timeout config** - users set timeout based on expected duration
- **Configurable at call time** - pass timeout as parameter to `call_tool()`

I also removed the `isError` check in my changes - that was a mistake on my part.

Would you be open to a revised approach with opt-in timeouts? Or is this better solved at the agent level rather than MCP client level? Happy to take direction.

## Separate Question: Code Quality Tooling

I noticed Vincent's issue #2849 around coding agents, and I've been thinking about code quality tooling for Hive.

I built [RefactorAI](https://refactor-ai-eight.vercel.app) - a code analysis tool that currently does:
- Basic code smell detection (long methods, deep nesting, magic numbers, god class)
- Complexity metrics (line count, parameter count, nesting depth)
- LLM-powered refactoring suggestions via Groq

It's not advanced static analysis like SonarQube - it's lightweight metrics + LLM suggestions.

**My question:** Is there interest in code quality tooling for Hive agents? If so, what capabilities would actually be useful?

For example:
- Basic metrics for PR review workflows?
- Style/formatting checks?
- Something more sophisticated?

I'd rather understand what's needed before proposing something that doesn't fit.

Appreciate your time.

PR #5166, opened 2026-02-20
feat: add Revenue Leak Detector agent with HubSpot CRM + Telegram + Gmail integration

Siddharth2624:
Fixes #5153
## Overview

Adds a new `revenue_leak_detector` agent template — an autonomous business health monitor that continuously scans a CRM pipeline, detects revenue leak patterns, sends Telegram alerts, dispatches Gmail follow-up emails to ghosted contacts, and halts automatically when severity reaches critical.

---

## Agent Architecture
monitor ──► analyze ──► notify ──► followup
│
◄───────────────────────────┘
(loops while halt != true)

| Node | Type | What it does |
|---|---|---|
| `monitor` | `event_loop` | Calls `scan_pipeline()` — fetches open deals from HubSpot CRM |
| `analyze` | `event_loop` | Calls `detect_revenue_leaks()` — classifies leaks, calculates severity |
| `notify` | `event_loop` | Calls `send_revenue_alert()` — prints structured report + sends Telegram alert |
| `followup` | `event_loop` | Calls `send_followup_emails()` — sends Gmail re-engagement emails to GHOSTED contacts |

Loop exits automatically when `severity = critical` (`halt = True`).

---

## What It Detects

| Pattern | Trigger | Risk |
|---|---|---|
| **GHOSTED** | Deal silent for 21+ days | Lost deal value |
| **STALLED** | Deal stuck in same stage 10–20 days | Slow pipeline velocity |
| **OVERDUE_PAYMENT** | Invoice unpaid past due date | Cash flow leak |
| **CHURN_RISK** | 3+ open support escalations | Customer churn |

---

### HubSpot CRM (real data tested ✅)
- Fetches open deals via CRM v3 API using `hs_lastmodifieddate` to calculate `days_inactive`
- Batch-fetches primary contact email per deal via Associations API
- Filters out `closedwon` / `closedlost` deals automatically
- Falls back to empty snapshot if `HUBSPOT_API_KEY` is not set

> **Note on testing**: During development, because newly created HubSpot deals always have `days_inactive = 0` (their `hs_lastmodifieddate` is today), the GHOSTED threshold was temporarily lowered to `>= 0` for demo purposes. Demo deals (DEMO-S01, DEMO-S02, DEMO-S03) were also injected to showcase all 4 leak types visually. **Both of these have been reverted/removed in this PR** — the agent now runs on clean production logic with the 21-day threshold.

### Telegram (real delivery tested ✅)
- Sends HTML-formatted alert messages with severity emoji, leak breakdown, and action items
- Falls back to console output when env vars not set

### Gmail SMTP (configured ✅)
- Sends re-engagement emails to every GHOSTED contact with a personalized body
- Dry-run preview printed to console when credentials not set

---

## Environment Variables Required

| Variable | Purpose |
|---|---|
| `HUBSPOT_API_KEY` | HubSpot Private App token (scopes: `crm.objects.deals.read`, `crm.objects.contacts.read`) |
| `TELEGRAM_BOT_TOKEN` | Token from @BotFather |
| `TELEGRAM_CHAT_ID` | Chat/group ID where alerts are delivered |
| `GMAIL_USER` | Gmail address used to send follow-up emails |
| `GMAIL_APP_PASSWORD` | Gmail App Password (requires 2FA enabled on account) |

All integrations are **optional** — the agent runs fully offline without any credentials (console output only).

---

## Running

```bash
# Console-only (no credentials needed)
uv run hive run examples/templates/revenue_leak_detector

# With TUI
uv run hive run examples/templates/revenue_leak_detector --tui

# Full integrations
export HUBSPOT_API_KEY="..."
export TELEGRAM_BOT_TOKEN="..."
export TELEGRAM_CHAT_ID="..."
export GMAIL_USER="..."
export GMAIL_APP_PASSWORD="..."
uv run hive run examples/templates/revenue_leak_detector --tui

Working Video of agent : https://drive.google.com/file/d/1_CqNK-SVk8j7ydqF9hEk8W8TqV99AaPL/view?usp=sharing

### CODE REVIEW

File: examples/templates/revenue_leak_detector/__init__.py [hunk 1]
```diff
@@ -0,0 +1,7 @@
+from .agent import goal, nodes, edges
```
bryanadenhq: Missing agent class, default_agnet instance, and config exports. All other template export these. WIthout an agent calss, this template can't be loaded or run the same way as others.

examples/templates/revenue_leak_detector/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+
+from framework.runner import AgentRunner
+
+
+async def main():
+    print("=" * 64)
+    print("  💰  REVENUE LEAK DETECTOR — Starting Agent")
+    print("=" * 64)
+
+    runner = AgentRunner.load("examples/templates/revenue_leak_detector")
```
bryanadenhq: Other templates don't use AgentRunner.load() in main.py, they all import and instantiate their own agent class in init.py per my previous comment.

examples/templates/revenue_leak_detector/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+    runner._setup()
```
bryanadenhq: Private class, shouldn't be called directly

examples/templates/revenue_leak_detector/tools.py
```diff
@@ -0,0 +1,656 @@
...
+"""
+
+import os
+
+
+# ---------------------------------------------------------------------------
+# Minimal @tool decorator (framework convention)
+# ---------------------------------------------------------------------------
+
+def tool(func):
```
bryanadenhq: We should be registering tools using the TOOLS dict + tool_executor() pattern.

examples/templates/revenue_leak_detector/tools.py
```diff
@@ -0,0 +1,656 @@
...
+    """Marks a function for auto-discovery by ToolRegistry."""
+    func._tool_metadata = {"name": func.__name__}
+    return func
+
+
+# ---------------------------------------------------------------------------
+# Shared in-process state — survives across node calls within the same run
+_CURRENT_CYCLE_DATA: dict = {}
```
bryanadenhq: Globals are not thread-sage and not session-isolated.

examples/templates/revenue_leak_detector/README.md
```diff
@@ -0,0 +1,92 @@
...
+| **STALLED** | Deal stuck in same stage 10-20 days | Slow pipeline velocity |
+| **OVERDUE_PAYMENT** | Invoice unpaid after due date | Cash flow leak |
+| **CHURN_RISK** | 3+ unresolved support escalations | Customer churn |
+
+---
+
+## Agent Graph
+
+```
+monitor ──► analyze ──► notify
```
bryanadenhq: readme no longer up to date

bryanadenhq (changes requested):
Hi @Siddharth2624, I just left some comments for you to review. Thanks for your contribution!

examples/templates/revenue_leak_detector/README.md
Siddharth2624: Fixed all. Please check if anything left.

examples/templates/revenue_leak_detector/config.py
```diff
@@ -0,0 +1,25 @@
+"""Runtime configuration for Revenue Leak Detector Agent."""
```
bryanadenhq: Missing mcp_server.json

examples/templates/revenue_leak_detector/agent.py
```diff
@@ -0,0 +1,291 @@
+"""Revenue Leak Detector Agent — LLM-driven event_loop nodes.
```
bryanadenhq: Missing success_criteria and constraints.

examples/templates/revenue_leak_detector/nodes/__init__.py
```diff
@@ -0,0 +1,87 @@
+"""Node definitions for Revenue Leak Detector Agent."""
+
+from framework.graph import NodeSpec
+
+monitor_node = NodeSpec(
```
bryanadenhq: Missing explicit max_node_visits.

examples/templates/hubspot_revenue_leak_detector/__main__.py
```diff
@@ -0,0 +1,164 @@
...
+        sys.exit(1)
+
+    from pathlib import Path
+
+    from framework.llm import LiteLLMProvider
+    from framework.runner.tool_registry import ToolRegistry
+    from framework.runtime.agent_runtime import create_agent_runtime
+    from framework.runtime.execution_stream import EntryPointSpec
+
+    async def run_with_tui():
```
bryanadenhq, line 76: TUI command duplicates setup logic from agent._setup() and skips CheckpointConfig. Use agent.start() instead of rebuilding the runtime inline.

bryanadenhq (changes requested):

examples/templates/hubspot_revenue_leak_detector/__main__.py
Siddharth2624: Fixed all the issues. Please let me know if any issue left to resolve.

examples/templates/revenue_leak_detector/tools.py
```diff
@@ -0,0 +1,753 @@
...
+import os
+from typing import Any
+
+from framework.llm.provider import Tool, ToolUse, ToolResult
+
+
+# ---------------------------------------------------------------------------
+# Session-isolated in-process state (contextvars — thread + session safe)
+# ---------------------------------------------------------------------------
+_cycle_data_var: contextvars.ContextVar[dict] = contextvars.ContextVar(
```
bryanadenhq: Mutable defaults ({}, []) in ContextVar are shared across all contexts that never call .set(). If two sessions run concurrently and one hasn't called scan_pipeline yet, it reads/mutates the same dict/list as the other. Drop the default= and initialize a fresh dict/list in _scan_pipeline instead.

bryanadenhq (changes requested):

examples/templates/revenue_leak_detector/tools.py
Siddharth2624: Fixed.

RichardTang-Aden:
@Siddharth2624 

I tested it with no credentials entered. The flow completes, but if I were a new user, I would be very confused about what this agent actually does.

From what I can see, the value of this agent relies on the HubSpot integration and can’t really function on its own. In that case, HubSpot credentials should be required (and it should use the official HubSpot MCP tools in aden_tools), similar to how other integrations are handled.

Also, to help users better understand the purpose of this agent, consider renaming it to something like “HubSpot Revenue Leak Detector” or similar — that would make its value proposition much clearer.

Siddharth2624:
@RichardTang-Aden 
Are you asking for just the rename (e.g., "HubSpot Revenue Leak Detector"), or are there other changes you'd like?

I want to emphasize that I'm looking to contribute this as an agent, not as another integration. The agent should leverage HubSpot's existing MCP tools (from aden_tools) as part of its workflow, not create a new integration layer. The agent itself should be the value-add — analyzing HubSpot data to detect revenue leaks.

Would love your thoughts on the direction here!

RichardTang-Aden:
I didn't find the hubspot tools getting used at all. It is suppose to guide users through a credential log in

Siddharth2624:
Summary of Fixes
🔧 [health_check.py]
Fixed F811 / F601 (ruff): Removed duplicate standalone TelegramHealthChecker class that was introduced during rebase — upstream already had a proper TelegramHealthChecker(BaseHttpHealthChecker) subclass at line 973 with correct ok-field parsing
Fixed ok-field parsing: TelegramHealthChecker.check() now correctly reads response.json()["ok"] — returns valid=False when ok is False, and includes the bot username in the success message ("Telegram bot token valid (bot: @<username>)")
Removed duplicate "telegram" registry entry from HEALTH_CHECKERS dict (F601)
🔧 [test_health_checks.py]
Added "telegram" to the expected keys set in test_all_expected_checkers_registered — was failing because upstream added the checker to the registry but the test didn't include it
🔧 [tools.py]
Removed (or "gmail") from prepare_followup_emails tool description — Gmail is not used, Resend is the only email provider
Fixed E741 ambiguous variable name l in three generator expressions (renamed to lead)
🔧 [__init__.py]
Removed Gmail fallback from followup_node system prompt — the agent was being instructed to retry with Gmail if Resend failed, which is incorrect
Replaced with log-and-skip behavior on send_email failure
🔧 [mcp_servers.json]
Corrected server description: removed "Gmail", replaced with "Email (Resend)" to accurately reflect the tools available
🔧 [checkpoint_store.py]
Fixed E501 line-too-long by splitting an overly long line into two

examples/templates/hubspot_revenue_leak_detector/agent.py
```diff
@@ -0,0 +1,430 @@
...
+# This agent requires Resend for follow-up emails.
+# The global CredentialSpec has required=False (optional for other agents),
+# so we override it here at module load time so AgentRunner / TUI enforces it.
+try:
+    from aden_tools.credentials import CREDENTIAL_SPECS as _CRED_SPECS
+
+    if "resend" in _CRED_SPECS:
+        _CRED_SPECS["resend"].required = True
+except ImportError:
+    pass
```
bryanadenhq: This block mutates the global CREDENTIAL_SPECS["resend"].required to True, but it's already True both explicitly in tools/src/aden_tools/credentials/email.py and by default in the CredentialSpec dataclass. The comment claiming "The global CredentialSpec has required=False" was resolved in another branch and is currently fixed in main. You can just delete this section

examples/templates/hubspot_revenue_leak_detector/nodes/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+    name="Monitor",
+    description=(
+        "Fetch all open HubSpot deals via MCP tools, assemble deal objects "
+        "with days_inactive and contact email, then call scan_pipeline to store them."
+    ),
+    node_type="event_loop",
+    client_facing=False,
+    max_node_visits=0,
+    input_keys=["cycle"],
+    output_keys=["cycle", "deals_scanned", "overdue_invoices", "support_escalations"],
```
bryanadenhq: overdue_invoices and support_escalations are dead output keys, these are hardcoded to "0" in the monitor node's system prompt, always returned as 0 from scan_pipeline() in tools.py, passed through to the analyze node as input keys, and never used. Remove them from output_keys, input_keys, and the system prompts, or clearly document them as unimplemented placeholders.

examples/templates/hubspot_revenue_leak_detector/tools.py
```diff
@@ -0,0 +1,735 @@
...
+    "closedlost": "Closed Lost",
+}
+
+
+# ---------------------------------------------------------------------------
+# Telegram chat_id auto-fetch helper
+# ---------------------------------------------------------------------------
+
+
+def _auto_fetch_telegram_chat_id() -> str:
```
bryanadenhq: _auto_fetch_telegram_chat_id() makes an uncached HTTP call every cycle, when TELEGRAM_CHAT_ID is not set, this calls httpx.get("...getUpdates") on every single cycle. Cache the result in a module-level variable after the first successful fetch.

core/framework/server/routes_events.py
```diff
@@ -106,7 +106,7 @@ async def on_event(event) -> None:
             except asyncio.QueueFull:
                 logger.warning(
                     "SSE client queue full on critical event; disconnecting session='%s'",
-                    session.id
+                    session.id,
```
bryanadenhq, line 113: unrelated change, please remove

bryanadenhq (changes requested):
Hey Siddharth, I just left some comments for you. Thanks!

Siddharth2624:
Hey @bryanadenhq — you flagged the trailing comma on [session.id] in [routes_events.py] as an unrelated change and asked me to remove it. I did revert it, but [ruff format] then flagged that file as needing reformatting (it wants the trailing comma there for multi-line function calls). I ran [ruff format] to auto-fix it and it added the comma back. So the change is actually required by the formatter — not unrelated.

RichardTang-Aden:
Hey @Siddharth2624 — ran a live test against our HubSpot instance. Requesting changes, the core pipeline is non-functional.                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                      
  The agent finds deals but then loses them. Monitor correctly pulled 7 ghosted deals totaling $71k, but every subsequent node got nothing:                                                                                                                                                           
                                                                                                                                                                                                                                                                                                      
  - Monitor — found 7 deals, $71k. Works.
  - Analyze — detect_revenue_leaks returns "No deal data found." 0 leaks.                                                                                                                                                                                                                             
  - Notify — reports "Pipeline Healthy." No alert sent.                                                                                                                                                                                                                                               
  - Followup — "No GHOSTED contacts." 0 drafts created.
                                                                                                                                                                                                                                                                                                      
  This repeated for all 3 cycles. The agent finished and reported "$0 at risk, pipeline is clean" while $71k in genuinely stalled deals went undetected.                                                                                                                                              
                                                                                                                                                                                                                                                                                                      
  Why: You're using contextvars (_deals_cache_var, _leaks_var) to pass data between nodes, but the framework runs each event_loop node in its own async context. By the time analyze runs, the contextvar is empty. This is a fundamental mismatch — the LLM even flagged it in its own reflection ("a
   disconnect between the storage in scan_pipeline and the detection tool's access to that data") and then kept going anyway.
                                                                                                                                                                                                                                                                                                      
  Other problems:            

  - Contact resolution is broken — hubspot_search_contacts gets called with deal names like "iNX - ERP Franchise Centralization" as the query. Obviously returns nothing. You need HubSpot's Associations API to look up contacts by deal ID.                                                         
  - Raw stage IDs in output — deals show up as stage 969753701 instead of readable names. _STAGE_MAP only covers default HubSpot stages.
  - Custom Telegram HTTP call is a problem — _auto_fetch_telegram_chat_id makes a raw httpx.get to the Telegram API outside our MCP tool system, bypasses credential management, and caches the result in a module-level global. We already have telegram_send_message as an MCP tool. Chat ID        
  discovery needs to go through the same MCP server, not be a side-channel baked into tool logic.                                                                                                                                                                                                     
  - Agent doesn't question contradictions — the LLM saw 7 inactive deals in Monitor, then 0 leaks in Analyze, noted the discrepancy, and still reported the pipeline as healthy. System prompts need to handle this case.                                                                             
                                                                                                                                                                                                                                                                                                      
  Before re-review:          
                                                                                                                                                                                                                                                                                                      
  1. Fix state transfer — don't use contextvars across nodes. Pass deal data through set_output/input_keys like cycle and halt already do, or use session state.                                                                                                                                      
  2. Fix contact lookup — use deal ID associations, not name search.
  3. Handle custom pipeline stages (don't show raw numeric IDs).                                                                                                                                                                                                                                      
  4. Remove the raw Telegram HTTP call — put chat ID resolution in the MCP server.
  5. Re-test end-to-end. Confirm the agent actually detects leaks, sends an alert, and drafts emails in one run.

PR #5239, opened 2026-02-22
 feat(executor): add credential-aware tool resolution to prevent hallucination on auth failures 

### LINKED ISSUE #5053: [Feature]: Capability based tool resolution -- auto select tools based on available credentials.
## Problem Statement
Agents hardcode specific tool names in their node definitions (e.g., `web_search`), so they crash at runtime if the user has a different provider's API key configured. The repo already has multiple tools that serve the same capability — for example, `web_search` (Brave/Google), `exa_search` (Exa), and `scholar_search` (SerpAPI) all do web search. But there's no way for agents to fall back to an available alternative when one provider's credentials are missing. The framework should auto-detect which tools have valid credentials and resolve to the best available option.

For example, I built a Scene Scout agent with a research node that declares `tools=["web_search"]`. The `web_search` tool requires `BRAVE_SEARCH_API_KEY` or `GOOGLE_API_KEY`, but I only have  `EXA_API_KEY`. The agent crashes at the research step, even though `exa_search` is already a registered tool in the framework doing the exact same job. Yes, I can manually swap to `exa_search`, but that means every agent builder has to know which specific tool matches their API keys. This won't scale as more providers are added — agent builders shouldn't need to care  which provider the user has.

  ## Proposed Solution

  Introduce a **capability-based tool resolution layer** in the framework. Agents declare the *capability* they need rather than a specific tool name:

  ```python
  # Instead of this (current):
  tools=["web_search"]

  # Something like this:
  tools=[Capability("search")]  # resolves to whichever search tool has valid credentials

  The ToolRegistry would check available credentials at startup and resolve capabilities to the best available tool automatically:

  Agent declares capability → ToolRegistry checks available credentials at startup → Resolves to the best available tool → Injects resolved tool name into the node


  Priority order would be configurable with sensible defaults:

  CAPABILITY_DEFAULTS = {
      "search": ["exa_search", "web_search", "scholar_search"],
      "email": ["send_email"],  # provider auto-detected from available keys
  }
```

  **Alternatives Considered:**

  1. Tool fallback chain — nodes try tools in order until one works. Simple but wasteful — burns LLM tokens on failed tool calls before finding one that works.
  2. Config-driven mapping — users define tool preferences in agent config. Works but doesn't scale — every user has to manually configure mappings, and every new tool integration requires    
  updating docs.
  3. LLM-driven selection — give the node all tools of a capability and let the LLM pick. Unpredictable, wastes API credits, and the LLM has no way to know which credentials are valid.        

  **Additional Context:**
  Image
  - The aden framework already have 184 tools across many providers. As more integrations get added, this problem will only get worse.
  - Overlapping capabilities already exist: web search (3 providers), email (2 providers), and more will come.
  - This affects every agent builder, not just my use case — anyone who doesn't have the exact API key an agent was built with will hit this.

  **Implementation Ideas:**

  Credential-aware resolution at the framework level is the best fit because:

  1. Zero burden on agent builders — they just declare Capability("search"), the framework handles the rest.
  2. Fits existing architecture — ToolRegistry already discovers tools at startup and CredentialStoreAdapter already knows which keys are available. Connecting these two is a natural
  extension.
  3. Deterministic — tool selection happens before the LLM runs, no wasted tokens.
  4. Scales with the ecosystem — automatic resolution beats manual config as the tool count grows.

**Suggested flow:**
  - Define a CAPABILITY_MAP that groups tools by capability
  - On agent startup, ToolRegistry checks which tools in each group have valid credentials
  - Resolved tool names are injected into node specs before execution
  - Power users can override defaults in agent config

Trinath052004:
## Description
 When a tool's API key is missing or invalid, the LLM was silently falling back to training data instead of stopping. This PR adds Tier 1 and Tier 2 tool resolution to catch credential issues early and halt the agent immediately.

**Tier 1** (`tools=["web_search"]`) — validates both tool registration and credential env var at startup. Stops the agent before it runs if credentials are missing.

**Tier 2** (`tools=[["web_search", "exa_search"]]`) — fallback group. Framework picks the first tool whose credential is set, logs skipped tools, and errors if none are available.

At runtime, any tool returning a 401/403/auth error is tagged with `credential_error: true` and the agent halts immediately instead of continuing with training data.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #5053 

## Changes Made

 When a tool's API key is missing or invalid, the LLM was silently falling back to training data instead of stopping. This PR adds Tier 1 and Tier 2 tool resolution to catch credential issues early and halt the agent immediately.

**Tier 1** (`tools=["web_search"]`) — validates both tool registration and credential env var at startup. Stops the agent before it runs if credentials are missing.

**Tier 2** (`tools=[["web_search", "exa_search"]]`) — fallback group. Framework picks the first tool whose credential is set, logs skipped tools, and errors if none are available.

At runtime, any tool returning a 401/403/auth error is tagged with `credential_error: true` and the agent halts immediately instead of continuing with training data.

## Testing:

- [x] Manual testing performed (scene_scout agent with TUI — credential errors surface correctly)

 New test file: `core/tests/test_tool_resolution.py` — 13 tests covering Tier 1 registration, Tier 2 fallback, `_is_auth_error`, `credential_error` flag, and  `all_tool_names`.

## Checklist

## Screenshots:
#Before:
Screenshot 2026-02-18 024942

#After:
Screenshot 2026-02-23 030735

@Hundao ,Thanks for the Trust, looking forward to  make further changes if required:)

### DIFF
--- a/core/framework/credentials/validation.py
+++ b/core/framework/credentials/validation.py
@@ -104,9 +104,11 @@ def validate_agent_credentials(nodes: list, quiet: bool = False, verify: bool = 
         quiet: If True, suppress the credential summary output.
         verify: If True (default), run health checks on present credentials.
     """
-    # Collect required tools and node types
-    required_tools = {tool for node in nodes if node.tools for tool in node.tools}
-    node_types = {node.node_type for node in nodes}
+    required_tools: set[str] = set()
+    for node in nodes:
+        if node.tools:
+            required_tools.update(node.all_tool_names)
+    node_types: set[str] = {node.node_type for node in nodes}
 
     try:
         from aden_tools.credentials import CREDENTIAL_SPECS
--- a/core/framework/graph/event_loop_node.py
+++ b/core/framework/graph/event_loop_node.py
@@ -1369,6 +1369,22 @@ class EventLoopNode(NodeProtocol):
                         result = raw
                     results_by_id[tc.tool_use_id] = self._truncate_tool_result(result, tc.tool_name)
 
+            # Credential error check: if any tool signalled credential_error,
+            # stop immediately — do not let the LLM continue with training data.
+            for tc in pending_real:
+                r = results_by_id.get(tc.tool_use_id)
+                if r and r.is_error:
+                    try:
+                        parsed = json.loads(r.content)
+                        if parsed.get("credential_error"):
+                            raise RuntimeError(
+                                f"Tool '{tc.tool_name}' failed: credential is invalid or "
+                                f"unauthorized. Check your API key and re-run the agent.\n"
+                                f"Details: {parsed.get('error', r.content)}"
+                            )
+                    except (json.JSONDecodeError, AttributeError):
+                        pass
+
             # Phase 3: record results into conversation in original order,
             # build logged/real lists, and publish completed events.
             for tc in tool_calls[:executed_in_batch]:
@@ -2283,7 +2299,7 @@ class EventLoopNode(NodeProtocol):
 
         # 4. Available tools reminder
         if spec.tools:
-            parts.append(f"AVAILABLE TOOLS: {', '.join(spec.tools)}")
+            parts.append(f"AVAILABLE TOOLS: {', '.join(spec.all_tool_names)}")
 
         # 5. Spillover files — list actual files so the LLM can load
         # them immediately instead of having to call list_data_files first.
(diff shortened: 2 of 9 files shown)


Trinath052004:
Hi @Hundao 👋,
just a friendly followup,
This PR is ready for review. Please let me know if any changes are needed.
Thanks!

PR #5410, opened 2026-02-26
fix(credentials): handle empty and whitespace env vars in EnvVarStorage

### LINKED ISSUE #5388: [Bug]: EnvVarStorage.exists() can disagree with load() for empty env/.env values
## Problem
`EnvVarStorage.exists()` and `EnvVarStorage.load()` use different truthiness semantics for credential values.

- `load()` returns `None` when value is falsy (`if not value`)
- `exists()` returns `True` when value is not `None`

This can produce inconsistent behavior for empty values (e.g., empty string from `.env`):
- `exists(id)` -> `True`
- `load(id)` -> `None`

## Expected Behavior
`exists()` and `load()` should agree on whether a credential is available.

## Proposed Fix
Align `exists()` behavior with `load()` semantics (or vice versa), then add tests for:
1. empty string value
2. whitespace-only value (if considered invalid)
3. normal non-empty value

## Affected files
- `core/framework/credentials/storage.py`
- `core/framework/credentials/tests/test_credential_store.py`

orphicdusk:
Resolves #5388

### **Problem**
The current implementation of `EnvVarStorage` fails to correctly process environment variables that are declared but left empty (e.g., `export API_KEY=`) or contain only whitespace. 
* In `_read_env_value`, the logic `if value:` evaluates to `False` for empty strings, causing the system to ignore the environment override and incorrectly fall back to the `.env` file.
* There is a state mismatch between `exists()` and `load()`. A key might technically "exist" in `os.environ`, but fail to load a valid credential if it's just an empty string or whitespace.

### **Solution**
Refactored the `EnvVarStorage` class to correctly recognize empty environment variables and strictly validate their contents before loading. 

* **Fixed Environment Override (`_read_env_value`):** Changed the check to `if env_var in os.environ:` to ensure explicitly declared variables (even if empty) are correctly read rather than bypassed.
* **Synchronized Validation:** * Updated `exists()` to verify that the environment variable contains non-whitespace characters: `return bool(value and value.strip())`.
  * Updated `load()` to mirror this validation, safely returning `None` if the string is empty or whitespace-only.
* **Added Test Coverage:** Added `test_empty_env_var` and `test_whitespace_env_var` in `test_credential_store.py` to prevent future regressions.

### **Testing**
- [x] Added unit tests for empty string variables (`""`).
- [x] Added unit tests for whitespace variables (`"   "`).
- [x] All existing tests pass successfully.

### DIFF
--- a/core/framework/credentials/storage.py
+++ b/core/framework/credentials/storage.py
@@ -331,9 +331,8 @@ class EnvVarStorage(CredentialStorage):
     def _read_env_value(self, env_var: str) -> str | None:
         """Read value from env var or .env file."""
         # Check os.environ first (takes precedence)
-        value = os.environ.get(env_var)
-        if value:
-            return value
+        if env_var in os.environ:
+            return os.environ[env_var]
 
         # Fallback: read from .env file (hot-reload)
         if self._dotenv_path.exists():
@@ -360,7 +359,7 @@ class EnvVarStorage(CredentialStorage):
         env_var = self._get_env_var_name(credential_id)
         value = self._read_env_value(env_var)
 
-        if not value:
+        if not value or not value.strip():
             return None
 
         return CredentialObject(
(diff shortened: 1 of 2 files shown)


orphicdusk:
"Hi! I have put together a fix for this and opened a PR. Could you please assign this issue to me so my PR can be processed?"

PR #5508, opened 2026-02-28
docs: clarify function node registration in execution paths

kostasuser01gr:
## Summary
- Addresses [#4725](https://github.com/aden-hive/hive/issues/4725)
- Clarifies that exported agents containing **function** nodes need to use the `GraphExecutor` path in `agent.py` to register them before execution.

## Root Cause
- The `create_agent_runtime()` path does not expose a way to register function implementations, leaving users confused about how to hook up custom Python functions when running agents via `hive run` or `hive tui`.

## Solution
- Added a new subsection titled **Agents with Function Nodes** under the `## Running an Agent` section in both `README.md` and `docs/getting-started.md`.
- Provided the recommended 3-step pattern (`GraphExecutor(...)` -> `register_function()` -> `execute()`).
- Added a reference to the `examples/templates/` directory for reference implementations.

## Validation
- [x] Local Markdown format checked.
- [x] Verified correct linking and rendering locally.

Fixes #4725


kostasuser01gr:
Ready for review. All checks passing. 🚀

PR #5536, opened 2026-03-01
fix(core): micro-fix prevent sse critical event queue from blocking event bus (#5533)

### LINKED ISSUE #5532: [Bug]: safe_path_segment allows '.' segment, creating ambiguous/deletable path targets in session routes
## Describe the Bug
safe_path_segment() blocks /, \, and .., but allows ".". Some routes combine this with filesystem operations (including deletion). Allowing "." can resolve to parent/current directories and creates ambiguous target resolution in sensitive paths.

## To Reproduce
1). Use a route that consumes {ws_id} and performs filesystem actions (e.g. delete worker session/checkpoint paths).
2).  ws_id as ".".
3). Observe route accepts the segment and resolves to parent/current directory semantics.

## Expected Behavior
safe_path_segment() should reject "." (and ideally enforce a strict allowlist like ^[A-Za-z0-9._-]+$ plus deny . and .. explicitly).

## Configuration
api_routes:
  uses_safe_path_segment: true

## Logs
No validation error for ws_id="."
Filesystem path resolves using '.' semantics.

## Additional Context
Relevant code paths:

- core/framework/server/app.py (safe_path_segment)
- core/framework/server/routes_sessions.py (handle_delete_worker_session uses shutil.rmtree)

### LINKED ISSUE #5533: [Bug]: SSE critical-event queue can block event pipeline under slow client backpressure
## Describe the Bug
In SSE route handling, each client has a bounded queue (maxsize=1000). For critical events, enqueue uses blocking await queue.put(...) when full. A slow/disconnected client can cause backpressure that blocks the handler path for critical events.

## To Reproduce
1). Open SSE connection to /api/sessions/{id}/events.
2). Simulate very slow/blocked consumer so queue fills.
3). Generate high event volume with critical event types.
4). Observe enqueue path blocks on await queue.put(...).

## Expected Behavior
Event pipeline should remain non-blocking for per-client backpressure (e.g., bounded drop policy, disconnect slow clients, or separate non-blocking critical buffering strategy).

## Configuration
sse:
  queue_maxsize: 1000
  critical_events_blocking_put: true

## Logs
SSE connected...
No immediate error, but server responsiveness degrades under slow client + high event throughput.

## Additional Context
Relevant code path:

-core/framework/server/routes_events.py (queue = asyncio.Queue(maxsize=1000), critical events use await queue.put)

Rajneesh180:
Fixes #5533, Fixes #5532

This PR prevents the publisher from blocking on the event loop by safely forcing a disconnect on slow clients when the critical queue fills up. Also patches a small path traversal vulnerability in  safe_path_segment.

### DIFF
--- a/core/framework/server/app.py
+++ b/core/framework/server/app.py
@@ -18,7 +18,7 @@ def safe_path_segment(value: str) -> str:
     traversal sequences.  aiohttp decodes ``%2F`` inside route params,
     so a raw ``{session_id}`` can contain ``/`` or ``..`` after decoding.
     """
-    if "/" in value or "\\" in value or ".." in value:
+    if not value or value == "." or "/" in value or "\\" in value or ".." in value:
         raise web.HTTPBadRequest(reason="Invalid path parameter")
     return value
 
--- a/core/framework/server/routes_events.py
+++ b/core/framework/server/routes_events.py
@@ -92,11 +92,23 @@ async def handle_events(request: web.Request) -> web.StreamResponse:
         "worker_loaded",
     }
 
+    client_disconnected = asyncio.Event()
+
     async def on_event(event) -> None:
         """Push event dict into queue; drop non-critical events if full."""
+        if client_disconnected.is_set():
+            return
+
         evt_dict = event.to_dict()
         if evt_dict.get("type") in _CRITICAL_EVENTS:
-            await queue.put(evt_dict)  # block rather than drop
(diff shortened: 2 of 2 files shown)


Hundao:
@Rajneesh180 Thanks for the fix. For future PRs, please split unrelated changes into separate PRs so they can be reviewed independently.

Issue #1728, opened 2026-01-28
[Bug]: Parallel branch memory writes ignore memory_conflict_strategy


SailakshmiSingh9:
## Describe the Bug

ParallelExecutionConfig defines a memory_conflict_strategy ("last_wins", "first_wins", "error"), but parallel branches currently write directly into shared memory without using this strategy. When multiple branches write the same key, the final value depends on scheduling order instead of the configured conflict behavior.

## To Reproduce

1. Define a graph with a node that fans out into multiple parallel branches (more than one outgoing traversable edge).
2. Have at least two branches write to the same memory key (e.g., both branches output {"some_shared_key": ...}).
3. Enable parallel execution with a custom ParallelExecutionConfig, e.g. set memory_conflict_strategy to "first_wins" or "error".
4. Run the graph multiple times and inspect the final value of the shared key in memory.
5. Observe that:
5.1: The final value can vary depending on which branch finishes last.
5.2: The configured memory_conflict_strategy does not appear to have any effect.

## Expected Behavior

When multiple parallel branches write to the same memory key, the final result should follow the configured memory_conflict_strategy:
"last_wins": the last write deterministically wins.
"first_wins": the first write wins; subsequent writes to that key are ignored.
"error": conflicting writes to the same key cause a clear error instead of silently overwriting.

## Screenshots

No screenshots; this is a logic/behavior issue in parallel execution and memory writes.

## Environment

OS: (e.g.) Ubuntu 22.04 / macOS 14 / Windows 11
Python version: 3.11.x
Docker version : N/A (not required to demonstrate the behavior)

## Configuration

Relevant core configuration is in code:

core/framework/graph/executor.py

1) ParallelExecutionConfig:
 class ParallelExecutionConfig:
     ...
     memory_conflict_strategy: str = "last_wins"  # "last_wins", "first_wins", "error"

 2) _execute_parallel_branches():
 - Writes to shared memory are done via memory.write_async(...)
   without checking or enforcing the memory_conflict_strategy
   when multiple branches write the same key.

## Logs

No explicit errors are logged. When multiple parallel branches write to the same key, the final value in memory depends on branch completion order, and the configured memory_conflict_strategy is not mentioned or enforced in logs.

## Additional Context

In _execute_parallel_branches (core/framework/graph/executor.py), each branch writes mapped inputs and outputs using memory.write_async(), but there is no conflict detection or use of memory_conflict_strategy.

[...]

PR #5604, opened 2026-03-01
Fix/5423 parallel safety audit

### LINKED ISSUE #5423: [Bug]: [RFC] Parallel Execution: Missing Runtime Enforcement of Disjoint output_keys Leads to Silent State Overwrite
###  **The Bug**

While reviewing the GraphExecutor parallel branch execution flow, I noticed a potential architectural risk around SharedMemory writes during fan-out execution.
Currently, when enable_parallel_execution=True, multiple branches may execute concurrently and write to SharedMemory using overlapping output_keys. If two parallel nodes write to the same key, the final value depends on whichever branch completes last.
There does not appear to be a runtime enforcement or safety check for disjoint output_keys during parallel execution, even when memory_conflict_strategy is configured.

This results in silent state overwrite with no warning or error surfaced to the user creating a real risk of silent data corruption in downstream nodes.

## To Reproduce

1. Define a GraphSpec with a fan-out from Node A to Node B and Node C.
2. Set enable_parallel_execution = True.
3. Configure both Node B and Node C with overlapping:
output_keys = ["analysis_result"]
4. Execute the graph.

Observed result:
Both nodes complete successfully.
SharedMemory ends with only the value from the branch that finished last.
No warning or error is logged.
The earlier value is silently discarded.

## Expected Behavior

At graph initialization (or before parallel execution begins), the executor should perform a Parallel Safety Audit:

If memory_conflict_strategy == "error"
Execution should halt early with a clear message indicating conflicting output_keys.
This prevents wasted LLM tokens and surfaces a design flaw immediately.

If memory_conflict_strategy == "last_wins" or "first_wins"
Execution may proceed, but a logger.warning should explicitly indicate:
-The conflicting keys
-The nodes involved
-The chosen resolution strategy
making the behavior intentional rather than accidental.

## Environment

-OS: Linux/WSL2 (Ubuntu 22.04)
-Python version: 3.11+
-Version: Current main branch (Feb 2026)

## Configuration

parallel_config:
  enabled: true
  memory_conflict_strategy: "error"
nodes:
  - id: "research_node_1"
    output_keys: ["summary"]
  - id: "research_node_2"
    output_keys: ["summary"]  # overlapping key

## Logs

[INFO] Starting execution: Market Research
[INFO] ▶ Step 1: node_a (router)
[INFO] ⑃ Fan-out: executing research_node_1 and research_node_2 in parallel
[INFO] ✓ research_node_1 success
[INFO] ✓ research_node_2 success
[INFO] ✓ Execution complete

## Additional Context

While investigating the behavior around parallel execution, I revisited the existing validation flow in GraphExecutor. At present, schema validation operates as a binary pass/fail check — when a node violates its output contract, execution transitions into standard failure/retry handling.
This works correctly from an execution-control perspective. However, it may be worth considering whether validation failures could optionally surface structured feedback into the retry context. Instead of treating contract violations purely as execution errors, they could serve as corrective signals to improve retry attempts.
I’m not proposing a change in failure semantics here — only highlighting a potential enhancement path where validation metadata (e.g., missing keys or schema mismatch details) could be exposed more explicitly to the retry loop.
This could strengthen deterministic reliability, particularly in parallel or high-concurrency workflows, without altering existing execution guarantees.

Happy to explore this further if aligned with the project direction.

**Suggested Direction**

I suggest adding a pre-execution audit step when parallel execution is enabled:
-Collect output_keys for all nodes in the fan-out group.
-Detect intersections.
-Apply memory_conflict_strategy deterministically.
This keeps the behavior explicit and aligns parallel execution semantics with the declared configuration.

Nikku29:
Description
Implements a Parallel Safety Audit in the GraphExecutor to prevent silent state overwrites during parallel fan-out execution.

Type of Change
[x] Bug fix (non-breaking change that fixes an issue)

[x] New feature (non-breaking change that adds functionality)

Related Issues
Fixes #5423
Addresses architectural debt in #1728

Changes Made
Audit Logic: Integrated a pre-dispatch check in _execute_parallel_branches that maps output_keys to node owners before task assignment.

Conflict Enforcement: Enforced memory_conflict_strategy. Raises a RuntimeError if set to "error" to prevent wasted LLM tokens; triggers logger.warning for "last_wins" or "first_wins".

Metadata Injection: Attached a structured metadata dictionary to the RuntimeError containing conflicting keys and node IDs to enable future self-healing/retry logic.

Testing
[x] Unit tests pass: 13/13 passed in core/tests/test_fanout.py.

[x] Lint passes via ruff check ..

[x] Manual testing performed: Verified both "error" and "warning" paths.

Checklist
[x] Code follows project style guidelines

[x] Performed self-review of code

[x] Added tests proving fix is effective

[x] New and existing unit tests pass locally

Screenshots
Screenshot 2026-03-01 200854
Screenshot 2026-03-01 201021
Screenshot 2026-03-01 201202

### DIFF
--- a/core/framework/graph/edge.py
+++ b/core/framework/graph/edge.py
@@ -712,26 +712,4 @@ class GraphSpec(BaseModel):
                     f"{client_facing_targets}. Only one branch may be client-facing."
                 )
 
-        # Output key overlap on parallel event_loop nodes
-        for source_id, targets in fan_outs.items():
-            event_loop_targets = [
-                t
-                for t in targets
-                if self.get_node(t) and getattr(self.get_node(t), "node_type", "") == "event_loop"
-            ]
-            if len(event_loop_targets) > 1:
-                seen_keys: dict[str, str] = {}
-                for node_id in event_loop_targets:
-                    node = self.get_node(node_id)
-                    for key in getattr(node, "output_keys", []):
-                        if key in seen_keys:
-                            errors.append(
-                                f"Fan-out from '{source_id}': event_loop nodes "
-                                f"'{seen_keys[key]}' and '{node_id}' both write to "
-                                f"output_key '{key}'. Parallel event_loop nodes must "
-                                f"have disjoint output_keys to prevent last-wins data loss."
-                            )
-                        else:
-                            seen_keys[key] = node_id
-
         return errors
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -1953,7 +1953,9 @@ class GraphExecutor:
             strategy = self._parallel_config.memory_conflict_strategy
             if strategy == "error":
                 self.logger.error(f"   ✗ {conflict_msg} (strategy='error')")
-                raise RuntimeError(conflict_msg)
+                err = RuntimeError(conflict_msg)
+                err.metadata = conflicting_keys
+                raise err
             else:
                 self.logger.warning(
                     f"   ⚠ {conflict_msg} Resolution strategy: '{strategy}'"
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -1929,6 +1929,36 @@ class GraphExecutor:
                 edge=edge,
             )
 
+        # Parallel Safety Audit: Check for disjoint output keys
+        output_key_owners: dict[str, list[str]] = {}
+        for branch in branches.values():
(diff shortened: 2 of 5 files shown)


Nikku29:
Linking this to the master issue #1728. I am ready to re-open this once the assignment is cleared or if the maintainers would like me to merge my logic into the current workflow

PR #5773, opened 2026-03-04
micro-fix: set compile-report node to client-facing (refs #4907)

### LINKED ISSUE #4907: bug: compile-report node has client_facing set to false despite delivering output directly to the user
## Description

In the `tech_news_reporter` example agent, the `compile-report` node's `client_facing` field is set to `false` in `agent.json`, but the node's system prompt explicitly instructs the agent to present a text summary to the user and then serve a file link for them to open.

This seems like a misconfiguration — if a node is producing user-visible output (summary text + a clickable report link), it should probably be marked `client_facing: true`.

## Where to look

- **File:** `examples/templates/tech_news_reporter/agent.json`
- **Node:** `compile-report` (third node in the graph)

The `client_facing` field is `false`:

```json
"client_facing": false
```

But the system prompt says:

> "STEP 1 — Respond to the user (text only, NO tool calls): Present a brief text summary of the report highlights..."

This contradicts the `client_facing: false` setting.

## Expected behavior

`compile-report` should have `client_facing: true` since it directly communicates with the user and delivers the final report.

## Impact

This could confuse new contributors trying to understand the node architecture. Example agents serve as reference implementations, so accuracy matters here. It may also affect how the TUI or streaming interface routes messages — if `client_facing` is used to determine whether to display node output to the user, this node's output might be silently hidden.

## Steps to reproduce

1. Open `examples/templates/tech_news_reporter/agent.json`
2. Check the `compile-report` node's `client_facing` value
3. Compare it with the node's `system_prompt` content

## Suggested fix

Set `client_facing` to `true` for the `compile-report` node in both `agent.json` and the corresponding node definition in `nodes/__init__.py`.

## Environment

- Hive version: latest main branch
- OS: macOS

kostasuser01gr:
Summary:
- Updated 'compile-report' node in tech_news_reporter template to be client-facing.
- This allows the final report to be delivered interactively to the user in the UI.

Root cause:
- The node was incorrectly set to client_facing=False, which prevented it from being visible or interactive in the UI when execution completed.

Solution:
- Flipped boolean to client_facing=True.

Validation:
- Verified locally with a dedicated test case ensuring the attribute is set correctly.
- Commands: uv run --project core python core/tests/repro_4907.py
- Results: [PASSED] compile_report_node.client_facing is now True.

Risk / edge cases:
- None. This is a configuration update for a template node.

Fixes #4907


RichardTang-Aden:
PR description filed as an issue — please submit as a pull request instead.

Issue #5699, opened 2026-03-03
[Bug] Opening a second tab for the same agent overwrites the first tab's session instead of creating an isolated parallel session


levxn:
## Summary

When a second tab is opened for an agent that already has an active tab (via **+** → Existing agent → same agent, or **+** → New agent → From scratch when a New Agent tab is already open), the new tab gets the correct "second" label but does not create an isolated session. Instead it mutates the first tab: the first tab's conversation is overwritten, the pipeline/graph panel clears or reloads incorrectly, and both tabs effectively share the same UI state. The two tabs do not behave as independent parallel workspaces.

This affects all agent types:
- **Existing agents** (e.g. Deep Research Agent → Deep Research Agent (second instance))
- **New Agent** (New Agent → New Agent (second instance))

---

### Existing agent variant
1. Open the UI and click an existing agent (e.g. "Deep Research Agent") from the **+** menu.
2. Wait for the session to load — pipeline graph appears, queen is ready.
3. Send one or two messages so there is visible conversation history.
4. Click **+** again → **Existing agent** → select the same "Deep Research Agent".
5. Observe:
   - A new tab appears labelled **Deep Research Agent (second instance)** ✓
   - The pipeline graph in tab 1 disappears or reloads ✗
   - The conversation history in tab 1 is wiped or replaced with tab 2's state ✗
   - Both tabs appear to share the same message thread ✗

### New Agent variant
1. Open the UI — default **New Agent** tab is present.
2. Send a few messages to Queen Bee.
3. Click **+** → **New agent** → **From scratch**.
4. Observe:
   - Tab is labelled **New Agent (second instance)** ✓
   - Tab 1's conversation is overwritten / blank ✗
   - Switching back to tab 1 shows tab 2's empty state instead of the prior conversation ✗

---

## Root Cause

`agentStates` and `sseSessions` in `workspace.tsx` are both keyed by **`agentType`** (a string like `"new-agent"` or `"exports/deep_research"`), not by the individual frontend session ID.

```ts
// agentStates shape
const [agentStates, setAgentStates] = useState<Record<string, AgentBackendState>>({});
//                                                       ^ agentType string
//                                                         shared across ALL tabs of the same agent

// sseSessions built from agentStates
const sseSessions = useMemo(() => {
  const map: Record<string, string> = {};
  for (const [agentType, state] of Object.entries(agentStates)) {
    if (state.sessionId && state.ready) {
[...]

PR #5815, opened 2026-03-04
Restoring session during server restart | smooth conversation picked from where left off | fix unhandled error in event routes | 

levxn:
## Summary

Queen conversation files written to `~/.hive/queen/session/{session_id}/conversations/` survive a server restart, but the server had no way to serve them afterwards. `GET /api/sessions/{id}` returned 404 (in-memory dict wiped), `GET /api/sessions/{id}/queen-messages` required a live session object, and the frontend silently created a brand-new session — orphaning the on-disk history forever. Even when the frontend did reconnect, the queen started a fresh conversation instead of continuing where the user left off. Separately, worker agents requesting free-text input (ask_user without options) never received the user's reply — the message was always routed to the queen instead.

This PR makes cold (post-restart) sessions first-class, ensures the queen continues the conversation naturally on reconnect, and fixes worker free-text input routing.

---

### `core/framework/server/session_manager.py`

**Cold session discovery** — added two static helper methods that inspect `~/.hive/queen/session/` on disk without requiring a live `Session` object in memory:

- **`get_cold_session_info(session_id)`** — checks whether conversation files exist for a given session ID and returns `{session_id, cold: True, has_messages, created_at}`, or `None` if no data is found.
- **`list_cold_sessions()`** — scans all subdirectories under `~/.hive/queen/session/` and returns them newest-first. Used by the history endpoint.

No SQLite dependency added — the conversation directory itself serves as the index.

**Cold restore continuity** — the `Session` dataclass now carries a `queen_resume_from: str | None` field. When set, `_start_queen` redirects all queen conversation I/O to the **original** session's directory (`~/.hive/queen/session/{original_id}/`) so messages accumulate in one place across restarts. `_queen_loop` passes `cold_restore: True` in the session state so the executor knows not to treat this as a fresh execution.

---

### `core/framework/graph/executor.py`

The `_is_fresh_shared` guard (which resets the conversation cursor and injects a "start fresh" transition marker) now excludes cold restores via `and not session_state.get("cold_restore")`. Without this, every reconnect wiped the cursor and told the LLM to treat the conversation as a brand-new request — losing all prior context.

---

### `core/framework/server/routes_sessions.py`

**`handle_get_live_session`** — instead of returning 404 immediately when a session is not in memory, it now calls `SessionManager.get_cold_session_info(session_id)` first. If disk data exists it returns HTTP 200 with `{cold: true, has_messages, created_at}`. A genuinely unknown session still returns 404.

**`handle_queen_messages`** — removed the `resolve_session(request)` live-session guard. The endpoint now derives the session directory path directly from the `session_id` URL parameter and reads from disk. Works for both live and cold sessions with no behaviour change for the live case.

**`handle_session_history`** (new) — `GET /api/sessions/history` scans all on-disk queen session directories and returns them as a list with `live`/`cold` flags. Live sessions (present in the in-memory dict) are marked `live: true, cold: false`; everything else is `live: false, cold: true`. Route registered before `{session_id}` so it takes priority.

---

### `core/frontend/src/api/types.ts`

Added `cold?: boolean` to `LiveSessionDetail` so TypeScript knows about the new field without a cast.

---

### `core/frontend/src/api/sessions.ts`

- Updated `queenMessages` return type to include `session_id` in the response shape.
- Added `history()` method for `GET /api/sessions/history`.

---

### `core/frontend/src/pages/workspace.tsx`

**Cold restore reconnect** — rewrote the reconnect logic in `loadAgentForType` for the `"new-agent"` path:

1. `sessionsApi.get(storedId)` returns `{cold: true}` → server restarted, files are on disk.
2. New live session is created with `queen_resume_from` set to the old session ID — queen writes to the original directory and continues the conversation.
3. `sessionsApi.queenMessages(oldSessionId)` fetches prior messages from disk and injects them into the chat panel sorted chronologically.
4. If no messages exist on disk (truly fresh scenario), the cache is cleared as before.
5. The initial prompt is only shown as a user message when zero messages were restored (avoids duplication).

**Worker free-text input routing** — when a worker fires `client_input_requested` without options (free-text `ask_user`), `pendingQuestionSource` was set to `null` instead of `"worker"`. This caused `handleSend` to always route the user's reply to the queen via `executionApi.chat()`, leaving the worker blocked forever. Fixed by:
- Always setting `pendingQuestionSource: "worker"` when the worker requests input (with or without options).
- Adding an early-return in `handleSend` that detects `awaitingInput && pendingQuestionSource === "worker" && !pendingOptions` and routes directly to `executionApi.workerInput()`.

**Tab label preservation** — session labels like `Agent #2` no longer reset to just `Agent` on reload. The existing label is kept when it already starts with the agent display name.

---

### `core/frontend/src/pages/home.tsx`

Removed the HistorySidebar component (was too buggy to ship; sessions list endpoint still exists for future use).

---

## Before / After

| Scenario | Before | After |
|---|---|---|
| Server restarts, user refreshes | Blank chat, history lost | Prior conversation restored, queen continues naturally |
| `GET /api/sessions/{id}` after restart | 404 | 200 with `cold: true` |
| `GET /api/sessions/{id}/queen-messages` after restart | 404 (live session required) | Returns messages from disk |
| `GET /api/sessions/history` | 404 (no endpoint) | Lists all sessions on disk |
| Queen conversation after reconnect | Fresh conversation, previous context lost | Continues from where it left off |
| Worker asks free-text question | User reply sent to queen, worker stuck | Reply routed to worker correctly |
| Tab label `Agent #2` after reload | Reset to `Agent` | Preserved as `Agent #2` |
| Normal refresh (server still running) | History restored | No change | 

closes issues #5692 and #5699 

---

## Recording

https://github.com/user-attachments/assets/13eadc5c-2c21-4cfa-aa2a-9a96336b3f49


levxn:
@RichardTang-Aden @TimothyZhang7 @bryanadenhq review it and let me know :))

PR #5874, opened 2026-03-05
fix(tools): correct multibyte-safe truncation and encoding in view_file

### LINKED ISSUE #5163: [Bug]: view_file truncates by character count instead of byte size and reports incorrect size_bytes
## Description

The `view_file` tool enforces a `max_size` limit (in bytes). However, truncation is currently performed using character slicing:

```python
if len(content.encode(encoding)) > max_size:
    content = content[:max_size]
```

Since Python strings represent Unicode code points, slicing by `max_size` treats the byte limit as a character limit. For multibyte encodings (e.g., UTF-8), this can:

- Exceed the intended byte limit
- Potentially split encoded sequences
- Cause inaccurate downstream memory/token budgeting

Additionally, the returned `size_bytes` metadata is computed using a hardcoded `"utf-8"` encoding instead of the requested `encoding` parameter.

---

### Case 1: Byte limit exceeded

```python
content = "🌍🌎🌏🌐🌑"
max_size = 6

truncated = content[:6]
assert len(truncated.encode("utf-8")) <= 6  # Fails
```

### Case 2: Incorrect metadata

```python
content = "Hello"  # 12 bytes in UTF-16 (with BOM)

# Current implementation reports:
# {"size_bytes": len(content.encode("utf-8"))}
# which does not reflect the requested encoding
```

---

## Expected Behavior

- Truncation should occur at the byte boundary, not character count.
- `size_bytes` should be calculated using the same encoding passed to the function.

---

## Suggested Fix

1. Encode content into bytes.
2. Slice at the `max_size` byte boundary.
3. Decode safely using `errors="ignore"` to drop partial multibyte sequences.
4. Compute `size_bytes` using the provided `encoding`.

---

If this approach aligns with the intended behavior, I can open a PR with regression tests.

navyabijoy:
## Description

This PR addresses two logic errors in the [view_file](cci:1://file:///c:/Users/navya/Downloads/aden-hive/hive-main/tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py:13:4-76:62) tool that affect files containing multibyte characters (e.g., UTF-8 emojis, CJK text) or files using non-UTF-8 encodings (e.g., UTF-16).

## Resolves Issue

Fixes #5163 

## Type of Change

- [ ] Documentation update

## Changes Made

- **Refined Truncation Logic:** Replaced bit-naive character slicing (`content[:max_size]`) with byte-buffer slicing to ensure max_size (bytes) is strictly respected for multibyte encodings.
- **Multibyte Safety:** Implemented `errors="ignore"` on the truncation boundary to prevent `UnicodeDecodeError` crashes caused by splitting multibyte sequences mid-way.
- **Dynamic Metadata Encoding:** Updated size_bytes return field to calculate byte length based on the requested encoding parameter rather than hardcoded UTF-8.

## Testing

- [x] Unit tests pass (`cd tools && uv run pytest tests/tools/test_file_system_toolkits.py -v`)
- [x] Lint passes (`python -m ruff check tools/src/aden_tools/tools/file_system_toolkits/view_file/view_file.py`)
- [x] Manual testing performed (Verified with standalone regression tests covering UTF-8 emojis, UTF-16 BOM/size offsets, and Latin-1).

## Checklist

- [x] I have made corresponding changes to the documentation (N/A)


navyabijoy:
hello maintainers @TimothyZhang7 @RichardTang-Aden @bryanadenhq @Hundao @austin931114 , I have created a PR for the issue that was assigned to me, can you please review it ?

Hundao:
The view_file tool has been removed, so this PR is no longer applicable. The fix itself was solid though, appreciate the effort and the test coverage. Closing.

PR #5896, opened 2026-03-05
fix: add minimax provider mapping and stream fallback

VasuBansal7576:
## Description

This PR adds MiniMax support in Hive’s LLM path by wiring provider-specific API key/env mapping and adding a safe non-stream fallback for MiniMax responses when LiteLLM stream chunk parsing fails.

## Type of Change

## Related Issues

Fixes #5642

## Changes Made

- Added MiniMax API key/env mapping support in runner config resolution.
- Added MiniMax stream fallback path in LiteLLM provider to avoid stream parser failures.
- Added/updated tests for:
  - MiniMax env var mapping
  - MiniMax stream fallback behavior

## Testing

### Executed
- `uv run --project core pytest -q core/tests/test_runner_api_key_env_var.py core/tests/test_litellm_provider.py -k minimax`
- Result: `4 passed, 60 deselected`

## Checklist

### DIFF
--- a/core/framework/runner/runner.py
+++ b/core/framework/runner/runner.py
@@ -956,11 +956,19 @@ class AgentRunner:
 
         # Fallback: load from agent.json (legacy JSON-based agents)
         agent_json_path = agent_path / "agent.json"
-        if not agent_json_path.exists():
+        if not agent_json_path.is_file():
             raise FileNotFoundError(f"No agent.py or agent.json found in {agent_path}")
 
         with open(agent_json_path, encoding="utf-8") as f:
-            graph, goal = load_agent_export(f.read())
+            export_data = f.read()
+
+        if not export_data.strip():
+            raise ValueError(f"Empty agent export file: {agent_json_path}")
+
+        try:
+            graph, goal = load_agent_export(export_data)
+        except json.JSONDecodeError as exc:
+            raise ValueError(f"Invalid JSON in agent export file: {agent_json_path}") from exc
 
         return cls(
             agent_path=agent_path,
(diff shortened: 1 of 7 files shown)


TimothyZhang7 (changes requested):
Did not see Minimax provider configuration in LiteLLMProvider
No quickstart menu selection
No minimax coding subscription support

Note: extra LLM provider support must be wired to ~/.hive/configuration.json

vakrahul (commented):
"I think removing the with open()... context manager here might be risky for large agent exports. Using .read_text().strip() loads the entire JSON blob into a raw string first. Also, if the file exists but is empty, raising a FileNotFoundError will be confusing for debugging. We should probably keep the standard stream loading or raise a ValueError if empty!"

core/framework/runner/runner.py
```diff
@@ -956,14 +956,11 @@ def load(
 
         # Fallback: load from agent.json (legacy JSON-based agents)
         agent_json_path = agent_path / "agent.json"
-        if not agent_json_path.is_file():
+        if not agent_json_path.exists():
```
Antiarin: revert `exists()` back to `is_file() `here, a directory named `agent.json` (legacy) will slip through and crash with IsADirectoryError instead of the clean FileNotFoundError                                                                                                                                       
                                                                                                                                                                                                                                                                                                       
  also the empty file check got dropped below, empty files will just blow up with a raw JSONDecodeError now. 
nit: raising `FileNotFoundError` for an empty file is kinda weird since the file literally exists, `ValueError` would make more sense there

VasuBansal7576: Thanks @Antiarin , switched back to is_file(), restored empty-file check, and now raise ValueError for empty/invalid JSON with file path context.

TimothyZhang7 (approved):

PR #5962, opened 2026-03-06
fix: queen bee multiple response error resolved

AryanNandanwar:
## Description

Queen Bee was sometimes posting multiple, near-duplicate replies for a single user message in the web UI. Each Queen streaming event was converted into a ChatMessage with a unique id. This change gives Queen streaming events a stable per‑execution message ID, so all snapshots for one run collapse into a single chat bubble.

## Type of Change

## Related Issues

Fixes #5759 

## Changes Made

Updated handleSSEEvent in core/frontend/src/pages/workspace.tsx so that:

For Queen events (stream_id === "queen") of type client_output_delta or llm_text_delta, we override the auto-generated ChatMessage.id with a stable ID of the form queen-stream-<execution_id | node_id>.

This causes upsertChatMessage to update the same Queen message in place instead of appending new messages for each snapshot.

Left worker behavior unchanged (workers still get distinct message IDs per iteration, so each step remains a separate bubble).

## Testing

## Checklist


AryanNandanwar:
@Hundao @RichardTang-Aden @bryanadenhq @TimothyZhang7 
Could you please take a look at this when you have time and review it?

Hundao (changes requested):
Good idea, using a stable ID to collapse Queen snapshots into one bubble is the right approach. A few things to fix:

1. The `<>...</>` fragment wrapping is unnecessary. The `div` is already a single root element.

2. The fallback chain `eid || nid || "queen"` is a bit off. `nid` defaults to `"queen"` via `?? "queen"`, so the third `|| "queen"` is dead code. More importantly, if `execution_id` is empty, it falls back to `node_id`, which could be the same across different user messages (same Queen node). That means different replies would overwrite each other.

Simpler fix: only apply the stable ID when `execution_id` is present.

```js
if (isQueen && (event.type === "client_output_delta" || event.type === "llm_text_delta") && event.execution_id) {
  chatMsg.id = `queen-stream-${event.execution_id}`;
}
```

This way it collapses correctly when we have an execution_id, and falls back to default behavior (no override) when we don't.

AryanNandanwar:
Thanks for the feedback!
I've updated the implementation to only apply the stable ID when execution_id is present, as suggested.
The fallback logic has been removed.
Please take another look when you get a chance.

@Hundao

Hundao (changes requested):
Hey Aryan, the stable ID fix looks good — but it looks like the refactor accidentally removed two pieces of existing behavior:

1. **`chatMsg.phase` is gone** — the original block sets `chatMsg.phase = queenPhaseRef.current[agentType]` for Queen messages. Right now only `role` is being set. Without this the UI won't know which phase (planning/executing/etc.) the Queen is in.

2. **`reconcileOptimisticUser` is gone** — the original `upsertChatMessage` call passes `{ reconcileOptimisticUser: event.type === "client_input_received" }`. This prevents user messages from showing up twice (once optimistic, once confirmed). Dropping it could bring back a duplicate message bug.

Both of these are unrelated to the stable ID change so they should stay as-is. Could you put them back?

Also two stray blank lines snuck in (around lines 2768 and 3114) — mind cleaning those up?

The core fix is solid though, nice catch on the root cause.

AryanNandanwar:
Hey @Hundao ,

I have added chatMsg.phase and reconcileOptimisticUser. Also, I tried to clean up blank lines wherever I could find.
Let me know if I need to modify something else.

Thanks!

Hundao:
@AryanNandanwar Thanks for the fix! Pushed a small formatting cleanup and merged.

PR #6042, opened 2026-03-08
fix: handleKeyError in slack _get_user_token for unregistered credential

Amdev-5:
#6041 
## Summary                                             
  
   - `_get_user_token()` in `slack_tool.py` calls `credentials.get("slack_user")`, but `"slack_user"` is not registered in       
   `CREDENTIAL_SPECS` (only `"slack"` is)                                                                                        
   - `CredentialStoreAdapter.get()` raises `KeyError` for unregistered credential names, crashing any code path that calls
   `_get_user_token()` (e.g. `slack_search`)
   - Added `try/except KeyError` to gracefully fall back to `SLACK_USER_TOKEN` env var, matching the existing fallback pattern

   ## Root Cause

   `CREDENTIAL_SPECS` in `slack.py` only defines `"slack"` (bot token). The `"slack_user"` name was never registered, so
   `credentials.get("slack_user")` hits the unknown-name guard in `store_adapter.py` (line 108-109) and raises `KeyError`.

   Confirmed this pattern is unique to the Slack tool — all other tools only call `credentials.get()` with registered
   credential names.

   Fixes #6041

### DIFF
--- a/tools/src/aden_tools/tools/slack_tool/slack_tool.py
+++ b/tools/src/aden_tools/tools/slack_tool/slack_tool.py
@@ -1164,7 +1164,10 @@ def register_tools(
     def _get_user_token() -> str | None:
         """Get Slack user token for search API."""
         if credentials is not None:
-            return credentials.get("slack_user")
+            try:
+                return credentials.get("slack_user")
+            except KeyError:
+                return os.getenv("SLACK_USER_TOKEN")
         return os.getenv("SLACK_USER_TOKEN")
 
     def _get_client(account: str = "") -> _SlackClient | dict[str, str]:


tools/src/aden_tools/tools/slack_tool/slack_tool.py
```diff
@@ -1164,7 +1164,10 @@ def _get_token(account: str = "") -> str | None:
     def _get_user_token() -> str | None:
         """Get Slack user token for search API."""
         if credentials is not None:
-            return credentials.get("slack_user")
+            try:
```
Antiarin, line 1167: Hey, the try/except KeyError masks the real problem. "slack_user" was never registered in CREDENTIAL_SPECS, so credentials.get("slack_user") always raises KeyError and the credential store is permanently dead for user tokens.                                                                    
                                                                                                                                                                                                                                                                                                       
  Three things to fix:

  1. Remove slack_search_messages from the "slack" (bot token) spec in credentials/slack.py, since search doesn't use the bot token. The spec itself even notes search:read (requires user token) in the instructions.
  2. Register "slack_user" in SLACK_CREDENTIALS, like PagerDuty does with pagerduty_from_email:

```
  "slack_user": CredentialSpec(
      env_var="SLACK_USER_TOKEN",
      required=False,
      startup_required=False,
      direct_api_key_supported=True,
      tools=["slack_search_messages"],
  ),
```

  Then credentials.get("slack_user") returns None cleanly, and the original code works without any try/except.

  3. _user_headers() shouldn't fall back to the bot token. Slack never accepts bot tokens for search.messages (https://api.slack.com/methods/search.messages), so self._user_token or self._token just wastes an API call before failing.

Antiarin:
Hey @Amdev-5, can you please fix these?

Amdev-5:
> Hey @Amdev-5, can you please fix these?                                                

Yeah sure

Antiarin:
Any updates @Amdev-5 ?

PR #6062, opened 2026-03-09
Add light/dark mode toggle to web UI header

### LINKED ISSUE #6035: [Feature]: Light/Dark mode toggle for the web UI
## Problem Statement

The current Hive web UI appears to use a **dark theme by default**, with no option to switch to a light theme.

While dark mode works well for many users, some people find **light mode easier to read**, especially in bright environments or during long reading sessions. Users who prefer light backgrounds currently have no way to change the UI theme.

## Proposed Solution

Add a **theme toggle** to allow users to switch between:

- Dark mode (current default)
- Light mode

The toggle could be placed in a common UI location such as the header or a settings menu. Ideally, the selected theme should persist across sessions so the user's preference is remembered.

## Alternatives Considered

One alternative would be to automatically follow the user's system color preference using the `prefers-color-scheme` CSS media query. However, allowing users to manually override the theme with a toggle would provide more flexibility.

## Implementation Ideas

Possible implementation approaches:

- Add a **toggle button** in the UI that switches between dark and light themes.
- Remember the user's theme preference so the UI loads the same theme the next time they visit.
- Optionally default to the user's system theme preference on first load.

Akashstztt:
This PR adds a light/dark theme toggle to the Hive web UI.

Features:
- Toggle button added to the header
- Allows switching between dark and light mode
- Theme preference stored in localStorage
- Theme persists across sessions

Fixes #6035

## Dark Mode
Screenshot 2026-03-09 131957

## Light Mode
Screenshot 2026-03-09 132007

## Summary by CodeRabbit

* **New Features**
  * Theme toggle button added to the top navigation bar for switching between light and dark modes.
  * Theme preferences are automatically saved and persist across sessions.
  * System color scheme preferences are respected on initial app load.
  * Dynamic theme switching updates the visual appearance across the entire application.

### DIFF
--- a/core/frontend/src/components/ThemeToggle.tsx
+++ b/core/frontend/src/components/ThemeToggle.tsx
@@ -1,0 +1,20 @@
+import { Sun, Moon } from "lucide-react";
+import { useTheme } from "../context/ThemeContext";
+
+export default function ThemeToggle() {
+  const { theme, setTheme } = useTheme();
+
+  return (
+    <button
+      onClick={() => setTheme(theme === "dark" ? "light" : "dark")}
+      className="px-3 py-1 text-xs border rounded-md flex items-center gap-1.5 text-muted-foreground hover:text-foreground hover:bg-muted/50 transition-colors"
+      aria-label="Toggle theme"
+    >
+      {theme === "dark" ? (
+        <Sun className="w-3.5 h-3.5" />
+      ) : (
+        <Moon className="w-3.5 h-3.5" />
+      )}
+    </button>
+  );
+}
--- a/core/frontend/src/components/TopBar.tsx
+++ b/core/frontend/src/components/TopBar.tsx
@@ -1,8 +1,14 @@
 import { useState, useCallback } from "react";
 import { useNavigate } from "react-router-dom";
 import { Crown, X } from "lucide-react";
+import {
+  loadPersistedTabs,
+  savePersistedTabs,
+  TAB_STORAGE_KEY,
+  type PersistedTabState,
+} from "@/lib/tab-persistence";
+import ThemeToggle from "./ThemeToggle";
 import { sessionsApi } from "@/api/sessions";
-import { loadPersistedTabs, savePersistedTabs, TAB_STORAGE_KEY, type PersistedTabState } from "@/lib/tab-persistence";
 import BrowserStatusBadge from "@/components/BrowserStatusBadge";
 
 export interface TopBarTab {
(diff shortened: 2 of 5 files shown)


Antiarin (changes requested):
Hey, thanks for picking this up! A couple things:                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                       
  The theme toggle itself is a nice addition, but I don't think TopBar is the right place to own theme state. Theme is global, so if we ever need to read or set it from somewhere else (like a settings page), we'd be stuck. The CSS already uses the right variable pattern (hsl(var(--background)), .dark class, etc.), so the theming infra is ready to go. 
 
 Just need a ThemeProvider context that wraps the app and toggles the dark class on documentElement + persists to localStorage, and a small ThemeToggle component that consumes it. TopBar renders the button but doesn't manage the state. 
                                                                                                                                                                                                                                                                                                     
  Also, this PR has a lot of unrelated changes mixed in: the JSDoc comments on TopBarProps got removed, inline comments in handleCloseTab are stripped out, and there are whitespace-only changes throughout. Could you revert those so the diff only shows the actual feature?

Akashstztt:
Thanks for the feedback. I’ve updated the implementation.

   -

   Moved theme state to a global ThemeProvider context
   -

   Created a ThemeToggle component that consumes the context
   -

   TopBar now only renders the toggle and no longer manages theme state
   -

   Removed unrelated changes so the diff only contains the theme feature

Please let me know if anything else needs adjustment.

core/frontend/src/components/TopBar.tsx
```diff
@@ -12,150 +18,181 @@ export interface TopBarTab {
 }
 
 interface TopBarProps {
-  /** Live tabs from workspace state. When omitted, reads from localStorage. */
```
Antiarin, line 16: Please re-add these comments

core/frontend/src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,32 @@
+import { createContext, useContext, useEffect, useState } from "react";
+
+type Theme = "light" | "dark";
+
+const ThemeContext = createContext<any>(null);
```
Antiarin: Use proper types here. Define an interface for the context value:

```
  interface ThemeContextValue {
    theme: Theme;
    setTheme: (theme: Theme) => void;
  }

  const ThemeContext = createContext<ThemeContextValue | null>(null);
```

core/frontend/src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }, [theme]);
+
+  return (
+    <ThemeContext.Provider value={{ theme, setTheme }}>
+      {children}
+    </ThemeContext.Provider>
+  );
+}
+
+export function useTheme() {
```
Antiarin: Add a return type and a null guard. If ThemeToggle ever gets rendered outside of ThemeProvider, this crashes with a generic TypeError instead of telling you what's wrong:

  ```
export function useTheme(): ThemeContextValue {
    const ctx = useContext(ThemeContext);
    if (!ctx) throw new Error("useTheme must be used within a ThemeProvider");
    return ctx;
  }
```

File: core/frontend/src/components/TopBar.tsx [hunk 4]
```diff
@@ -12,150 +18,181 @@ export interface TopBarTab {
...
+  onTabClick,
+  onCloseTab,
+  canCloseTabs,
+  afterTabs,
+  children,
+}: TopBarProps) {
   const navigate = useNavigate();
 
-  // Fallback: read persisted tabs when no live tabs provided
+  /* ---------------- EXISTING CODE ---------------- */
```
Antiarin: This looks like a leftover editing marker. Remove it.

Antiarin (changes requested):
1. ThemeProvider.tsx shouldn't be in components/, it's a context provider not a UI component. Make a `src/context/` folder and move it there as ThemeContext.tsx. That way it's separate from the plain hooks in hooks/ and the actual UI components.                                                  
  2. There's a bunch of unrelated stuff in the TopBar diff. `flex-shrink-0` got removed from the outer bar, logo, separator, tab buttons, and children container. That class is there so things don't get squished when you have a lot of tabs open. `hover:!opacity-100` on the close button makes the X  go fully opaque when you hover right on it. These are intentional, please revert them so the diff only has the theme toggle changes.                                                                                                                                                                 
  3. All the JSDoc and inline comments got removed from TopBar. The prop docs explain stuff like the localStorage fallback on tabs, the default for canCloseTabs, and the difference between afterTabs and children. The comments in handleCloseTab explain the session kill side-effect and fire-and-forget pattern. Please put them back.

core/frontend/src/components/ThemeToggle.tsx
```diff
@@ -0,0 +1,15 @@
...
+export default function ThemeToggle() {
+
+  const { theme, setTheme } = useTheme();
+
+  return (
+    <button
+      onClick={() => setTheme(theme === "dark" ? "light" : "dark")}
+      className="px-3 py-1 text-xs border rounded-md"
+    >
+      {theme === "dark" ? "☀" : "🌙"}
```
Antiarin: nit: use Lucide icons (Sun, Moon) instead of emoji

Akashstztt:
Hey @Antiarin, thanks for the review! I've addressed all the feedback - here's a summary of what was changed:

**ThemeContext** (`src/context/ThemeContext.tsx`)
- Moved out of `components/` into a new `src/context/` folder and renamed to `ThemeContext.tsx`
- Replaced `createContext<any>` with a proper `ThemeContextValue` interface
- Added return type and null guard to `useTheme()` so it throws a descriptive error if used outside the provider

**ThemeToggle** (`src/components/ThemeToggle.tsx`)
- Replaced `*` / 🌙 emoji with Lucide `<Sun>` and `<Moon>` icons
- Updated import path to point to the new `src/context/ThemeContext`

**TopBar** (`src/components/TopBar.tsx`)
- Removed the leftover `/* EXISTING CODE */` editing marker
- Restored all JSDoc comments on props (`tabs`, `onTabClick`, `onCloseTab`, `canCloseTabs`, `afterTabs`, `children`)
- Restored the fire-and-forget comment in `handleCloseTab`
- Reverted the unintentional `flex-shrink-0` removals on the outer bar, logo, separator, tab buttons, and children container
- Restored `hover:!opacity-100` on the close button

Let me know if anything else needs to be changed.

core/frontend/src/App.tsx
```diff
@@ -1,15 +1,18 @@
 import { Routes, Route } from "react-router-dom";
+import { ThemeProvider } from "./context/ThemeContext";
 import Home from "./pages/home";
 import MyAgents from "./pages/my-agents";
 import Workspace from "./pages/workspace";
 
 function App() {
   return (
+    <ThemeProvider>
```
Antiarin: main.tsx is where our global providers live (BrowserRouter is already there). having it inside App.tsx wrapping Routes works today but if we ever render something outside App it won't have theme access. just bump it up one level:
                                                                                                                                                                                                                                                                                                 
```
  <ThemeProvider>
    <BrowserRouter>                                                                                                                                                                                                                                                                                    
      <App />     
    </BrowserRouter>                                                                                                                                                                                                                                                                                   
  </ThemeProvider>
```
  and keep App.tsx as just routes.

core/frontend/src/components/TopBar.tsx
```diff
@@ -124,38 +157,47 @@ export default function TopBar({ tabs: tabsProp, onTabClick, onCloseTab, canClos
...
+        {/* THEME BUTTON */}
+        <ThemeToggle />
+
+        {children}
+      </div>
     </div>
   );
 }
 
-/** Derive TopBarTab[] from persisted localStorage state (used outside workspace). */
```
Antiarin, line 145: revert these comments

core/frontend/src/components/TopBar.tsx
```diff
@@ -124,38 +157,47 @@ export default function TopBar({ tabs: tabsProp, onTabClick, onCloseTab, canClos
...
+
+    const hasRunning =
+      sessionData?.graphNodes?.some(
+        (n) => n.status === "running" || n.status === "looping",
+      ) ?? false;
+
     tabs.push({
       agentType: tab.agentType,
       label: tab.label,
-      isActive: false, // no active tab outside workspace
```
Antiarin, line 160: same as above

Antiarin:
index.html still has `class="dark"` hardcoded. ThemeProvider sets the class in useEffect which runs after paint, so light mode users get a dark flash on every reload (FOUC). just make it <html lang="en"> and let ThemeProvider handle it. shadcn's Vite guide does this: https://ui.shadcn.com/docs/dark-mode/vite                                                                                                                                                                                                                                                            
   
  if you want zero flash, add a blocking script in <head> that reads localStorage before paint (https://notanumber.in/blog/fixing-react-dark-mode-flickering), but optional for now.                                                                                                                   
                  
  inline comments are still missing. JSDoc on TopBarProps came back but these are still gone:

  - // Fallback: read persisted tabs when no live tabs provided
  - // Kill the backend session (queen/judge/worker) even outside workspace
  - // fire-and-forget
  - // Fallback: update localStorage directly (non-workspace pages)
  - /** Derive TopBarTab[] from persisted localStorage state (used outside workspace). */
  - // no active tab outside workspace

  these explain non-obvious behavior, please restore them.

core/frontend/src/components/TopBar.tsx
Akashstztt: Hi @Antiarin , these comments were already re-added in a previous commit. They should be visible on line 21. Please let me know if anything else needs to be updated.

core/frontend/src/components/TopBar.tsx
Akashstztt: Yes, sure @Antiarin. I am reverting these comments.

core/frontend/src/components/TopBar.tsx
Akashstztt: Yes, sure @Antiarin. I am reverting these comments.

core/frontend/src/App.tsx
Akashstztt: Thanks for the suggestion. I’ve moved the ThemeProvider to main.tsx and kept App.tsx focused on routes.

Akashstztt:
Good catch.  I’ve removed the hardcoded "dark" class from index.html so the ThemeProvider handles the theme. I’ll also restore the missing inline comments.

Antiarin:
@Akashstztt can you fix the conflicts

Akashstztt:
Hi @Antiarin , I have resolved the merge conflicts and updated the code. Please review.

Antiarin (changes requested):
Hey @Akashstztt, looks like the merge conflict resolution introduced a couple of regressions:                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                       
  agent_path.endsWith(agentType) got changed to agent_path === agentType in handleCloseTab. The .endsWith() was there on purpose because agent_path is a full path and agentType is just the leaf name. With strict equality it'll never match, so closing a tab outside the workspace won't kill the
  backend session. Please revert to .endsWith().                                                                                                                                                                                                                                                       
                  
  The comment says "queen/judge/worker" but main has "queen/worker". "judge" was removed in an earlier refactor. Please keep it as "queen/worker".                                                                                                                                                     
                  
  Both are likely from the conflict resolution, just need a quick fix.                                                                                                                                                                                                                                 
   
  Also, please soft reset to main and force push as one single commit.

Akashstztt:
Hi @Antiarin, thanks for catching that!

You're right , those regressions came from the merge conflict resolution. I've now:

* Restored `agent_path.endsWith(agentType)` in `handleCloseTab`
* Updated the comment to "queen/worker" to match current main
* Double-checked related logic to avoid similar issues

Also rebased/squashed the branch and force-pushed as a single clean commit.

Please take another look 🙌

Akashstztt:
Hi @Antiarin

Just a gentle follow-up on this PR. I've incorporated all the requested changes and cleaned up the history into a single commit.

Whenever you have time, could you please review it again? Appreciate your help!

Antiarin:
@Hundao @RichardTang-Aden @TimothyZhang7  lgtm!

Antiarin (approved):

PR #6070, opened 2026-03-09
 fix: prevent mid-stream retry from duplicating published tokens (#5923)

### LINKED ISSUE #5923: [Bug]: LiteLLMProvider.stream() retries mid-stream without guard — clients receive duplicated partial response concatenated with full retry response
## Summary

`LiteLLMProvider.stream()` in `core/framework/llm/litellm.py:816–1015` retries on transient errors and rate limits without checking whether `TextDeltaEvent`s were already yielded and published to the event bus. When an error fires after K chunks have streamed, the retry replays the full response from the beginning — permanently concatenating the partial first attempt with the complete second attempt in the client's UI stream.

`EventBus.publish()` is fire-and-forget with no retract, reset, or replay mechanism. The corruption is irreversible once the first chunk hits the bus.

## Root Cause

The retry loop resets accumulators at the top of each attempt:
```python
# litellm.py:816–826
    accumulated_text = ""        # ← reset on every attempt
    tool_calls_acc = {}          # ← reset on every attempt
```

`TextDeltaEvent`s are yielded immediately as chunks arrive — before the stream completes:
```python
# litellm.py:839–844
yield TextDeltaEvent(            # ← published to event bus immediately, NOT buffered
    content=delta.content,
    snapshot=accumulated_text,
)
```

Both exception handlers retry unconditionally with no guard on `accumulated_text`:
```python
# litellm.py:989 — RateLimitError
except RateLimitError as e:
    if attempt < RATE_LIMIT_MAX_RETRIES:
        await asyncio.sleep(wait)
        continue                 # ← no check: were chunks already published?

# litellm.py:1002 — transient errors
except Exception as e:
        await asyncio.sleep(wait)
        continue                 # ← same: no guard
```

## User-Visible Effect

When a connection reset or 529 fires after 30-50 tokens have streamed:

**Attempt 0 (partial, before error):**
> "I'll analyze your repository and check—"

**Attempt 1 (full retry from beginning):**
> "I'll analyze your repository and check for any issues. Let me start by searching for..."

**Client receives:**
> "I'll analyze your repository and check—I'll analyze your repository and check for any issues. Let me start by searching for..."

Users see this as the agent "repeating itself" with no error message or indication a retry occurred.

## Blast Radius

| Scenario | Trigger | Client effect |
|----------|---------|---------------|
| RateLimitError at chunk 50 | Provider streams 49 chunks then raises 429 | Partial + full response concatenated |
| Transient error at chunk 1 | ConnectionReset after first character | Single character + full response |
| 10 retries, each failing after K chunks | Persistent transient failure | K-chunk partial × 11 before terminal error |
| Tool-call-only response | Exception mid-argument delta | **Safe** — tool deltas are buffered, never yielded as `TextDeltaEvent` |

**Internal state is always correct.** `accumulated_text = event.snapshot` in `_do_stream()` (line 1658) is an assignment that ends up holding the full text of the last successful attempt. Conversation history, tool call results, and L3 logs are unaffected. Only the client-facing event bus stream is corrupted.

**`RATE_LIMIT_MAX_RETRIES = 10`** means up to 11 concatenated partial attempts before a terminal error — worst case on a flapping connection.

## Complete Call Chain
```
LiteLLMProvider.stream()          litellm.py:816
  yield TextDeltaEvent()           litellm.py:841
EventLoopNode._do_stream()         event_loop_node.py:1658
  _publish_text_delta()            event_loop_node.py:1659
    emit_client_output_delta()     event_bus.py:692   ← UI stream (permanent)
    OR emit_llm_text_delta()       event_bus.py:571   ← observability pipeline
      EventBus.publish()           fire-and-forget, no retract exists
```

## No Existing Tests Cover Mid-Stream Retry

All six tests in `TestTransientErrorRetry` (`test_event_loop_node.py:1288–1486`) use a mock LLM that raises **before** yielding any `TextDeltaEvent` — the safe, non-duplicating case. `test_litellm_streaming.py` is entirely `@pytest.mark.skip(reason="Requires valid live API keys")`.

## Proposed Fix

Add a guard in both exception handlers before `continue`. If content was already published, yield `StreamErrorEvent(recoverable=True)` instead of retrying — the existing outer retry loop in `EventLoopNode` (`event_loop_node.py:661`) resets `accumulated_text` and restarts the turn cleanly without touching the already-published stream:
```python
# litellm.py:989 — RateLimitError handler
except RateLimitError as e:
    if accumulated_text or tool_calls_acc:
        yield StreamErrorEvent(error=str(e), recoverable=True)
        return
    if attempt < RATE_LIMIT_MAX_RETRIES:
        await asyncio.sleep(wait)
        continue
    yield StreamErrorEvent(error=str(e), recoverable=False)
    return

# litellm.py:1002 — transient error handler  
except Exception as e:
    if (_is_stream_transient_error(e)
            and attempt < RATE_LIMIT_MAX_RETRIES
            and not accumulated_text
            and not tool_calls_acc):
        await asyncio.sleep(wait)
        continue
    recoverable = _is_stream_transient_error(e)
    yield StreamErrorEvent(error=str(e), recoverable=recoverable)
    return
```

`StreamErrorEvent(recoverable=True)` reaches `_do_stream():1678–1682`, sets `_stream_error`, raises `ConnectionError` at line 1706, and the outer retry loop at `event_loop_node.py:661` handles clean restart — no duplication, no new infrastructure needed.

## Files Affected

- `core/framework/llm/litellm.py:989–1015` — both exception handlers missing the guard
- `core/tests/test_event_loop_node.py:1288–1486` — existing retry tests cover pre-yield errors only

EsraaKamel11:
## Summary

Fixes #5923 — when the LiteLLM streaming layer retried after a mid-stream
`RateLimitError` or transient error, it re-streamed from token 1, duplicating
content that had already been yielded to callers and published to the event bus.
Since published events cannot be recalled, the retry must be abandoned when a
partial stream has already been emitted.

## Root Cause

`LiteLLMProvider.stream()` has an internal retry loop for transient errors. Both
the `RateLimitError` handler and the generic transient-error handler would
unconditionally `continue` — restarting the entire stream from the beginning —
even when `accumulated_text` was non-empty (i.e., chunks had already been
yielded upstream and emitted on the event bus).

Before — both handlers did this unconditionally:
```python
except RateLimitError as e:
    if attempt < RATE_LIMIT_MAX_RETRIES:
        wait = _compute_retry_delay(...)
        await asyncio.sleep(wait)
        continue   # re-streams from token 0, duplicating all prior output
```

The event bus publishes token deltas eagerly as they stream in. There is no
mechanism to retract already-published events, so retrying produced a second
copy of every token the client had already received.

## Fix

In both exception handlers, check `accumulated_text` before retrying. If any
text has already been yielded, emit a `recoverable=True` `StreamErrorEvent` and
return immediately. `EventLoopNode`'s existing empty-response guard at line 1706
detects the non-empty `accumulated_text` and suppresses the outer retry,
preserving the partial turn.
```python
except RateLimitError as e:
    if attempt < RATE_LIMIT_MAX_RETRIES:
        if accumulated_text:
            # Text already published to event bus — cannot be recalled.
            # Yield recoverable error; EventLoopNode will commit the partial
            # text and skip the outer retry (accumulated_text non-empty).
            yield StreamErrorEvent(error=str(e), recoverable=True)
            return
        wait = _compute_retry_delay(attempt, exception=e)
        ...
        continue

except Exception as e:
        if accumulated_text:
            yield StreamErrorEvent(error=str(e), recoverable=True)
            return
        ...
        continue
```

## Changes

| File | Change |
|------|--------|
| `core/framework/llm/litellm.py` | Add `accumulated_text` guard to `RateLimitError` handler (L989) and transient `Exception` handler (L1012) |
| `core/tests/test_event_loop_node.py` | Add `PartialStreamThenErrorLLM` helper and `TestMidStreamRetryNoDuplication` class with 5 new tests |

## Tests

5 new automated tests added to `TestMidStreamRetryNoDuplication` in `core/tests/test_event_loop_node.py` — all passing (74/74 total):

- `test_mid_stream_error_no_duplicate_deltas_3_chunks` — 3 chunks + error -> exactly 3 deltas on bus, no outer retry
- `test_mid_stream_error_no_duplicate_deltas_50_chunks` — 50 chunks + error -> exactly 50 deltas, no outer retry
- `test_mid_stream_error_at_chunk_0_triggers_outer_retry` — error before first chunk -> outer retry fires correctly, no duplication
- `test_mid_stream_tool_only_error_inner_retry_unaffected` — tool-only error -> inner retry safe, guard does not block
- `test_mid_stream_recoverable_error_partial_text_committed` — partial text committed to history, `_call_index == 1`

## Notes

- Guard uses `accumulated_text` only, not `tool_calls_acc` — tool deltas are buffered locally and never published before stream completion, so mid-tool-stream errors remain safe to retry internally
- The empty-stream retry path (L1019-1020) is correctly guarded by `not has_content` and is unaffected
- `EventLoopNode` empty-response guard at `runtime_logger.py:1706` is the cooperating mechanism that absorbs the early exit without crashing the turn

### DIFF
--- a/core/framework/credentials/storage.py
+++ b/core/framework/credentials/storage.py
@@ -13,12 +13,16 @@ from __future__ import annotations
 import json
 import logging
 import os
+import threading
 from abc import ABC, abstractmethod
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
 
+from filelock import FileLock
 from pydantic import SecretStr
+
+from framework.utils.io import atomic_write
 
 from .models import CredentialDecryptionError, CredentialKey, CredentialObject, CredentialType
 
@@ -160,6 +164,8 @@ class EncryptedFileStorage(CredentialStorage):
                 )
 
         self._fernet = Fernet(self._key)
+        self._lock = threading.RLock()
+        self._index_lock_path = self.base_path / "metadata" / "index.lock"
 
     def _ensure_dirs(self) -> None:
         """Create directory structure."""
@@ -181,9 +187,9 @@ class EncryptedFileStorage(CredentialStorage):
         # Encrypt
         encrypted = self._fernet.encrypt(json_bytes)
 
-        # Write to file
+        # Write to file atomically (tmp + fsync + rename — no truncation crash window)
         cred_path = self._cred_path(credential.id)
-        with open(cred_path, "wb") as f:
+        with atomic_write(cred_path, mode="wb") as f:
             f.write(encrypted)
 
         # Update index
@@ -215,21 +221,32 @@ class EncryptedFileStorage(CredentialStorage):
     def delete(self, credential_id: str) -> bool:
         """Delete a credential file."""
         cred_path = self._cred_path(credential_id)
-        if cred_path.exists():
-            cred_path.unlink()
-            self._update_index(credential_id, "delete")
-            logger.debug(f"Deleted credential '{credential_id}'")
-            return True
-        return False
+        if not cred_path.exists():
+            return False
+        with self._lock:
+            with FileLock(self._index_lock_path):
+                # Remove from index before unlinking the .enc file.
+                # If unlink later fails, the index is already correct (no ghost entry).
+                self._update_index_locked(credential_id, "delete")
+                cred_path.unlink()
+        logger.debug(f"Deleted credential '{credential_id}'")
+        return True
 
     def list_all(self) -> list[str]:
         """List all credential IDs."""
         index_path = self.base_path / "metadata" / "index.json"
         if not index_path.exists():
             return []
-        with open(index_path, encoding="utf-8-sig") as f:
-            index = json.load(f)
-        return list(index.get("credentials", {}).keys())
+        try:
+            with open(index_path) as f:
+                index = json.load(f)
+            return list(index.get("credentials", {}).keys())
+        except (json.JSONDecodeError, OSError):
+            logger.warning(
+                "index.json is corrupt or unreadable; scanning credential files for recovery"
+            )
+            cred_dir = self.base_path / "credentials"
+            return [p.stem for p in cred_dir.glob("*.enc")]
 
     def exists(self, credential_id: str) -> bool:
         """Check if credential exists."""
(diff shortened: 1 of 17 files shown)


EsraaKamel11:
Superseded by a cleaner branch, reopening with single-commit diff

PR #6079, opened 2026-03-09
fix(credentials): remove orphaned google_sheets.py credential spec

Waryjustice:
## Summary

   Removes `tools/src/aden_tools/credentials/google_sheets.py`, which was completely
   dead code.

   ## Problem

   The file defined `GOOGLE_SHEETS_CREDENTIALS` — an API-key-based credential spec
   for reading public sheets via `GOOGLE_SHEETS_API_KEY` — but was never connected
   to anything:

   - Never imported in `credentials/__init__.py` (every other ~50 credential files are)
   - Never merged into `CREDENTIAL_SPECS`
   - Never listed in `__all__`
   - The tool (`google_sheets_tool.py`) never calls `credentials.get("google_sheets_key")` —
     it uses `credentials.get("google")` (OAuth2 via `email.py`)
   - The tool names listed in the spec (`sheets_get_spreadsheet`, `sheets_read_range`,
     `sheets_batch_read`) don't match any actual function names in the tool

   ## Fix

   The `"google"` credential in `email.py` already correctly covers all 10 Google Sheets
   tool functions under OAuth2 and is fully registered. No functionality is lost.

   Removing the orphaned file eliminates a misleading spec that would cause confusion
   for contributors trying to understand the credential system.

   ## Verification

   ```bash
   grep -r "GOOGLE_SHEETS_CREDENTIALS\|google_sheets_key\|GOOGLE_SHEETS_API_KEY" tools/
   # Returns nothing after this change

  Closes #6077


bryanadenhq (approved):
Looks good to me

PR #6095, opened 2026-03-10
Enhancement/browser history management | worker nodes reasoning collapsible UI

levxn:
## Browser UI — Chat Reliability, UX Polish, and Session Fixes

This PR delivers a batch of improvements to the browser-based interface covering conversation restore fidelity, scroll behavior, message ordering, drag-and-drop UX, keyboard accessibility, and several backend fixes for queen/worker session handling.

---

### Features

**Worker group collapse: auto-collapse during streaming**

Worker steps now automatically collapse while they are actively streaming. As each step streams in, it folds into the collapsed group so the chat panel does not grow excessively. The group re-expands when streaming is complete, and any manual open/close decision made by the user before streaming finishes is respected and not overridden.

**Introducing slash commands - currently resume feature**

The Resume Session picker (session history browser) is a new feature that now supports full keyboard navigation. Arrow Up / Arrow Down move the highlight through results. Enter selects the highlighted session.

**Re-arrange Tabs**

The tab reorder drag-and-drop is now position-aware. The drop target is detected by which half of a tab the pointer is over (left half = insert before, right half = insert after), making it much easier to place a tab precisely without pixel-perfect accuracy. A visual border indicator appears on the correct side of the target tab during the drag. The entire tab bar container acts as a drop zone so releasing the drag anywhere in the bar still registers.

---

### Bug Fixes

**Queen messages missing node ID — unstable message IDs across old sessions opened**

Queen conversation parts stored without a subdirectory had no `_node_id` set by `build_transcript`, so the frontend generated IDs of `backend-undefined-{seq}` or `backend-null-{seq}`. On every restore these messages could not be deduped against live-streamed versions, causing phantom duplicates. Fixed by defaulting `_node_id` to `phase_id` or `"queen"` in both `build_transcript` and `handle_queen_messages`, and by using `_source` (always `"queen"` or `"worker"`) instead of `_node_id` as the stable part of the message ID in the frontend.

**Message ordering broken after restore**

After a cold restore or live resume, messages could appear out of order. The previous merge logic prepended restored messages in front of the existing list without a global sort, and the `seq` fallback was accidentally used as an absolute timestamp (creating 1970-era dates). Fixed: all restore merges now perform a global sort keyed on `(createdAt, seq)` as a tiebreaker, and `seq` is stored as a separate field rather than coerced into the `createdAt` timestamp slot.

**Backend: `_node_id` not set for flat queen messages in live SSE stream**

The `handle_queen_messages` endpoint was emitting queen messages from flat `conversations/parts/` files without a `_node_id` field, causing the frontend to generate unstable or duplicate IDs when those messages were merged with transcript-restored versions of the same messages. The endpoint now applies the same `phase_id`-or-`"queen"` default as `build_transcript`.

**Fixed pause/resume feature in pipeline for GUI**

The mode_state parameter was never a part of session object, hence it was corrected with actual phase_state and as well fixed the bugs followed by to enable resume feature in pipeline

---

### Files Changed

Backend:
- `core/framework/server/session_manager.py` — `is_client_visible_message`, `build_transcript`, `_node_id` defaulting
- `core/framework/server/routes_sessions.py` — `handle_queen_messages` node ID default, `_session_to_live_dict` consistency

Frontend:
- `core/frontend/src/pages/workspace.tsx` — restore merge strategy, scroll position per thread, `loadingRef` cleanup
- `core/frontend/src/components/ChatPanel.tsx` — collapse auto-behaviour, accurate step count, scroll pinning logic
- `core/frontend/src/components/TopBar.tsx` — cursor fix, position-aware drag-to-reorder
- `core/frontend/src/components/ResumeSessionPicker.tsx` — keyboard navigation
- `core/frontend/src/lib/chat-helpers.ts` — stable message IDs, `seq` tiebreaker, `createdAt` fallback removed

This closes my issue #5904 
Screenshot 2026-03-10 at 6 10 00 AM
Screenshot 2026-03-10 at 6 10 10 AM
Screenshot 2026-03-10 at 6 11 05 AM
Screenshot_2026-03-09_at_7 59 42_PM


TimothyZhang7:
Several UI/UX issues
- Nobody will know they need to type /resume for the option to pop up, that option should show up whenever the user focuses on the input box
- Pushing "Enter" on the /resume selection will bring up the previous session selection, which is good, but people might also click it with the mouse (does not work yet)

core/framework/server/session_manager.py
```diff
@@ -823,6 +823,241 @@ def list_sessions(self) -> list[Session]:
...
+                            part.setdefault("created_at", part_file.stat().st_mtime)
+                            part["_source"] = "worker"
+                            transcript.append(part)
+                        except (json.JSONDecodeError, OSError):
+                            continue
+
+        # --- Filter ---
+        visible = [m for m in transcript if SessionManager.is_client_visible_message(m)]
+
+        # Worker-specific filtering: include all visible worker messages
```
Antiarin, line 1129: client_facing_nodes is passed in but never actually checked here. all worker user/assistant messages get through regardless of which node produced them, so internal stuff (health judge, routing, etc.) shows up in the chat. should be something like:                                             
                                                                                                                                                                                                                                                                                                     
```
  if m.get("_source") == "worker":                                                                                                                                                                                                                                                                     
      if client_facing_nodes is not None:                                                                                                                                                                                                                                                            
          keep = m["role"] == "user" or (                                                                                                                                                                                                                                                              
              m["role"] == "assistant" and m.get("_node_id") in client_facing_nodes                                                                                                                                                                                                                  
          )
      else:
          keep = m["role"] in ("user", "assistant")
```

core/framework/server/tests/test_api.py
```diff
@@ -14,7 +14,7 @@
 from aiohttp.test_utils import TestClient, TestServer
```
Antiarin, line 14: This test file isn't in the CI path btw. Workflow runs core/tests/ but this is in core/framework/server/tests/ so nothing here actually executes.

If you run them locally (uv run pytest framework/server/tests/test_api.py::TestQueenHistory -v) you'll see test_get_queen_messages_from_flat_parts_layout, test_history_preview_reads_flat_parts_layout, and test_get_messages_client_only_no_runner_returns_all all fail. missing is_client_input on user messages and message_count not being a real field in the response.

fix: Add `"is_client_input": True` to user messages in the test data, change `assert history["message_count"] == 2` to `assert history["last_role"] == "assistant"`, and either move this file to core/tests/ or add the path to CI.

core/frontend/src/pages/workspace.tsx
```diff
@@ -1460,16 +1553,47 @@ export default function Workspace() {
...
+                return {
+                  ...prev, [agentType]: {
+                    ...cur,
+                    isTyping: false, isStreaming: false, queenIsTyping: false, workerIsTyping: false,
+                    ...(keepWorkerQ ? {} : { awaitingInput: false, workerInputMessageId: null, pendingQuestion: null, pendingOptions: null, pendingQuestionSource: null }),
+                  }
+                };
+              });
+            } else {
+              updateAgentState(agentType, { isTyping: false, isStreaming: false, queenIsTyping: false, workerIsTyping: false, awaitingInput: false, workerInputMessageId: null, pendingQuestion: null, pendingOptions: null, pendingQuestionSource: null });
```
Antiarin: The worker branches for execution_paused, execution_failed, and node_loop_iteration all unconditionally null out pendingQuestion/pendingOptions/pendingQuestionSource.

If the queen has a question showing when any of these fire, it gets wiped. queen handlers guard against this with keepWorkerQ but no symmetric check on the worker side. also the two back to back updateAgentState calls here and in execution_failed can just be merged into one.

fix: mirror what the queen handlers do. use setAgentStates with a keepQueenQ guard:

```typescript
setAgentStates(prev => {
  const cur = prev[agentType] || defaultAgentState();
  const keepQueenQ = cur.pendingQuestionSource === "queen";
  return {
    ...prev, [agentType]: {
      ...cur,
      isTyping: false, isStreaming: false, queenIsTyping: false, workerIsTyping: false,
      workerRunState: "paused",
      currentExecutionId: null,
      pausedWorkerSessionId: event.execution_id || cur.currentExecutionId || null,
      ...(keepQueenQ ? {} : { awaitingInput: false, workerInputMessageId: null, pendingQuestion: null, pendingOptions: null, pendingQuestionSource: null }),
    }
  };
});
```

same pattern for execution_failed and node_loop_iteration.

levxn:
updating based on reviews and recent commits, will open this PR with a working solution soon, thanks :))

PR #6181, opened 2026-03-11
docs: clarify quickstart auto-launch is conditional on frontend build

Aayush-engineer:
Fixes #6177

## What changed
Updated README.md to clarify that the dashboard auto-launch only happens 
when the frontend build succeeds.

Added a troubleshooting note for users whose frontend build fails.

## Before
"At last, it will initiate the open hive interface in your browser"

## After
"If the frontend build succeeds, it will automatically open the Hive 
dashboard in your browser. If the build fails, run: 
cd core/frontend && npm run build, then relaunch with hive open"

## Type
- [x] Documentation fix


Hundao:
@Aayush-engineer The fix is too verbose. The fallback instructions make it hard to read, and the tip below already covers `hive open`. Simplify to just:

"If the frontend build succeeds, it will open the Hive dashboard in your browser"

Aayush-engineer:
> Collaborator
> ###

Updated! Simplified the wording as suggested. Thanks for the review!

Hundao:
Closing this one -- the original wording is accurate enough since the quickstart already handles the build-fail case. Thanks though!

PR #6185, opened 2026-03-11
fix: sessions with CANCELLED status should not be resumable

builtbyshekhar:
## Description
Updated the `is_resumable` logic in the `SessionState` schema. Previously, the property only checked if a status was not `COMPLETED`, which incorrectly allowed `CANCELLED` sessions to be flagged as resumable.

## Type of Change

## Related Issues
Fixes #4 (or the specific issue number from their board)

## Changes Made
- Modified `core/framework/schemas/session_state.py`
- Updated `@property is_resumable` to return `False` if status is `SessionStatus.CANCELLED`.
- Updated docstring to reflect the state machine logic for terminal states.

## Testing
- [x] Manual testing performed: Verified logic change in schema definition.
- [ ] Unit tests (Pending CI run)

## Checklist

### DIFF
--- a/core/framework/schemas/session_state.py
+++ b/core/framework/schemas/session_state.py
@@ -161,12 +161,11 @@ class SessionState(BaseModel):
     def is_resumable(self) -> bool:
         """Can this session be resumed?
 
-        Every non-completed session is resumable. If resume_from/paused_at
-        aren't set, the executor falls back to the graph entry point —
-        so we don't gate on those. Even catastrophic failures are resumable.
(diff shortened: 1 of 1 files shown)


Antiarin:
Thanks for the PR, but this was intentional. Cancelled sessions are kept resumable by design

builtbyshekhar:
Ah, that makes total sense! Thanks for the clarification, Antiarin. I'll close this PR to keep things clean.

I've also submitted a PR for the Light/Dark mode toggle (#6191)—really enjoyed working on the frontend side of Hive as well. Looking forward to your thoughts on that one when you have a moment

PR #6195, opened 2026-03-11
fix(llm): separate retry counters in stream() - transient errors used wrong cap

### LINKED ISSUE #6105: [Bug]: stream() transient error retries use wrong cap (RATE_LIMIT_MAX_RETRIES=10) and share attempt counter with empty-stream retries
## Summary
Two related retry logic bugs in the `stream()` method of `core/framework/llm/litellm.py` cause inconsistent retry behaviour across failure types.

## Bug 1: Wrong retry cap for transient errors

The `except Exception` transient error branch uses RATE_LIMIT_MAX_RETRIES (10) as its cap:
```python
except Exception as e:
```

Transient stream errors (network blips, APIConnectionError, timeouts) should not consume the rate-limit retry budget. The empty-stream path already has its own constant EMPTY_STREAM_MAX_RETRIES = 3 for this reason, transient errors need the same treatment.

**Impact:** A flaky network can stall a stream for several minutes (10 retries × exponential backoff up to 120s) before giving up.

## Bug 2: Shared `attempt` counter entangles all three retry budgets

All three retry paths share the same `attempt` variable from the outer loop.
Mixed failure sequences consume each other's budget, e.g. 3 empty-stream retries leave only 7 rate-limit retries remaining.

## Proposed Fix
- Add `STREAM_TRANSIENT_MAX_RETRIES = 3` constant
- Track `empty_stream_attempts`, `transient_attempts`, `rate_limit_attempts`
  as independent counters inside `stream()`
- Each branch checks and increments only its own counter

## Files Affected
`core/framework/llm/litellm.py`, `stream()` method and module-level constants

I'd like to work on this, happy to submit a PR once assigned.

Sri-Likhita-adru:
Closes #6105

## Changes

`core/framework/llm/litellm.py`
- Added `STREAM_TRANSIENT_MAX_RETRIES = 3` module-level constant
- In `stream()`, the `except Exception` transient error handler now checks `attempt < STREAM_TRANSIENT_MAX_RETRIES` instead of `attempt < RATE_LIMIT_MAX_RETRIES`, and the log line reflects the same constant


core/framework/llm/litellm.py
```diff
@@ -927,6 +928,11 @@ async def stream(
             kwargs.pop("max_tokens", None)
             kwargs.pop("stream_options", None)
 
+        # Independent retry counters — each path has its own budget
+        empty_stream_attempts = 0
+        transient_attempts = 0
+        rate_limit_attempts = 0
+
         for attempt in range(RATE_LIMIT_MAX_RETRIES + 1):
```
Antiarin, line 1632: The outer loop is still for attempt in range(RATE_LIMIT_MAX_RETRIES + 1), so total iterations are capped at 11 regardless of these counters. 
They're not actually independent.

core/framework/llm/litellm.py
```diff
@@ -1053,17 +1059,12 @@ async def stream(
                             yield event
                         return
 
-                    # Empty stream — always retry regardless of last message
```
Antiarin, line 1843: These comments explain why the retry logic works the way it does. Don't remove them.

core/framework/llm/litellm.py
```diff
@@ -1072,23 +1073,19 @@ async def stream(
...
                             f"~{token_count} tokens ({token_method}). "
                             f"Request dumped to: {dump_path}. "
                             f"Retrying in {EMPTY_STREAM_RETRY_DELAY}s "
-                            f"(attempt {attempt + 1}/{EMPTY_STREAM_MAX_RETRIES})"
+                            f"(attempt {empty_stream_attempts}/{EMPTY_STREAM_MAX_RETRIES})"
                         )
                         await asyncio.sleep(EMPTY_STREAM_RETRY_DELAY)
                         continue
 
-                    # All retries exhausted — log and return the empty
```
Antiarin, line 1875: Don't remove these, they explain how EventLoopNode handles the empty result after retries are exhausted. Useful context for anyone reading this later.

Sri-Likhita-adru:
Hey @Antiarin, thanks for the feedback! Scoped this down to exactly what you suggested, added STREAM_TRANSIENT_MAX_RETRIES = 3 as a module-level constant and swapped it into the transient error handler (condition + log line). Dropped the independent counters entirely. Would you mind taking another look when you get a chance?

Antiarin (approved):
lgtm, Thank you for your contribution

Antiarin (approved):

PR #6217, opened 2026-03-11
Add command sanitizer module and enhance command validation

sundaram2021:
## Description

This PR hardens command execution against prompt/shell injection in the tools layer.

It adds a shared command sanitizer, validates commands before execution in both execution paths, and includes comprehensive tests for safe/blocked command behavior. It also includes a follow-up fix to normalize `.exe` handling safely (`endswith`-based) to satisfy lint and avoid misleading string stripping behavior.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #6213
Fixes #714

## Changes Made

- Added `command_sanitizer` with blocked executable/pattern checks to prevent dangerous shell commands.
- Enforced sanitizer validation in:
  - `tools/src/aden_tools/tools/file_system_toolkits/execute_command_tool/execute_command_tool.py`
  - `tools/coder_tools_server.py`
- Added/updated sanitizer-focused tests (`tools/tests/test_command_sanitizer.py`, `tools/tests/test_coder_tools_server.py`) covering safe, blocked, and edge cases.
- Fixed sanitizer executable normalization logic (`.exe` suffix handling) and resolved Ruff lint findings.

## Testing

Additional verification run:

- `cd tools && uv run pytest tests/test_command_sanitizer.py tests/test_coder_tools_server.py` (111 passed)
- `cd tools && uv run ruff check .` (passes)
- Manual sanitizer smoke check: safe command allowed, blocked command denied

## Checklist

### DIFF
--- a/tools/coder_tools_server.py
+++ b/tools/coder_tools_server.py
@@ -52,6 +52,60 @@ if "--stdio" in sys.argv:
 
 from fastmcp import FastMCP  # noqa: E402
 
+# Import command sanitizer — shared module in aden_tools
+try:
+    from aden_tools.tools.file_system_toolkits.command_sanitizer import (
+        CommandBlockedError,
+        validate_command,
+    )
+except ImportError:
+    # Fallback: inline minimal blocklist when aden_tools is not on sys.path.
+    # This keeps the coder-tools server self-contained for standalone usage.
+    import re as _re
+
+    class CommandBlockedError(Exception):  # type: ignore[no-redef]
+        pass
+
+    _BLOCKED_EXEC = {
+        "curl", "wget", "nc", "ncat", "netcat", "nmap", "ssh", "scp", "sftp",
+        "ftp", "telnet", "rsync", "invoke-webrequest", "iwr", "irm", "certutil",
+        "useradd", "userdel", "usermod", "adduser", "deluser", "passwd",
+        "net", "shutdown", "reboot", "halt", "poweroff", "mkfs", "fdisk",
+        "diskpart", "format", "bash", "sh", "zsh", "dash", "csh", "ksh",
+        "powershell", "pwsh", "cmd", "cmd.exe", "wscript", "cscript",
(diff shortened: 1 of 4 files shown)


sundaram2021:
@Hundao @bryanadenhq please look into this and let me know if there is any changes needed

Hundao (commented):
Hey @sundaram2021, nice sanitizer module. Few things I noticed:

1. `_extract_executable` doesn't strip directory prefixes, so `/usr/bin/curl` won't match `curl` in the blocklist. Apply `os.path.basename()` before matching.

2. The `python -c` regex requires a space (`\bpython[23]?\s+-c\b`), so `python -c'code'` without the space slips through.

3. The 62-line inline fallback in `coder_tools_server.py` duplicates the sanitizer. If one gets updated the other won't. Can you make the import reliable instead, or at least add a comment flagging the sync requirement?

4. `bash`/`sh` are blocked as executables but `subprocess.run(command, shell=True)` already runs everything through the shell. Worth noting this contradiction in the docstring so future readers understand the limitation.

sundaram2021:
Thanks for the detailed review. I’ve addressed all four points:

1. `_extract_executable` now normalizes to the executable basename before matching, so path-prefixed commands like `/usr/bin/curl` and `C:\Windows\System32\cmd.exe` are blocked correctly.
2. I tightened the `python -c` detection to cover quoted inline-code variants and added regression coverage for those cases.
3. I removed the duplicated inline sanitizer from `coder_tools_server.py`. It now bootstraps `tools/src` onto `sys.path` and imports the shared sanitizer directly, so there’s a single source of truth.
4. I updated the sanitizer / command runner docs to explicitly call out the `shell=True` limitation, so it’s clear that we block explicit nested shell executables but do not eliminate shell parsing entirely.

I also added tests for the basename-path case and the `python -c` edge cases.

Local checks passing:

1. `uv run ruff check .`
2.  `uv run pytest tests/test_command_sanitizer.py`

cc: @Hundao

PR #6233, opened 2026-03-12
fix(server): honor session_id in one-step session creation

### LINKED ISSUE #6232: [Bug]: POST /api/sessions ignores session_id when agent_path is provided
## Summary
`POST /api/sessions` documents `session_id` as an optional request field, but that value is currently ignored when `agent_path` is provided for one-step session + worker creation.

## Reproduction
1. Send `POST /api/sessions` with a body like:
   ```json
   {
     "session_id": "my-custom-session",
     "agent_path": "examples/templates/deep_research_agent"
   }
   ```
2. Inspect the response.

## Actual behavior
The server returns an auto-generated session ID instead of the requested `session_id`.

## Expected behavior
If `session_id` is accepted for queen-only session creation, it should also be honored for one-step session creation when `agent_path` is provided.

## Code pointers
- `core/framework/server/routes_sessions.py`
  - `handle_create_session()` reads `session_id`, but the `agent_path` branch does not pass it onward.
- `core/framework/server/session_manager.py`
  - `create_session_with_worker()` has no `session_id` parameter and always creates a generated session ID.

## Why this matters
This makes the API contract inconsistent depending on whether the worker is loaded at creation time, and it blocks callers that rely on deterministic session IDs.

Mactavish28:
## Summary
- honor `session_id` when `POST /api/sessions` is used with `agent_path` for one-step worker session creation
- thread the custom session ID through `create_session_with_worker(...)` so the API contract matches queen-only session creation
- add a regression test covering the forwarded `session_id` route behavior

Fixes #6232

## Test plan
- [x] Verified route forwarding with a direct `aiohttp` test-client script against the patched handler path
- [x] Ran `python3 -m py_compile core/framework/server/session_manager.py core/framework/server/routes_sessions.py core/framework/server/tests/test_api.py`
- [ ] `uv run pytest core/framework/server/tests/test_api.py -k create_session_with_worker_forwards_session_id` (not run in this environment because `uv` is unavailable locally)

### DIFF
--- a/core/framework/server/routes_sessions.py
+++ b/core/framework/server/routes_sessions.py
@@ -140,6 +140,7 @@ async def handle_create_session(request: web.Request) -> web.Response:
             session = await manager.create_session_with_worker(
                 agent_path,
                 agent_id=agent_id,
+                session_id=session_id,
                 model=model,
                 initial_prompt=initial_prompt,
                 queen_resume_from=queen_resume_from,
--- a/core/framework/server/session_manager.py
+++ b/core/framework/server/session_manager.py
@@ -147,6 +147,7 @@ class SessionManager:
         self,
         agent_path: str | Path,
         agent_id: str | None = None,
+        session_id: str | None = None,
         model: str | None = None,
         initial_prompt: str | None = None,
         queen_resume_from: str | None = None,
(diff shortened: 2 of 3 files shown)


Mactavish28:
I attempted to self-assign the linked issue, but GitHub is rejecting assignment for @Mactavish28 because the account is not currently assignable on this repository. A maintainer assignment on issue #6232 should clear this PR warning.

PR #6235, opened 2026-03-12
fix: escape browser_wait text matching

### LINKED ISSUE #6234: [Bug]: browser_wait(text=...) breaks on quotes and newlines
## Summary
The browser GCU tool `browser_wait(text=...)` currently interpolates raw text directly into a JavaScript string passed to Playwright. If the text contains a single quote, backslash, or newline, the generated script can become invalid or change meaning.

## Reproduction
Call `browser_wait` with a `text` value containing a quote or newline, for example:
- `O'Reilly`
- `line1\nline2`

## Actual behavior
The tool builds JavaScript like:

```python
await page.wait_for_function(
    f"document.body.innerText.includes('{text}')",
    timeout=timeout_ms,
)
```

This can fail before the wait condition is even evaluated.

## Expected behavior
`browser_wait(text=...)` should treat the text as data, not source code, so quoted or multiline values work the same as plain strings.

## Code pointer
- `tools/src/gcu/browser/tools/advanced.py`

## Suggested fix
Pass `text` as an argument to `page.wait_for_function(...)` instead of interpolating it into the JavaScript source string.

Mactavish28:
Closes #6234.

## Summary
- pass `browser_wait(text=...)` values to Playwright as a function argument instead of interpolating them into JavaScript source
- preserve correct waiting behavior for quoted and multiline text values
- add a focused regression test for the quoted/multiline case

## Test plan
- `python3 -m py_compile tools/src/gcu/browser/tools/advanced.py tools/tests/test_browser_advanced_tools.py`
- `pytest` could not be run locally because this environment does not have `pytest` or `fastmcp` installed

### DIFF
--- a/tools/src/gcu/browser/tools/advanced.py
+++ b/tools/src/gcu/browser/tools/advanced.py
@@ -56,7 +56,8 @@ def register_advanced_tools(mcp: FastMCP) -> None:
                 return {"ok": True, "action": "wait", "condition": "selector", "selector": selector}
             elif text:
                 await page.wait_for_function(
-                    f"document.body.innerText.includes('{text}')",
+                    "(text) => document.body.innerText.includes(text)",
+                    arg=text,
                     timeout=timeout_ms,
                 )
                 return {"ok": True, "action": "wait", "condition": "text", "text": text}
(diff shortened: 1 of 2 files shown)


Mactavish28:
I linked this PR to issue #6234, but GitHub does not currently allow `@Mactavish28` to self-assign on this repository, so the contribution-requirements warning may still appear. If a maintainer can assign me to the linked issue, that should clear the requirement.

PR #6241, opened 2026-03-12
fix: preserve custom session ids in runtime logs

### LINKED ISSUE #6240: [Bug]: custom session IDs disappear from worker-session logs and history
## Summary
Custom worker session IDs are not treated as first-class session-backed runs across the logging and browsing paths.

When a run uses a custom session ID instead of a generated `session_*` name, multiple code paths still assume only `session_*` IDs are valid. That causes the run to disappear from worker-session browsing and prevents runtime logs from being written/read in the unified session location.

## Impact
- worker session directories with custom IDs can be skipped by the session browser
- runtime logs can fall back to the deprecated `runs/` path instead of `sessions/<id>/logs/`
- log/history views no longer line up with the actual execution session ID

## Root cause
There are three places using the `session_` prefix as a proxy for "real session-backed run":

1. `core/framework/graph/executor.py`
   - only forwards a session ID to `RuntimeLogger.start_run(...)` when `storage_path.name.startswith("session_")`
2. `core/framework/runtime/runtime_log_store.py`
   - only treats `session_*` directories as the new unified session log layout
3. `core/framework/server/routes_sessions.py`
   - worker-session listing filters directories by the same prefix

## Expected behavior
Any valid worker session directory under `sessions/<id>/` should be treated as a session-backed run, regardless of whether the ID starts with `session_`.

## Suggested fix
- derive the runtime-log session ID from the storage path being under `sessions/<id>/`, not from the directory name prefix
- make `RuntimeLogStore` support session-backed logs for arbitrary session IDs
- make worker-session listing include custom session IDs as long as the directory is a valid session directory

Mactavish28:
Closes #6240.

## Summary
- treat any execution stored under `sessions/<id>` as a session-backed run for runtime logging, even when the ID does not start with `session_`
- write and discover unified runtime logs for custom session IDs instead of falling back to deprecated `runs/` storage
- include custom worker session IDs in session and log API regression coverage

## Test plan
- `python3 -m py_compile core/framework/runtime/runtime_log_store.py core/framework/runtime/runtime_logger.py core/framework/graph/executor.py core/framework/server/routes_sessions.py core/framework/server/tests/test_api.py core/framework/runtime/tests/test_runtime_logging_paths.py`
- `pytest` could not be run locally because this environment does not have the test dependencies installed

### DIFF
--- a/core/framework/server/routes_sessions.py
+++ b/core/framework/server/routes_sessions.py
@@ -410,10 +410,12 @@ async def handle_list_worker_sessions(request: web.Request) -> web.Response:
     for d in sorted(sess_dir.iterdir(), reverse=True):
         if not d.is_dir():
             continue
+        state_path = d / "state.json"
+        if not d.name.startswith("session_") and not state_path.exists():
+            continue
 
         entry: dict = {"session_id": d.name}
 
-        state_path = d / "state.json"
         if state_path.exists():
             try:
                 state = json.loads(state_path.read_text(encoding="utf-8"))
--- a/core/framework/server/routes_sessions.py
+++ b/core/framework/server/routes_sessions.py
@@ -408,7 +408,7 @@ async def handle_list_worker_sessions(request: web.Request) -> web.Response:
 
     sessions = []
     for d in sorted(sess_dir.iterdir(), reverse=True):
-        if not d.is_dir() or not d.name.startswith("session_"):
+        if not d.is_dir():
             continue
 
         entry: dict = {"session_id": d.name}
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -403,6 +403,14 @@ class GraphExecutor:
         )
         return s1 + "\n\n" + s2
 
+    def _get_runtime_log_session_id(self) -> str:
+        """Return the session-backed execution ID for runtime logging, if any."""
+        if not self._storage_path:
+            return ""
+        if self._storage_path.parent.name != "sessions":
+            return ""
+        return self._storage_path.name
+
     async def execute(
         self,
         graph: GraphSpec,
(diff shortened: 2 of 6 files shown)


Mactavish28:
This PR is linked to issue #6240, but GitHub still does not allow `@Mactavish28` to self-assign in this repository. If a maintainer can assign me to the linked issue, that should clear any contribution-requirements warning.

core/framework/server/routes_sessions.py
```diff
@@ -408,7 +408,7 @@ async def handle_list_worker_sessions(request: web.Request) -> web.Response:
 
     sessions = []
     for d in sorted(sess_dir.iterdir(), reverse=True):
-        if not d.is_dir() or not d.name.startswith("session_"):
+        if not d.is_dir():
```
Hundao, line 411: Removing the filter entirely means any stray directory under `sessions/` shows up as a ghost entry in the UI. Can you keep the `session_` prefix as a fast path and fall back to checking `state.json` for custom IDs?

```python
if not d.is_dir():
    continue
if not d.name.startswith("session_") and not (d / "state.json").exists():
    continue
```

Hundao (changes requested):

core/framework/server/routes_sessions.py
Mactavish28: Applied in ca3aa305. I kept the session_ prefix as the fast path and only fall back to custom IDs when a backing state.json exists, so stray directories under sessions/ will not appear as ghost entries in the UI.

I also re-ran the targeted coverage after the change from the core workspace:
uv run pytest framework/server/tests/test_api.py -k 'test_list_sessions or test_list_sessions_includes_custom_id or test_logs_list_summaries or test_logs_list_summaries_with_custom_id or test_logs_session_summary' framework/runtime/tests/test_runtime_logging_paths.py

Result: 8 passed.

PR #6298, opened 2026-03-13
fix: surface rate limit error in chat UI

ASHUTOSH-KUMAR-RAO:
## Description

When an LLM hits a rate limit (429 error), the agent was silently 
getting stuck with no feedback in the chat UI. This fix surfaces 
the error as a visible message in the chat.

## Type of Change

## Changes Made

- Added `emit_execution_failed` in `event_loop_node.py` for 
  client-facing nodes when 429/rate limit error is detected
- Added `emit_execution_failed` for non-client-facing nodes as well
- Error now surfaces in chat UI as "Model rate limit reached"

## Testing

- Tested with RATE_LIMIT_MAX_RETRIES = 1 to trigger rate limit quickly
- Verified error pill appears in chat UI
- Verified agent does not silently hang

Fixes #6216

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for model rate limit situations with clearer messaging to users
  * Improved retry mechanism for rate-limited requests to increase operational reliability
  * Fixed error message consistency and delivery across different execution scenarios
  * Better user experience and feedback when rate limits are encountered


RichardTang-Aden:
A couple things to fix before merging, the first one is critical:

- The bare raise on the non-client-facing path makes all the existing logging/metrics below it unreachable. That needs to stay.
- Instead of string-matching "429" / "rate limit", you can use isinstance(e, RateLimitError) since _is_transient_error already imports it.
- emit_execution_failed in the client-facing path conflicts with the recovery flow (user can still retry) — probably drop that.

ASHUTOSH-KUMAR-RAO:
Thanks for the review! 🙏  
I've implemented all three requested changes:

1. Removed the bare `raise` so logging/metrics are no longer unreachable  
2. Replaced string-based detection with `isinstance(e, RateLimitError)`  
3. Removed the `emit_execution_failed` call in the client-facing path to keep the recovery flow intact  

Also added a safe optional import fallback for `RateLimitError` for compatibility with older litellm versions.

And apologies — I initially missed the bare `raise` issue.  
Thanks for pointing it out! It's fixed now. 👍

PR is ready for re-review.

bryanadenhq:
Hey, thanks for working on this! The intent is solid but there are a couple bugs that need fixing before we can merge:

Duplicate error_msg assignment: the old error_msg = f"LLM call failed: {e}" line wasn't removed, so it always overwrites your new rate-limit message. The if isinstance(e, RateLimitError) branch will never actually surface "Model rate limit reached."

Missing _llm_turn_failed_waiting_input = True in the client-facing path: the PR moved this flag to the non-client-facing block only, so after the break the outer loop will fall through to token processing instead of skipping it.

For the non-client-facing rate-limit path: you're emitting execution_failed but then continuing the loop instead of raising. This silently swallows the error, was that intentional? On main, non-client-facing errors always crash.

ASHUTOSH-KUMAR-RAO:
> Hey, thanks for working on this! The intent is solid but there are a couple bugs that need fixing before we can merge:
> 
> Duplicate error_msg assignment: the old error_msg = f"LLM call failed: {e}" line wasn't removed, so it always overwrites your new rate-limit message. The if isinstance(e, RateLimitError) branch will never actually surface "Model rate limit reached."
> 
> Missing _llm_turn_failed_waiting_input = True in the client-facing path: the PR moved this flag to the non-client-facing block only, so after the break the outer loop will fall through to token processing instead of skipping it.
> 
> For the non-client-facing rate-limit path: you're emitting execution_failed but then continuing the loop instead of raising. This silently swallows the error, was that intentional? On main, non-client-facing errors always crash.

Hi Bryan! Thanks for the detailed review. I've addressed all 3 points:
1. Fixed the duplicate `error_msg` assignment — wrapped it in `if not isinstance(e, RateLimitError)`
2. Added `_llm_turn_failed_waiting_input = True` in the client-facing path
3. Added `raise` in the non-client-facing rate limit path after `emit_execution_failed`

One quick question — VS Code is flagging `_llm_turn_failed_waiting_input = True` as unreachable code after `await self._await_user_input()`. Should this flag be set before or after the await call? Want to make sure the logic is correct before pushing!

PR #6308, opened 2026-03-13
fix: handle clipboard write errors in SystemPromptTab

### LINKED ISSUE #6306: [Bug]: Clipboard write in SystemPromptTab lacks error handling and may show false "Copied" feedback
## Describe the Bug

In `core/frontend/src/components/NodeDetailPanel.tsx`, the `SystemPromptTab` component allows users to copy the system prompt using the Clipboard API.

The current implementation calls `navigator.clipboard.writeText(prompt)` without error handling:

navigator.clipboard.writeText(prompt);
setCopied(true);
setTimeout(() => setCopied(false), 1500);

If the clipboard write fails (e.g., due to insecure context, browser permission restrictions, or unsupported environments), the failure is silent and the UI still shows the **"Copied"** state.

This can mislead users into believing the text was successfully copied even when the clipboard operation failed.

---

## To Reproduce

1. Open a node in the **Node Detail Panel**
2. Navigate to the **System Prompt** tab
3. Click the **Copy** button
4. If clipboard access fails (for example due to browser restrictions or insecure context), the UI still shows **"Copied"**

---

## Expected Behavior

The **"Copied"** state should only appear if the clipboard write succeeds.
If the clipboard operation fails, the error should be handled gracefully and the UI should not indicate a successful copy.

---

## Screenshots

N/A

---

## Environment

* OS: Any
* Browser: Chrome / Firefox / Edge
* Frontend: React + TypeScript
* Python version: N/A (frontend UI behavior)
* Docker version: N/A

---

## Configuration

N/A

---

## Logs

No logs are produced because clipboard failures are currently not handled.

---

## Additional Context

A small improvement would be to wrap the clipboard operation in a `try/catch` block and optionally guard against environments where `navigator.clipboard` is unavailable.

Example approach:

```javascript
const handleCopy = async () => {
  try {
    if (!navigator.clipboard) {
      throw new Error("Clipboard API not supported");
    }

    await navigator.clipboard.writeText(prompt);
    setCopied(true);
    setTimeout(() => setCopied(false), 1500);
  } catch (err) {
    console.error("Clipboard copy failed:", err);
  }
};
```

This would prevent silent failures and ensure the UI reflects the actual result of the copy operation.

If this aligns with the intended behavior, I'd be happy to work on implementing the fix.

akdevv:
## Description
Add error handling to the clipboard write operation in `SystemPromptTab` 
to prevent false "Copied" feedback when the clipboard operation fails.

## Motivation
The current implementation calls `navigator.clipboard.writeText()` without 
any error handling. If the clipboard write fails (e.g., insecure context, 
browser permission restrictions, or unsupported environments), the failure 
is silent and the UI still shows the "Copied" state, misleading users into 
thinking the copy succeeded.

## Changes
- Wrapped `navigator.clipboard.writeText()` in a `try/catch` block in `NodeDetailPanel.tsx`
- Made `handleCopy` function async
- Moved `setCopied(true)` inside the success path so it only fires on successful copy
- Added `console.error` logging for clipboard failures
- Added guard for environments where `navigator.clipboard` is unavailable

## Testing
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [x] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Checklist
- [x] Code follows style guidelines (ruff)
- [x] Self-review completed
- [ ] Documentation updated
- [x] No breaking changes (or documented if unavoidable)

Closes #6306

### DIFF
--- a/core/frontend/src/components/NodeDetailPanel.tsx
+++ b/core/frontend/src/components/NodeDetailPanel.tsx
@@ -170,10 +170,17 @@ function SystemPromptTab({ systemPrompt }: { systemPrompt?: string }) {
   const prompt = systemPrompt || "";
   const [copied, setCopied] = useState(false);
 
-  const handleCopy = () => {
-    navigator.clipboard.writeText(prompt);
-    setCopied(true);
-    setTimeout(() => setCopied(false), 1500);
+  const handleCopy = async () => {
+    try {
+      if (!navigator.clipboard) {
+        throw new Error("Clipboard API not supported");
+      }
+      await navigator.clipboard.writeText(prompt);
+      setCopied(true);
+      setTimeout(() => setCopied(false), 1500);
+    } catch (err) {
+      console.error("Clipboard copy failed:", err);
+    }
   };
 
   if (!prompt) {


core/frontend/src/components/NodeDetailPanel.tsx
```diff
@@ -170,10 +170,17 @@ function SystemPromptTab({ systemPrompt }: { systemPrompt?: string }) {
...
+    try {
+      if (!navigator.clipboard) {
+        throw new Error("Clipboard API not supported");
+      }
+      await navigator.clipboard.writeText(prompt);
+      setCopied(true);
+      setTimeout(() => setCopied(false), 1500);
+    } catch (err) {
+      console.error("Clipboard copy failed:", err);
+    }
```
jayam04, line 183: @akdevv It doesn't handle the case when `navigator.clipboard` is present but still fails due to other reasons.

I think moving copy functions to `.then()` will do the work. Additionally, can add some warning using `.catch()` block.

PR #6452, opened 2026-03-15
docs(developer-guide): fix agent.json examples to match current schema

Jose37456:
## Description
The manual \`agent.json\` example in \`docs/developer-guide.md\` used stale field names (\`node_id\` instead of \`id\`, flat top-level \`nodes\`/\`edges\` instead of nested under \`graph\`). Nodes were silently dropped on load because the loader expected \`NodeSpec.id\`. The \`AgentRunner.from_file()\` call also doesn't exist — it's \`AgentRunner.load()\`.

## Type of Change
- [x] Documentation fix / Bug

## Related Issues
Fixes #897

## Changes Made
- \`docs/developer-guide.md\` — updated agent.json examples to match current \`NodeSpec\`/\`GraphSpec\` schema (\`id\`, nested \`graph.nodes\`/\`graph.edges\`, correct \`success_criteria\` placement)
- Fixed two more inline snippets using \`node_id\` → \`id\`

## Checklist
- [x] Self-review done
- [x] Verified against \`examples/templates/deep_research_agent/agent.json\`

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -283,33 +283,48 @@ Use the coder-tools MCP tools from your IDE agent chat (e.g., initialize_and_bui
 
 If you prefer to build agents manually:
 
-```python
-# exports/my_agent/agent.json
+```json
 {
-  "goal": {
-    "goal_id": "support_ticket",
-    "name": "Support Ticket Handler",
-    "description": "Process customer support tickets",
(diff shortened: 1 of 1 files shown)


Hundao:
Hey @Jose37456, we used your approach and landed the fix in #6878. You're credited as co-author on the commit. Thanks for the contribution\!

Issue #5706, opened 2026-03-03
[Bug]: ParallelExecutionConfig.branch_timeout_seconds and memory_conflict_strategy are never enforced


gavksingh:
## Describe the Bug

`ParallelExecutionConfig` in `core/framework/graph/executor.py:87-99` declares three config fields for parallel fan-out behavior. Two of them `branch_timeout_seconds` and `memory_conflict_strategy` are never read anywhere in the codebase. Setting them has no effect.

The third field, `on_branch_failure`, works correctly (enforced at `executor.py:2114-2121`).

## `branch_timeout_seconds` (line 99)  never enforced

```python
branch_timeout_seconds: float = 300.0
```

The only reference in the entire codebase is the definition itself. The parallel execution path at `executor.py:2091-2093` runs branches with a bare `asyncio.gather()`:

```python
tasks = [execute_single_branch(b) for b in branches.values()]
results = await asyncio.gather(*tasks, return_exceptions=False)
```

No `asyncio.wait_for()` wraps individual branch tasks. If an LLM provider hangs (rate limit backoff, network partition, provider outage), a single stuck branch blocks the entire graph execution indefinitely. No timeout fires, no event is emitted, no fallback triggers — the `on_branch_failure` strategies never get a chance to run.

## `memory_conflict_strategy` (line 96) - never enforced

```python
memory_conflict_strategy: str = "last_wins"  # "last_wins", "first_wins", "error"
```

The comment documents three strategies, but the field is never read. Branch result merging at `executor.py:2042-2045` unconditionally calls `memory.write_async(key, value)` for every output key:

```python
if result.success:
    for key, value in result.output.items():
        await memory.write_async(key, value)
```

The `"first_wins"` and `"error"` strategies don't exist. When parallel branches write the same output key, last-writer-wins silently — no detection, no logging, no conflict signal.

## To Reproduce

```bash
grep -rn "branch_timeout" core/ --include="*.py"
# Only hit: core/framework/graph/executor.py:99 (the definition)

grep -rn "memory_conflict" core/ --include="*.py"
# Only hit: core/framework/graph/executor.py:96 (the definition)
```

## Expected Behavior

- Setting `branch_timeout_seconds=30` should actually time out branches after 30 seconds
- Setting `memory_conflict_strategy="error"` should actually raise on conflicting writes

## Where the fix goes

[...]

PR #6499, opened 2026-03-15
feat(events): emit GOAL_ACHIEVED on terminal-node success

### LINKED ISSUE #5736: [Feature]: emit GOAL_ACHIEVED event on terminal-node success
## Problem Statement

When an agent graph reaches a terminal node successfully, there is no programmatic signal distinguishing "the agent achieved its goal" from "it simply ran out of nodes." The `EventBus` defines `GOAL_ACHIEVED` in the `EventType` enum (line 75 of `event_bus.py`) and has convenience emit methods for related lifecycle events like `GOAL_PROGRESS` (line 424) and `CONSTRAINT_VIOLATION` (line 442), but `GOAL_ACHIEVED` has no corresponding emit method and is never emitted anywhere in the execution path.

The TUI layer already subscribes to and handles this event (`app.py:1157`, `chat_repl.py:1656-1658`, `log_pane.py:32`), and `EVENT_TYPES.md` notes it as "reserved for explicit goal completion signals." The consumer side is ready, but the backend never fires the event.

## Proposed Solution

- Add an `emit_goal_achieved()` convenience method to `EventBus` following the existing `emit_goal_progress()` pattern
- Emit `GOAL_ACHIEVED` from the executor's terminal-node success path with goal description and final output as evidence in the payload
- This completes the agent lifecycle observability story: developers and the TUI can now react to goal completion, not just progress

## Alternatives Considered

- **Infer goal completion from EXECUTION_COMPLETED:** This event fires on any execution end (success or failure), so it cannot distinguish goal achievement from exhaustion or error
- **Let developers emit it manually in custom nodes:** Possible, but the framework should emit it for the standard terminal-node-reached case since it already tracks this internally

## Additional Context

- `EVENT_TYPES.md` (lines 385-388) explicitly states this event is "reserved for explicit goal completion signals"
- The TUI rendering is fully implemented and waiting: green star icon, bold green text, log pane formatting
- `emit_goal_progress()` already exists and is fully wired, so the pattern and infrastructure are established

## Implementation Ideas

- New `emit_goal_achieved()` method in `event_bus.py` mirroring `emit_goal_progress()` signature
- Emission at the terminal-node success path in `executor.py` (around the "Reached terminal node" log line)
- Optional integration with `OutcomeAggregator` for threshold-based goal completion

gavksingh:
## Summary

- Add `emit_goal_achieved()` convenience method to `EventBus`, mirroring the pattern of existing emitters like `emit_constraint_violation()`
- Emit `GOAL_ACHIEVED` from `GraphExecutor` when execution completes at a terminal node successfully
- Update `EVENT_TYPES.md` to document the payload shape and emission point
- Add unit + integration tests for the new emission

The `GOAL_ACHIEVED` `EventType` was already defined in the enum and fully handled by the TUI (subscription, `chat_repl` handler, `log_pane` formatting), but no backend code ever emitted it. This completes the agent lifecycle observability story.

Fixes #5736

## Test plan

- [x] `emit_goal_achieved()` publishes an `AgentEvent` with correct type, stream_id, execution_id, and data
- [x] Default output is `{}` when None is passed
- [x] `GraphExecutor` calls `emit_goal_achieved` on successful terminal-node completion
- [x] `GraphExecutor` works without crash when `event_bus` is None
- [x] Existing `test_executor_skips_events_for_event_loop_nodes` updated to distinguish node-level vs graph-level events
- [x] All 774 existing tests pass (no regressions)
- [x] `ruff check` passes on all modified files

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -1574,6 +1574,15 @@ class GraphExecutor:
                     execution_quality=exec_quality,
                 )
 
+            if self._event_bus:
+                await self._event_bus.emit_goal_achieved(
+                    stream_id=self._stream_id,
+                    execution_id=self._execution_id,
+                    goal_name=goal.name,
+                    goal_description=goal.description,
+                    output=output,
+                )
+
             return ExecutionResult(
                 success=True,
                 output=output,
--- a/core/framework/runtime/EVENT_TYPES.md
+++ b/core/framework/runtime/EVENT_TYPES.md
@@ -392,7 +392,15 @@ Goal completion progress update.
 
 ### `goal_achieved`
 
-Not currently emitted — reserved for explicit goal completion signals.
+Emitted when a graph execution reaches a terminal node successfully, signalling that the goal has been achieved.
+
+| Data Field         | Type   | Description                              |
+| ------------------ | ------ | ---------------------------------------- |
+| `goal_name`        | `str`  | Name of the achieved goal                |
+| `goal_description` | `str`  | Human-readable description of the goal   |
+| `output`           | `dict` | Final output from the execution          |
+
+**Emitted by:** `GraphExecutor` at terminal-node success, via `emit_goal_achieved()`.
 
 ---
 
--- a/core/framework/runtime/event_bus.py
+++ b/core/framework/runtime/event_bus.py
@@ -619,6 +619,28 @@ class EventBus:
             )
         )
 
+    async def emit_goal_achieved(
+        self,
+        stream_id: str,
+        execution_id: str,
+        goal_name: str,
+        goal_description: str,
+        output: dict[str, Any] | None = None,
+    ) -> None:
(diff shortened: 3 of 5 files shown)


bryanadenhq (changes requested):
Thanks for wiring up the GOAL_ACHIEVED event, that part looks clean and straightforward. However, this PR bundles in significant fan-out executor changes (memory conflict detection, per-branch timeouts, refactored result processing) that aren't related to the goal achieved event or #5736. Could you split those into a separate PR? It'll make both easier to review and safer to merge/revert independently.

gavksingh:
@bryanadenhq thanks for pointing that out! The executor changes (branch timeout + memory conflict) were from #5706 which I also worked on and got merged separately via PR #6504. Forgot to rebase this branch after that. Rebased and force pushed now so this PR only has the GOAL_ACHIEVED event emission.

Issue #2805, opened 2026-01-30
[Integration]: Expand agent capabilities by adding more integrations/tools


RichardTang-Aden:
## Problem Statement

Currently, Hive agents have limited integrations with external business tools and services. This restricts their ability to automate complex workflows and forces users to rely on manual processes or multiple disconnected systems. For example,  users cannot automatically sync lead data from their CRM to their ERP system and notify the team via Slack—each step requires manual intervention or custom scripting.

## Proposed Solution

Expand Hive's integration ecosystem to support a broader range of business services and APIs. This will enable agents to:

- Connect directly with popular platforms (Salesforce, Slack, HubSpot, QuickBooks, etc.)
- Interact with custom APIs and internal systems
- Orchestrate complex, multi-step workflows without manual intervention
- Provide a consistent authentication and configuration experience across all integrations

## How to Contribute

We welcome community contributions for new integrations! Each integration should be proposed as a separate issue.

### Creating an Integration Proposal

**Issue title format**: `[Integration]: <Service or Tool Name>`

**Required information in your issue**:
- **Reason**: Why this integration is needed
- **Use case**: Specific business processes it will enable or improve
- **Scope**: What endpoints/features will be included
- **Link**: Reference this parent issue in your proposal

### Example Issues
- [Integration]: Notion - https://github.com/adenhq/hive/issues/7

### Tips for Faster Review and Assignment
- Link your issue to this parent feature issue
- Provide concrete use cases with real-world examples
- Specify the most critical endpoints/features for MVP

## How to Build an Integration

A complete integration consists of four key components:

### 1. **Functions**
The integration endpoints and core logic that interact with the external service.

### 2. **Credentials**
Authentication mechanism for users to securely connect their accounts. Must follow the `credentialSpec` and `credentialStore` patterns to ensure compatibility with Hive's node system.

### 3. **Documentation**
Clear documentation covering:
- Scope and coverage of the integration
- Supported endpoints and features
- Setup and authentication instructions
- Usage examples

### 4. **Tests**
Unit tests to verify the integration functions correctly.

### Example Implementation
See this example PR for reference: https://github.com/adenhq/hive/pull/2601

## Getting Started

1. Check existing issues to avoid duplicates
2. Create a new issue using the format above
3. Wait for maintainer feedback and assignment
4. Follow the implementation guidelines
5. Submit a PR referencing your issue

---

**Questions?** Feel free to comment on this issue or reach out to the maintainers.

[...]

PR #6527, opened 2026-03-16
feat(tools): add IPInfo tool — IP geolocation and network intelligence

nikhilvarmakandula:
Closes #6508

## Summary
Adds IPInfo integration using ipinfo.io — free tier, 50k requests/month.

## Tools Added
- `ipinfo_get_ip_details(ip)` — location, org, timezone for any IP
- `ipinfo_get_my_ip()` — details for current machine's IP

## Credential
- IPINFO_TOKEN (free signup at ipinfo.io)

## Testing
- ipinfo_get_ip_details("8.8.8.8") → Google DNS location ✅
- ipinfo_get_my_ip() → current machine location ✅

## Notes
- Follows existing credential pattern (brevo_tool)
- Parent issue: #2805

### DIFF
--- a/tools/src/aden_tools/credentials/__init__.py
+++ b/tools/src/aden_tools/credentials/__init__.py
@@ -67,6 +67,7 @@ from .azure_sql import AZURE_SQL_CREDENTIALS
 from .base import CredentialError, CredentialSpec
 from .bigquery import BIGQUERY_CREDENTIALS
 from .brevo import BREVO_CREDENTIALS
+from .ipinfo import IPINFO_CREDENTIALS
 from .browser import get_aden_auth_url, get_aden_setup_url, open_browser
 from .calcom import CALCOM_CREDENTIALS
 from .calendly import CALENDLY_CREDENTIALS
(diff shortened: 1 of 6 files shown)


Hundao (changes requested):
Hey, a few things to fix before this can move forward:

1. **Credential guard missing.** The tool needs to check credentials before making API calls and return `{error, help}` when the token is missing. Look at how other tools like `wandb_tool.py` or `brevo_tool.py` use `_get_creds()` with `CredentialStoreAdapter`. Right now `ipinfo_get_my_ip()` succeeds without a token since ipinfo's free tier doesn't require one, which breaks the credential test framework.

2. **Health check needs a checker function.** You set `health_check_endpoint` in the credential spec but there's no corresponding checker registered. See `test_specs_with_endpoint_have_checkers` for what's expected.

3. **Missing unit tests.** Other tools have a `tools/tests/tools/test_*_tool.py` file.

4. **Import ordering.** `ipinfo` import in `credentials/__init__.py` breaks alphabetical order (should be after `intercom`, not between `brevo` and `browser`).

Also please rebase on latest main when you push the fixes. Thanks!

Issue #6365, opened 2026-03-14
[Skill System]: Apply default skill config overrides and runtime heuristics


TimothyZhang7:
## Why This Matters

Agent builders who set `checkpoint_every_n: 10` or `warn_at_usage_ratio: 0.4` in their config expect that value to take effect — today it is silently ignored. The config is parsed, logged, and discarded. This breaks the trust contract between the framework and the developer. Runtime heuristics like batch auto-detection and token usage warnings are what make default skills genuinely useful rather than just boilerplate instructions the agent may or may not follow.

## Context

Implements PRD requirements **DS-12**, **DS-13** from [`docs/skill-registry-prd.md`](https://github.com/aden-hive/hive/blob/main/docs/skill-registry-prd.md) §5.3–5.4, and addresses config override application.

Parent issue: #6330

## Current State

- `DefaultSkillConfig` parses per-skill overrides (e.g., `checkpoint_every_n: 10`, `warn_at_usage_ratio: 0.4`)
- `SkillsConfig.get_default_overrides(skill_name)` returns the override dict
- But **overrides are never applied** — they're parsed and logged but don't affect behavior

### Config Override Application

Default skill SKILL.md bodies contain placeholder values (e.g., "every 5 iterations"). When overrides are configured, the injected protocol text should reflect the overridden values. Options:

1. **Template substitution**: Replace placeholder values in SKILL.md body before injection
2. **Config appendix**: Append a `## Configuration` section to each skill's protocol with active overrides
3. **Parametric prompts**: Use `{{variable}}` syntax in SKILL.md and substitute at load time

### DS-12: Batch Auto-Detection Heuristic (P1)

`hive.batch-ledger` should auto-detect batch scenarios:

- Heuristic: if the goal/input mentions processing a list, collection, or set of items
- When detected, proactively initialize the batch ledger
- Configurable via `auto_detect_batch` (default true)

### DS-13: Context Preservation Token Usage Warning (P0)

`hive.context-preservation` should warn at 45% token usage (before the framework's 60% prune threshold):

- Monitor token usage during execution
- Inject a preservation nudge when usage crosses `warn_at_usage_ratio` (default 0.45)
- Configurable via override

## Acceptance Criteria

- [ ] Config overrides from `default_skills` agent config are applied to default skill prompts
- [ ] Batch auto-detection heuristic identifies batch scenarios from goal text
- [ ] Context preservation warning fires at configurable token usage threshold
- [ ] Override values are logged at startup
[...]

PR #6610, opened 2026-03-18
feat(skills): DS-12 and DS-13 — config override application, batch auto-detection, and context preservation warning

levxn:
## Summary

Implements DS-12, DS-13, and config override application for default skills. Config overrides set in agent definitions (e.g. `checkpoint_every_n: 10`, `warn_at_usage_ratio: 0.4`) were previously parsed and logged but silently discarded — they now take effect at runtime.

- **Config overrides**: `{{placeholder}}` substitution added to SKILL.md bodies; overrides from `default_skills` agent config replace default values before the protocol is injected into the system prompt
- **DS-12**: `hive.batch-ledger` auto-detects batch scenarios from goal/input text at node startup and prepends a ledger-init nudge to the system prompt; configurable via `auto_detect_batch` override
- **DS-13**: `hive.context-preservation` monitors token usage during execution and injects a one-time preservation warning when usage crosses `warn_at_usage_ratio` (default 0.45), before the framework's 0.6 prune threshold fires

---

## What Changed

**`core/framework/skills/_default_skills/quality-monitor/SKILL.md`**
- Replaced hardcoded `5` with `{{assessment_interval}}` placeholder

**`core/framework/skills/_default_skills/error-recovery/SKILL.md`**
- Replaced hardcoded `3` with `{{max_retries_per_tool}}` placeholder

**`core/framework/skills/_default_skills/context-preservation/SKILL.md`**
- Added a line: `You will receive an alert when context reaches {{warn_at_usage_ratio_pct}}% — preserve immediately.`

**`core/framework/skills/defaults.py`**
- Added `_SKILL_DEFAULTS` dict with per-skill default values for all placeholders
- Added `_apply_overrides()` — merges defaults with agent-supplied overrides and substitutes `{{key}}` in skill body text; handles `warn_at_usage_ratio` float to percent conversion
- `build_protocols_prompt()` now calls `_apply_overrides()` per skill before appending to the combined prompt
- Added `is_batch_scenario(text)` — keyword heuristic for DS-12 batch detection (11 keywords: "list of", "for each", "records", "items", etc.)
- Added `batch_init_nudge` property — returns nudge text when `hive.batch-ledger` is active and `auto_detect_batch` is not False
- Added `context_warn_ratio` property — returns `warn_at_usage_ratio` override or 0.45 default; returns None when `hive.context-preservation` is disabled

**`core/framework/skills/manager.py`**
- `SkillsManager` stores a reference to `DefaultSkillManager` after `_do_load()`
- Exposes `batch_init_nudge` and `context_warn_ratio` as passthrough properties

**`core/framework/runtime/agent_runtime.py`**
- Reads `context_warn_ratio` and `batch_init_nudge` from `SkillsManager` after load; stores as instance attributes and passes to `ExecutionStream`

**`core/framework/runtime/execution_stream.py`**
- Accepts and threads `context_warn_ratio` and `batch_init_nudge` through to `GraphExecutor`

**`core/framework/graph/executor.py`**
- Accepts and threads both new params through to `NodeContext` creation

**`core/framework/graph/node.py`**
- Added `default_skill_batch_nudge: str | None` and `default_skill_warn_ratio: float | None` to `NodeContext`

**`core/framework/graph/event_loop_node.py`**
- DS-12: after protocols are injected into the system prompt, checks `ctx.default_skill_batch_nudge`; runs `is_batch_scenario()` against `goal_context + input_data`; appends nudge if batch is detected
- DS-13: initializes `_context_warn_sent = False` per execution; after each successful LLM turn, checks `usage_ratio() >= ctx.default_skill_warn_ratio`; injects a user-facing `CONTEXT ALERT` message once and sets the flag

**`core/tests/test_default_skills.py`**
- 19 new tests across three new classes: `TestConfigOverrideSubstitution`, `TestBatchAutoDetection`, `TestContextWarnRatio`

---

## Why

Agents configured with `assessment_interval: 10` or `warn_at_usage_ratio: 0.4` expect those values to change agent behavior — previously they were no-ops. Batch auto-detection removes the manual overhead of telling the agent to initialize a ledger every time. Context preservation warnings give the agent advance notice at 45% usage rather than waiting until the framework silently prunes at 60%.

---

## How to Test

Run the updated test file:

```
uv run pytest core/tests/test_default_skills.py -v
```

All 44 tests pass. Full suite regression:

```
uv run pytest core/tests/ -q
```

1002 passed, 0 failures.

---

## Out of Scope

`checkpoint_every_n` is parsed and stored as an override but has no runtime hook yet — it is included in `_SKILL_DEFAULTS` so it will substitute cleanly into any future SKILL.md placeholder that references it. Iteration and node-completion hook infrastructure (DS-9, DS-10) are tracked separately.

---

solves issue #6365


levxn:
@TimothyZhang7 let me know if it works!

Issue #6632, opened 2026-03-19
feat: Add Antigravity subscription support


RichardTang-Aden:
## Description

Add Antigravity as a subscription-based LLM provider, following the same pattern as Claude Code and OpenAI Codex. Users with an existing Antigravity subscription should be able to use it without purchasing a separate API key.

## Requirements

1. **Quickstart & Worker Setup Integration**
   - Antigravity must appear as a selectable option under "Subscription modes" in both `quickstart.sh` and `scripts/setup_worker_model.sh`
   - Auto-detect existing Antigravity credentials and show `(credential detected)` when found
   - If credentials are not found, guide the user through OAuth login (provide the link to complete auth in-browser)
   - Display a warning on selection:
     > ⚠ Using Antigravity can technically cause your account suspension. Please use at your own risk.

2. **Token & Config Plumbing**
   - Token retrieval, refresh, config flags (`use_antigravity_subscription`), `api_base`, and `extra_kwargs` — for both queen and worker LLM paths
   - Follow the established pattern in `runner.py`, `config.py`, and `litellm.py`

3. **Verification**
   - Must include a screenshot or recording showing:
     - Quickstart selecting Antigravity and completing setup
     - A Hive agent successfully making an LLM call through the Antigravity subscription
   - Add basic integration tests (similar to existing `test_codex_eventloop.py`)

levxn:
@RichardTang-Aden I would like to take up this issue :))

PR #6648, opened 2026-03-19
Antigravity subscription support as an LLM provider

levxn:
Summary

This PR adds first-class support for Antigravity (Google's unified Cloud Code Assist gateway) as a selectable LLM provider for both the main queen agent and worker models. Antigravity routes requests to Gemini, Claude, and GPT-OSS models through a single Gemini-style API. Authentication is handled via Google OAuth2 directly — no local proxy or antigravity-auth serve required.

---

What changed

New file: core/framework/llm/antigravity.py

A full standalone LLM provider class (AntigravityProvider) implementing the LLMProvider interface. This is not built on LiteLLM — the provider owns the entire request lifecycle:

- Credential loading from two sources in priority order: the opencode-antigravity-auth JSON accounts file (~/.config/opencode/antigravity-accounts.json, V1–V4 schema with expiry support) and the Antigravity IDE's VSCode-style SQLite state database on macOS and Linux as a fallback.
- Token expiry tracking with a 60-second refresh buffer. Refresh is done directly against the Google OAuth2 token endpoint using the client secret read from config or environment.
- Endpoint fallback chain: daily sandbox → autopush sandbox → production, so requests automatically route to the next available endpoint on failure.
- Custom OpenAI-format to Gemini-format message conversion, including tool call/result batching, function name sanitisation (spaces and slashes to underscores, 64-char cap), and handling of Gemini's requirement that the first turn must be a user turn.
- SSE stream parsing that yields TextDeltaEvent, ToolCallEvent, FinishEvent, and StreamErrorEvent objects compatible with the existing event loop infrastructure.
- Thread-executor-backed async stream() to keep blocking HTTP I/O off the event loop.

core/framework/config.py

Added get_antigravity_client_id() and get_antigravity_client_secret() helpers that read from ANTIGRAVITY_CLIENT_ID / ANTIGRAVITY_CLIENT_SECRET environment variables first, then from the antigravity_client_id / antigravity_client_secret fields in configuration.json. The client ID has a well-known public fallback so token refresh works even when the config entry is absent.

core/framework/runner/runner.py

Added Antigravity credential detection and token loading at session initialisation. The runner checks both credential sources, performs an early token refresh if the stored token is close to expiry, and injects the bearer token into the provider at run time. Also added the credential detection flag (ANTIGRAVITY_CRED_DETECTED) used by the quickstart and session manager to decide whether to offer the Antigravity option.

core/framework/server/session_manager.py

Session creation now detects Antigravity credentials and selects AntigravityProvider when the configuration specifies use_antigravity_subscription. The provider is instantiated with the model from configuration.json (defaults to gemini-3-flash).

quickstart.sh and scripts/setup_worker_model.sh

Both scripts gain a new Antigravity option in the provider selection menu (option 8 for the main model, matching the worker model setup):

- If credentials are not yet present, the script guides the user through installing opencode-ai via npm (if not already installed), writing or updating ~/.config/opencode/opencode.json with the opencode-antigravity-auth@latest plugin entry, and running opencode auth login to complete the Google OAuth browser flow.
- After login the script re-checks for the accounts file and exits with a clear message if authentication did not produce credentials.
- The save_configuration / save_worker_configuration Python block writes use_antigravity_subscription, antigravity_client_secret, and antigravity_client_id into configuration.json. The secret and client ID are read from the installed npm package's constants.js at setup time rather than being hardcoded — no literal credentials appear in source.
- The informational display text now reflects the direct OAuth model ("no proxy required") instead of the old antigravity-auth serve instruction.

core/tests/test_antigravity_eventloop.py

Integration smoke test (mirrors the structure of test_codex_eventloop.py) that runs a real EventLoopNode against the live Antigravity backend, validates tool-use output via the set_output mechanism, and includes a fallback bare provider.stream() path for diagnosing streaming issues independently of the event loop.

---

How it works end-to-end

1. User runs quickstart and selects Antigravity.
2. If not already authenticated, the script installs the npm plugin and opens the Google OAuth consent page via opencode auth login.
3. Credentials are saved to ~/.config/opencode/antigravity-accounts.json by the plugin.
4. The quickstart reads the client secret and client ID from the npm package's constants.js and writes them into configuration.json alongside use_antigravity_subscription: true.
5. At agent startup, the runner detects the subscription flag, loads the credential file, refreshes the token if needed, and hands a live bearer token to AntigravityProvider.
6. All LLM calls go directly to the cloudcode-pa.googleapis.com endpoint (or sandbox variants) over HTTPS with no intermediary proxy.

---

Not changed

No existing provider (LiteLLMProvider, AnthropicProvider, OpenRouterProvider) was modified to accommodate this. The new provider is fully additive. The Codex subscription flow, Claude Code subscription flow, and all other provider paths are untouched.

---

Verification

To test manually:
- Authenticate with opencode auth login and confirm ~/.config/opencode/antigravity-accounts.json exists.
- Run quickstart, select option 8, and confirm configuration.json contains use_antigravity_subscription, antigravity_client_id, and antigravity_client_secret.
- Run core/tests/test_antigravity_eventloop.py directly (requires valid credentials) and confirm the EventLoopNode test passes and the stream test produces token counts.
- Confirm that grep for the literal OAuth client secret string returns no results in quickstart.sh or scripts/setup_worker_model.sh.

---

Screen Recording:

https://github.com/user-attachments/assets/69cf228b-6422-4603-b03b-5ba6978ba723

---

solves issue #6632

### DIFF
--- a/core/antigravity_auth.py
+++ b/core/antigravity_auth.py
@@ -1,0 +1,452 @@
+#!/usr/bin/env python3
+"""Antigravity authentication CLI.
+
+Implements OAuth2 flow for Google's Antigravity Code Assist gateway.
+Credentials are stored in ~/.hive/antigravity-accounts.json.
+
+Usage:
+    python -m antigravity_auth auth account add
+    python -m antigravity_auth auth account list
+    python -m antigravity_auth auth account remove <email>
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import os
+import secrets
+import socket
+import subprocess
+import sys
+import time
+import urllib.parse
+import urllib.request
+import webbrowser
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from pathlib import Path
+from typing import Any
+
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger(__name__)
+
+# OAuth endpoints
+_OAUTH_AUTH_URL = "https://accounts.google.com/o/oauth2/v2/auth"
+_OAUTH_TOKEN_URL = "https://oauth2.googleapis.com/token"
+
+# Scopes for Antigravity/Cloud Code Assist
+_OAUTH_SCOPES = [
+    "https://www.googleapis.com/auth/cloud-platform",
+    "https://www.googleapis.com/auth/userinfo.email",
+    "https://www.googleapis.com/auth/userinfo.profile",
+]
+
+# Credentials file path in ~/.hive/
+_ACCOUNTS_FILE = Path.home() / ".hive" / "antigravity-accounts.json"
+
+# Default project ID
+_DEFAULT_PROJECT_ID = "rising-fact-p41fc"
+_DEFAULT_REDIRECT_PORT = 51121
+
+# OAuth credentials fetched from the opencode-antigravity-auth project.
+# This project reverse-engineered and published the public OAuth credentials
+# for Google's Antigravity/Cloud Code Assist API.
+# Source: https://github.com/NoeFabris/opencode-antigravity-auth
+_CREDENTIALS_URL = "https://raw.githubusercontent.com/NoeFabris/opencode-antigravity-auth/dev/src/constants.ts"
+
+# Cached credentials fetched from public source
+_cached_client_id: str | None = None
+_cached_client_secret: str | None = None
+
+
+def _fetch_credentials_from_public_source() -> tuple[str | None, str | None]:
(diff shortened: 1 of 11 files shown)


RichardTang-Aden:
During testing encounter this error

`2026-03-19 18:46:56 [INFO    ] Antigravity token expired or near expiry, refreshing...
2026-03-19 18:46:56 [WARNING ] Antigravity token refresh failed. Re-open the Antigravity IDE or run 'antigravity-auth accounts add'.`

levxn:
@RichardTang-Aden @TimothyZhang7 let me know your review on this PR, thanks :))

Issue #5558, opened 2026-03-01
[Security] H-02: Bypassable attribute blocklist in safe_eval (High)


shibinsp:
visit_Attribute uses a name-based blocklist that is bypassable via type chain attacks. Fix: add isinstance() type allowlist restricting attribute access to safe built-in types only.

shibinsp:
I'd like to take this on! I've already prepared the fix with tests on branch `shibinsp/hive`. Ready to submit PR once assigned.

PR #6704, opened 2026-03-22
fix(security): restrict safe_eval attribute access to safe types and validate method receivers

MILTONADINA:
## Summary

Fixes a sandbox escape vulnerability in `SafeEvalVisitor` where:

- **Unrestricted `getattr()`**: `visit_Attribute` blocked `_`-prefixed attributes but called `getattr(val, node.attr)` on any object type. If the evaluation context contained custom objects (Pydantic models, config objects, etc.), all non-underscore attributes and methods were accessible.
- **Untyped method dispatch**: `visit_Call` checked method names (get, keys, lower, etc.) but not the receiver's type — so `.get()` on a malicious object with a custom `get()` method was allowed.

> **Note:** This completes the fix originally proposed in #5558 (H-02: Bypassable attribute blocklist in safe_eval), which was closed on March 2 without a fix being merged. The vulnerability still exists in the current `main` branch. This PR implements the type allowlist from #5558 and goes further by adding type-validated method dispatch via `_SAFE_METHOD_RECEIVERS`.

### Changes

1. **Type-restricted attribute access** — Added `_SAFE_ATTR_TYPES` allowlist. Attribute access is now only permitted on `dict`, `list`, `tuple`, `set`, `str`, `int`, `float`, `bool`, and `NoneType`.

2. **Type-validated method dispatch** — Added `_SAFE_METHOD_RECEIVERS` mapping. Method calls are validated against expected receiver types (e.g., `.get()` only on `dict`, `.lower()` only on `str`).

3. **20 new tests** covering:
   - Custom object attribute access blocked
   - Custom object with `.get()` method blocked
   - All existing dict/string method calls still work
   - Cross-type method calls blocked (`.get()` on list, `.lower()` on list)
   - Nested dict access patterns still work
   - Guard patterns (`x is not None and x.get('key')`) still work
   - Dataclass/Pydantic-like objects blocked

### All 130 tests pass (110 existing + 20 new).

Closes #6700
Related: #5558

## Test plan

- [x] All existing `test_safe_eval.py` tests pass (no regressions)
- [x] New `TestAttributeTypeSafety` test class covers attack vectors and regressions
- [x] `ruff check` and `ruff format --check` pass
- [ ] Verify with full `pytest tests/` suite

### DIFF
--- a/core/framework/graph/safe_eval.py
+++ b/core/framework/graph/safe_eval.py
@@ -2,6 +2,25 @@ import ast
 import operator
 from typing import Any
 
+# Types that are allowed for attribute access and method dispatch.
+# Objects not in this set cannot have their attributes read or methods called,
+# preventing sandbox escape via custom objects in the evaluation context.
+_SAFE_ATTR_TYPES = (dict, list, tuple, set, str, int, float, bool, type(None))
+
+# Mapping of allowed method names to the receiver types they may be called on.
+# This prevents a custom object with a `.get()` method from being treated like
+# a dict inside the sandbox.
+_SAFE_METHOD_RECEIVERS: dict[str, tuple[type, ...]] = {
+    "get": (dict,),
+    "keys": (dict,),
+    "values": (dict,),
+    "items": (dict,),
+    "lower": (str,),
+    "upper": (str,),
+    "strip": (str,),
+    "split": (str,),
+}
+
 # Safe operators whitelist
 SAFE_OPERATORS = {
     ast.Add: operator.add,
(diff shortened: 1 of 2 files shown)


MILTONADINA:
Hi maintainers (@TimothyZhang7 @RichardTang-Aden @Hundao @Antiarin) — I was invited by Vincent to contribute as part of the Applied AI Engineer evaluation. 

I cannot self-assign the linked issue (#6700) because the GitHub API returns 403 for non-collaborators. Could someone please assign me (@MILTONADINA) to #6700 so this PR passes the requirements check? 

This is a security fix for the safe_eval sandbox (unrestricted `getattr` + untyped method dispatch) with 20 new tests. All tests pass, ruff clean. Thank you!

PR #6706, opened 2026-03-22
fix(core): resolve race condition in event_bus debug log initialization

### LINKED ISSUE #6701: [Bug]: event_bus debug log initialization has race condition — concurrent coroutines open duplicate file handles
## Description

The debug event logging in `core/framework/runtime/event_bus.py` uses module-level global variables (`_event_log_file`, `_event_log_ready`) that are initialized lazily without any synchronization, creating a race condition under concurrent async execution.

### Root Cause

In the `publish()` method (around lines 303-307), the debug log file is initialized using a check-then-act pattern on global state **without an `asyncio.Lock`**:

```python
if _DEBUG_EVENTS_ENABLED:
    global _event_log_file, _event_log_ready
    if not _event_log_ready:
        _event_log_file = _open_event_log()   # ← not protected by any lock
        _event_log_ready = True
```

When multiple coroutines call `publish()` concurrently (which is the normal case — multiple streams publishing events simultaneously), two or more coroutines can:

1. Both read `_event_log_ready == False`
2. Both call `_open_event_log()`, opening separate file handles
3. Only the **last** assignment to `_event_log_file` survives — the earlier file handle is leaked (never closed, never flushed)

### Additional Problem: No Cleanup Path

The opened file handle is never closed — there is no shutdown hook, no `atexit` handler, and no integration with the `EventBus.shutdown()` lifecycle. In long-running agent processes, this is a resource leak.

### Impact

- **File handle leak**: Each race occurrence leaks one file descriptor. Long-running agents with frequent restarts/reconnects can exhaust file descriptors.
- **Lost events**: Events written to the leaked file handle are silently discarded since the global reference is overwritten.
- **Debug mode only**: This only affects deployments with `HIVE_DEBUG_EVENTS=1` enabled, but that's precisely when reliability of event capture matters most.

### Reproduction

1. Set `HIVE_DEBUG_EVENTS=1`
2. Start an agent with multiple concurrent entry points (webhooks + scheduler)
3. Both streams publish events simultaneously on startup
4. Check file descriptors — leaked handles visible via `lsof` or `/proc/<pid>/fd`

### Proposed Fix

Replace the module-level global pattern with an async-safe singleton class that uses `asyncio.Lock` for one-time initialization, and integrate file handle cleanup with the `EventBus` shutdown lifecycle (or an `atexit` handler).

I'd like to work on this — happy to submit a PR with the fix and tests.

MILTONADINA:
## Summary

- The debug event log (`HIVE_DEBUG_EVENTS=1`) used module-level global variables (`_event_log_file`, `_event_log_ready`) initialized lazily in `publish()` **without any `asyncio.Lock`**
- When multiple coroutines called `publish()` concurrently (normal in multi-stream agents), both could see `_event_log_ready == False`, both call `_open_event_log()`, and the first file handle was leaked (never closed)
- No cleanup path existed — the file handle was never closed on shutdown

### Changes

1. **Replaced global state with `_EventDebugLog` singleton** — uses `asyncio.Lock` to guarantee one-time initialization regardless of concurrent access
2. **Added proper lifecycle management** — `close()` flushes and closes the file handle; `reset()` discards the singleton for testing and shutdown
3. **Integrated cleanup with `EventBus.close_session_log()`** — the debug log is reset when the session log is closed, preventing file handle leaks

### 9 new tests covering:

| Test | What it verifies |
|------|-----------------|
| `test_singleton_returns_same_instance` | `get()` returns same object |
| `test_concurrent_get_initializes_once` | 20 parallel coroutines → 1 file open |
| `test_reset_allows_reinitialisation` | Fresh instance after `reset()` |
| `test_write_event_to_file` | JSONL serialisation works |
| `test_write_event_noop_when_file_is_none` | No crash when disabled |
| `test_close_closes_file_handle` | File handle properly closed |
| `test_publish_writes_to_debug_log` | End-to-end via `EventBus.publish()` |
| `test_publish_skips_debug_when_disabled` | No singleton created when disabled |
| `test_close_session_log_resets_debug_log` | Cleanup integration |

Closes #6701

## Test plan

- [x] All 9 new tests pass
- [x] `ruff check` and `ruff format --check` pass
- [ ] Verify existing event bus consumers are not affected (the public API is unchanged)

### DIFF
--- a/core/framework/runtime/event_bus.py
+++ b/core/framework/runtime/event_bus.py
@@ -52,8 +52,63 @@ def _open_event_log() -> IO[str] | None:
     return open(path, "a", encoding="utf-8")  # noqa: SIM115
 
 
-_event_log_file: IO[str] | None = None
-_event_log_ready = False  # lazy init guard
+class _EventDebugLog:
+    """Async-safe singleton for the debug event log file.
+
+    Replaces the previous module-level globals that were vulnerable to a race
+    condition: multiple concurrent ``publish()`` coroutines could both see the
+    ``_event_log_ready`` flag as ``False`` and open duplicate file handles,
+    leaking the first one.
+
+    The singleton uses an ``asyncio.Lock`` to guarantee one-time initialisation,
+    and exposes a ``close()`` method for orderly shutdown.
+    """
+
+    _instance: "_EventDebugLog | None" = None
+    _init_lock: asyncio.Lock | None = None
+
+    def __init__(self) -> None:
+        self._file: IO[str] | None = None
+
+    @classmethod
+    async def get(cls) -> "_EventDebugLog":
+        """Return the singleton, lazily opening the log file on first call."""
+        if cls._init_lock is None:
+            cls._init_lock = asyncio.Lock()
+        async with cls._init_lock:
+            if cls._instance is None:
+                inst = cls()
+                inst._file = _open_event_log()
+                cls._instance = inst
+            return cls._instance
+
+    def write_event(self, event: "AgentEvent") -> None:
+        """Serialise and append *event* to the debug log."""
+        if self._file is None:
+            return
+        try:
+            line = json.dumps(event.to_dict(), default=str)
+            self._file.write(line + "\n")
+            self._file.flush()
+        except Exception:
+            pass  # never break event delivery
+
+    def close(self) -> None:
+        """Flush and close the underlying file handle."""
+        if self._file is not None:
+            try:
+                self._file.close()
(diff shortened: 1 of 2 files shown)


MILTONADINA:
Hi maintainers (@TimothyZhang7 @RichardTang-Aden @Hundao @Antiarin) — could someone please assign me (@MILTONADINA) to #6701? I cannot self-assign (403 for non-collaborators). This PR fixes a race condition in event_bus debug log initialization with 9 tests. Thanks\!

PR #6710, opened 2026-03-22
feat(runtime): add idempotency key support to trigger()

### LINKED ISSUE #5947: Add idempotency key support for webhook-triggered executions to prevent duplicates
I was looking at how `AgentRuntime.trigger()` handles webhook-based entry points, and there's no deduplication mechanism. If a webhook source retries a delivery (which most providers do on timeout or 5xx), Hive will happily start a second execution for the same event.

**Why this matters:** Say you're using Hive to process Stripe webhook events for subscription changes. Stripe retries webhooks up to 3 times. Without idempotency, a single `customer.subscription.updated` event could kick off 3 separate execution streams — each one potentially making API calls, sending emails, or updating records. That's not just wasted compute; imagine getting 3 "your plan has been updated" emails.

**Suggested approach:**
1. Accept an optional `idempotency_key` param in `trigger()` / the webhook route
2. Maintain a bounded TTL cache (e.g., LRU with 1-hour expiry) of recently-seen keys
3. If a key is seen twice within the TTL, return the existing execution_id instead of starting a new one
4. For webhook routes, default to hashing `method + path + body` as the idempotency key if none is provided

This is a pretty standard pattern in payment systems and event-driven architectures — feels like a natural fit for Hive's webhook handling.

**Labels:** `enhancement`, `runtime`, `webhooks`

Rohit23SR:
Fixes #5947

Adds an optional `idempotency_key` param to `trigger()` and `trigger_and_wait()`. If the same key comes in again within the TTL window (default 5 min), we return the cached execution_id instead of starting a new run. Cache uses OrderedDict with TTL expiry and FIFO eviction at 10k keys. Placeholders + async lock handle the race between concurrent retries.

Backward compatible — defaults to None, existing callers unchanged.

20 tests added.

### DIFF
--- a/core/framework/runtime/agent_runtime.py
+++ b/core/framework/runtime/agent_runtime.py
@@ -9,6 +9,7 @@ import asyncio
 import logging
 import time
 import uuid
+from collections import OrderedDict
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from datetime import datetime
@@ -44,6 +45,9 @@ class AgentRuntimeConfig:
     max_history: int = 1000
     execution_result_max: int = 1000
     execution_result_ttl_seconds: float | None = None
+    # Idempotency cache for trigger() deduplication
+    idempotency_ttl_seconds: float = 300.0
+    idempotency_max_keys: int = 10000
     # Webhook server config (only starts if webhook_routes is non-empty)
     webhook_host: str = "127.0.0.1"
     webhook_port: int = 8080
@@ -247,6 +251,10 @@ class AgentRuntime:
         self._timer_tasks: list[asyncio.Task] = []
         # Next fire time for each timer entry point (ep_id -> datetime)
         self._timer_next_fire: dict[str, float] = {}
+
+        # Idempotency cache for trigger() deduplication
+        self._idempotency_keys: OrderedDict[str, str] = OrderedDict()
+        self._idempotency_times: dict[str, float] = {}
 
         # State
         self._running = False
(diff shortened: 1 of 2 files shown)


Hundao:
There's a race condition with the placeholder. If trigger A is mid-execution and trigger B hits with the same key, B gets back `"pending-abc123"`, which isn't a real execution_id. Caller can't do anything useful with that.

How about dropping the placeholder entirely? Just don't write to the cache until execute() finishes. That way the cache only ever holds real execution_ids. Webhook retries are usually seconds to minutes apart, so two triggers hitting within milliseconds shouldn't really happen in practice.

Also, this uses `self._lock` which is the same lock as start/stop. Separate `_idempotency_lock` would be cleaner.

Rohit23SR:
Good catch — you're right, returning a pending placeholder breaks the contract since callers use the ID for tracking.

I dropped the placeholder and the lock entirely. The cache only gets written after execute() returns, so it always holds a real execution_id. No separate lock needed either — without a placeholder there's nothing to protect atomically, and the dict operations between awaits run on the event loop without interleaving.

The tradeoff is a tiny window where a concurrent trigger could slip through, but execute() is non-blocking (just creates a task and returns), so that window is microseconds — and webhook retries are seconds apart anyway.

Pushed the update.

Rohit23SR:
Hey @Hundao — just checking in. I pushed the update dropping the placeholder and lock as discussed. Let me know if there's anything else you'd like changed.

PR #6727, opened 2026-03-22
perf(test): enable parallel test execution with pytest-xdist

Rohit23SR:
Fixes #6612

Wires up pytest-xdist for parallel test execution. xdist was already a dependency in core but never used — this adds `-n auto` to the Makefile, CI, and release workflows. Also adds the missing xdist dependency to tools and sets `asyncio_mode = "auto"` in core's pytest config to match tools.

test-live is excluded since `-s` is incompatible with xdist workers.

Tested locally — core and tools both pass with identical results in parallel, roughly 2x faster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
         working-directory: core
         run: |
           uv sync
-          uv run pytest tests/ -v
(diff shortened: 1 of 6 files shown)


Rohit23SR:
Hey — any chance someone could approve the CI workflow run on this one? Happy to address any feedback.

PR #6751, opened 2026-03-23
Enhancements/UI - UI Polish: Worker Collapsibles, Ask-User Reply Thread, Session Discoverability

levxn:
### Summary

- Fixed stuck "Working" label on completed worker collapsibles with no duration
- Removed redundant CPU icon from collapsible header; scoped gradient animation to active node only
- Threaded consecutive same-node messages inside collapsibles — only the first message in a run shows the node header
- Fixed tool pills (web_scrape, set_output, etc.) not appearing in resumed and cold-restored sessions
- Built Discord-style reply thread for ask_user questions from both worker and queen nodes
- Added persistent "Resume session" button on the home page so users no longer need to discover the slash command
- Made session picker arrow navigation wrap cyclically instead of clamping at boundaries

### Worker Collapsible Fixes

The collapsible header showed a stuck "Working" label on completed groups that had only one message or shared identical timestamps — even though the spinner was absent. All non-active groups now correctly show "Worked" or "Worked for Xm Ys".

The CPU icon has been removed from the header. The live gradient animation on the label text is preserved while the worker is running, and the node name inside the collapsible now also animates while active. The gradient is scoped to the currently running node only — earlier completed nodes in the same group revert to static color.

Inside an expanded collapsible, consecutive messages from the same node no longer each repeat the full node name and Worker badge. Only the first message shows the header; subsequent messages from the same node show a thin vertical thread line. When the node changes, the new node's header appears fresh.

### Tool Pill Visibility on Session Resume

Tool pills (activity chips like `web_scrape x12`, `set_output`) were invisible in resumed and cold-restored sessions — only live sessions rendered them. This is now fixed so tool activity displays consistently across all session states.

### Ask-User Reply Thread

The ask_user flow from worker and queen nodes now renders correctly in the chat.

Worker questions were previously absorbed silently into the worker group collapsible instead of appearing in the message history. Queen questions were never stored in the message history at all — only held in ephemeral UI state — so the user's reply showed as a plain bubble with no context.

Both cases now produce a Discord/Snapchat-style reply thread: the question text appears as a small dimmed quote with a colored left-border accent above the right-aligned reply bubble. The pairing logic uses a tracked variable through the render loop rather than positional adjacency, so other messages appearing between the question and reply do not break it.

### Home Page Session Discoverability

The resume option was only reachable by users who already knew to type `/resume` — effectively invisible. A persistent "Resume session" button is now in the action row on the home page alongside "Try a sample agent" and "My Agents".

Keyboard navigation in the session picker modal now wraps cyclically: down on the last item moves to the first, up on the first moves to the last.

### Test Plan

- [ ] Start a single-message worker task and verify the collapsible shows "Worked" not "Working" after completion
- [ ] Expand a multi-node collapsible (e.g. Research then Compile Report) and confirm only the first message per node shows the header, and only the active node name animates
- [ ] Resume a previous session and confirm tool pills are visible on historical messages
- [ ] Trigger an ask_user from a worker node, answer it, and confirm the reply thread renders with the question quoted above the answer
- [ ] Trigger an ask_user from the queen node (staging phase), answer it, and confirm the same reply thread renders
- [ ] Visit the home page and confirm "Resume session" button is visible without typing anything
- [ ] Open the session picker, navigate to the last item with the down arrow, and confirm it wraps to the first

---
### Visuals
Screenshot 2026-03-23 at 11 47 30 AM
Screenshot 2026-03-23 at 4 40 07 PM

https://github.com/user-attachments/assets/be9fea30-b79c-454e-b2fc-e6d4ec879065

Screenshot 2026-03-24 at 1 18 00 AM

---

issue #5904 , this is clean working branch intact with latest, hence following up on my previous closed PR #6095


levxn:
@TimothyZhang7 @RichardTang-Aden let me know how this UI looks :))

levxn:
https://github.com/user-attachments/assets/87a24a7a-b9af-48e3-ad64-ff7c7d6d5b69

includes only the worker collapsible UI part.

RichardTang-Aden:
image


During testing found this bug:

I have a session that I terminated earlier. After I resume the session (shut down tab and the script, resume from my agent), the worker can be rerun be the worker message are aggregated in the previous messages above my recent conversation with the queen. I can't see the worker message in the latest run. The new worker message group is suppose to happen after my queen say "Yes I will rerun it"

PR #6786, opened 2026-03-24
fix: serialize _write_progress() race condition with lock + atomic_write

genyarko:
## Summary

- **Add `threading.Lock`** to `_write_progress()` to serialize the read-modify-write cycle on `state.json`, preventing concurrent fan-out branches from clobbering each other's progress updates
- **Add 21 tests** covering basic functionality, atomic write integrity, concurrency stress (up to 50 threads × 5 writes), and edge cases

Closes #6696

## Changes

| File | Change |
|---|---|
| `core/framework/graph/executor.py` | Add `import threading`, `self._progress_lock` in `__init__()`, rewrite `_write_progress()` with lock to serialize concurrent writes |
| `core/tests/test_executor_progress_race.py` | New — 21 tests across 4 categories |

## Context

During `_execute_parallel_branches()`, multiple branches call `_write_progress()` concurrently. Without synchronization, threads read the same stale `state.json`, patch independently, and silently overwrite each other's updates. The `threading.Lock` ensures each writer reads the state left by the previous writer. The existing `atomic_write` swap prevents readers from ever seeing a partial file.

No new dependencies — uses only stdlib `threading` and the project's existing `framework.utils.io.atomic_write`.

## Test plan

- [ ] Verify 21 new tests pass (`pytest core/tests/test_executor_progress_race.py`)
- [ ] Verify no regressions in existing executor tests (`pytest core/tests/test_graph_executor.py`)
- [ ] Verify full test suite passes

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -11,6 +11,7 @@ The executor:
 
 import asyncio
 import logging
+import threading
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -230,6 +231,9 @@ class GraphExecutor:
         # Track the currently executing node for external injection routing
         self.current_node_id: str | None = None
 
+        # Serialize read-modify-write cycles on state.json
+        self._progress_lock = threading.Lock()
+
     def _write_progress(
         self,
         current_node: str,
@@ -244,6 +248,10 @@ class GraphExecutor:
         state.json as the single source of truth — readers always see
         current progress, not stale initial values.
 
+        The entire read-modify-write cycle is serialized with a threading
+        lock to prevent concurrent fan-out branches from clobbering each
+        other's progress updates.
+
         The write is synchronous and best-effort: never blocks execution.
         """
         if not self._storage_path:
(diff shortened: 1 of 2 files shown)


levxn (commented):
Kindly submit your branch which is upto date with the latest upstream. I see your current branch is on a very old head. can you and push that one again? Will review your PR

genyarko:
> Kindly submit your branch which is upto date with the latest upstream. I see your current branch is on a very old head. can you and push that one again? Will review your PR

Done

levxn (commented):
Everything looks good to me

HarshTomar1234:
I was independently working on this fix and arrived at the same approach — threading.Lock to serialize the read-modify-write cycle + atomic_write to prevent crash corruption. Reviewed the changes and everything looks correct:

Lock correctly wraps the entire read → patch → write cycle in 

_write_progress()
atomic_write matches what SessionStore.write_state() already uses
No new dependencies — just stdlib threading + existing framework.utils.io.atomic_write
Best-effort semantics preserved (progress writes never block execution)
Tests cover concurrency stress, lock serialization, and edge cases
LGTM....

Hundao:
Hi @genyarko, CI lint and tests are failing, can you take a look?

Also a heads up: PR description says "Replace bare write_text() with atomic_write()", but main already uses atomic_write. The actual change here is adding the threading.Lock, which is the right fix. Might want to update the description.

genyarko:
The PR description was inaccurate from the start. The only actual change our PR makes to executor.py is adding the     
  threading.Lock.

bryanadenhq:
Hey @genyarko, appreciate the work here but I think the race condition might not actually exist. _write_progress is only called from the main execution loop (lines 1293, 1351, 1416), never from inside the parallel branches, execute_single_branch doesn't call it. And _execute_parallel_branches uses asyncio.gather, not threads, so there's no concurrent access to serialize.

Also worth noting threading.Lock wouldn't help with async concurrency anyway, you'd need asyncio.Lock for that.

Could you share a repro or stack trace from #6696 showing the clobber? Possible I'm missing a code path.

genyarko:
You're right — I traced through the call sites again and `_write_progress` is only called from the main execution loop (lines 1302, 1360, 1425), never from within `execute_single_branch` or the parallel branches. `asyncio.gather` runs the branches concurrently but the progress writes happen sequentially before/after that call, so there's no concurrent access to serialize.

And good catch on `threading.Lock` vs `asyncio.Lock` — even if there were async concurrency here, a threading lock wouldn't help.

I wasn't able to independently reproduce the clobber from #6696; I assumed concurrent access based on the issue description rather than verifying the call graph myself. That's on me.

I'll close this PR. If #6696 does turn out to have a real repro pointing to a different code path, the `atomic_write` portion might still be worth revisiting, but the lock isn't solving an actual problem as-is. Sorry for the noise.

PR #6847, opened 2026-03-28
fix: reset node_retry_counts on fresh node visits to restore retry budget in feedback-loop graphs (Fixes #6605)

EsraaKamel11:
## Problem

  In graphs where an `ON_FAILURE` edge routes back to a previously-failed node,
  that node permanently loses its retry budget after the first failure cycle.
  From visit 2 onward it gets **zero retries** — every transient failure routes
  immediately to failure handling again, making self-healing graphs non-functional.

  ## Root cause

  `node_retry_counts` is initialised once per `execute()` call and only ever
  incremented. It is never reset when the main loop starts a fresh visit to a node.

  The per-visit bookkeeping block at lines 820–824 correctly scopes
  `node_visit_counts` per visit, but has no corresponding reset for
  `node_retry_counts`:

  ```python
  # executor.py  lines 820–824  (before fix)
  if not _is_retry:
      cnt = node_visit_counts.get(current_node_id, 0) + 1
      node_visit_counts[current_node_id] = cnt
      # ← node_retry_counts[current_node_id] never reset here
  _is_retry = False
  ```

  The retry gate at line 1100 uses the cumulative total across all visits:

  ```python
  if node_retry_counts[current_node_id] < max_retries:   # stale total
      _is_retry = True
      continue
  else:
      # routes to ON_FAILURE — fires immediately on visit 2
  ```

  By the time a node reaches its second visit via a feedback edge,
  `node_retry_counts[node_id]` is already `>= max_retries`. The check fires
  on the very first failure of the new visit, granting zero retries.

  ## Failure trace

  Graph: `DataFetcher (max_retries=3) --ON_FAILURE--> ErrorRecovery --> DataFetcher`

  | Visit | Attempt | `node_retry_counts["DataFetcher"]` | Outcome |
  |---|---|---|---|
  | 1 | 1 | 1 | retry |
  | 1 | 2 | 2 | retry |
  | 1 | 3 | 3 | `3 < 3` → False → route to ErrorRecovery |
  | 2 | 1 | 4 | `4 < 3` → False → **route to ErrorRecovery, 0 retries granted** |
  | 3 | 1 | 5 | `5 < 3` → False → **route to ErrorRecovery, 0 retries granted** |

  Each subsequent visit depletes the counter further. The run either hits
  `max_steps` in an infinite failure loop, or the graph terminates if no
  further `ON_FAILURE` edge exists from `ErrorRecovery`.

  ## Why it's hard to spot in production

  **The log looks like legitimate overshoot, not a missing reset.**
  The line `✗ Max retries (3) exceeded for node DataFetcher` appears on every visit.
  With a counter of 4 against a max of 3 it reads as continued retrying, not as a
  budget that was never replenished for the new visit.

  **Only manifests in graphs with ON_FAILURE feedback edges.**
  Linear graphs and graphs with terminal failure paths are completely unaffected.
  The bug surfaces exclusively in graphs designed for self-healing — the exact graphs
  where retry reliability matters most.

  **`EventLoopNode` is immune**, masking the pattern in most integration tests.
  The executor unconditionally overrides `max_retries` to `0` for `EventLoopNode`
  instances (lines 1092–1097). Most high-level agent nodes are `EventLoopNode`, so
  standard tests rarely exercise the broken path.

  ## Fix

  One line added to the fresh-visit bookkeeping block at line 823:

  ```python
  # executor.py  lines 820–825  (after fix)
  if not _is_retry:
      cnt = node_visit_counts.get(current_node_id, 0) + 1
      node_visit_counts[current_node_id] = cnt
      node_retry_counts[current_node_id] = 0  # fresh visit = fresh retry budget
  _is_retry = False
  ```

  **Why this is correct:**

  `_is_retry = True` causes the loop to `continue` before reaching this block,
  so intra-visit retry iterations still accumulate the counter correctly — the
  exponential backoff formula at lines 1105–1107 (`1.0 * (2 ** (retry_count - 1))`)
  continues to work exactly as designed within a single visit.

  `_is_retry = False` (fresh visit from a new edge traversal) resets the counter
  to `0`, giving the node its full `max_retries` allocation regardless of prior
  visit history. This is symmetric with `node_visit_counts` handling, which was
  always the intended design.

  ## Tests added

  `core/tests/test_retry_count_reset_on_revisit.py` — 3 new tests, all passing.

  **`test_retry_budget_reset_on_revisit_via_on_failure`**
  - `FlakyNode (max_retries=3) --ON_FAILURE--> RecoveryNode --> FlakyNode`
  - Node fails on calls 1–3 (visit 1 exhausts budget) → `RecoveryNode` →
    node succeeds on call 4 (visit 2, attempt 1)
  - Asserts `attempt_count == 4` and `result.success is True`
  - Without the fix: `node_retry_counts["flaky"]` is `3` at visit 2 start;
    `3 < 3` is `False`; node is immediately re-routed with 0 retries, never
    reaches call 4

  **`test_retry_budget_independent_per_visit`**
  - Same topology; node fails on calls 1–3 AND 4–6, succeeds on call 7
  - Asserts `attempt_count == 7` and `RecoveryNode.execute_count == 2`
  - Verifies each successive re-visit receives its own independent full budget

  **`test_linear_graph_retry_unaffected`**
  - Single node, no feedback loop; fails twice then succeeds
  - Asserts `attempt_count == 3` and `result.success is True`
  - Confirms the fix introduces no regression in the common single-visit case

  ## Existing test coverage (unchanged, all still passing)

  | File | What it covers |
  |---|---|
  | `test_executor_max_retries.py` | Single-node retry with configurable `max_retries` |
  | `test_on_failure_edges.py` | ON_FAILURE routing to a terminal handler |
  | `dummy_agents/test_retry.py` | Retry exhaustion + ON_FAILURE to terminal error handler |
  | `test_executor_feedback_edges.py` | Feedback loops, `max_node_visits`, visit counts |

  None of these covered the retry + feedback-loop re-visit combination. The gap
  is now closed by the three tests above.
  
## Not covered by existing issues

| Issue | Topic | Overlap |
|---|---|---|
| #3532 | Duplicate counter increments + undefined variable | Different failure mode |
| #2410 | Retry counts not persisted across pause/resume boundaries | Different scope, cross-lifecycle, not within-run feedback loops |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry tracking: retry budgets are refreshed on node re-entry, while cumulative retry totals are now used for execution metrics, failure reporting, checkpoint cleanliness, and final execution outcomes.
* **Tests**
  * Added regression coverage to verify retry-budget reset on revisits, independent behavior across multiple revisit cycles, and unchanged retry behavior for single-pass linear graphs.

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -550,7 +550,8 @@ class GraphExecutor:
         path: list[str] = []
         total_tokens = 0
         total_latency = 0
-        node_retry_counts: dict[str, int] = {}  # Track retries per node
+        node_retry_counts: dict[str, int] = {}  # Per-visit retry budget (reset on fresh visits)
+        node_retry_totals: dict[str, int] = {}  # Cumulative retry counts (never reset)
         node_visit_counts: dict[str, int] = {}  # Track visits for feedback loops
         _is_retry = False  # True when looping back for a retry (not a new visit)
 
@@ -821,6 +822,7 @@ class GraphExecutor:
                 if not _is_retry:
                     cnt = node_visit_counts.get(current_node_id, 0) + 1
                     node_visit_counts[current_node_id] = cnt
+                    node_retry_counts[current_node_id] = 0  # fresh visit = fresh retry budget
                 _is_retry = False
                 max_visits = getattr(node_spec, "max_node_visits", 0)
                 if max_visits > 0 and node_visit_counts[current_node_id] > max_visits:
@@ -946,7 +948,7 @@ class GraphExecutor:
                         current_node=node_spec.id,
                         execution_path=list(path),
                         memory=memory,
-                        is_clean=(sum(node_retry_counts.values()) == 0),
+                        is_clean=(sum(node_retry_totals.values()) == 0),
                     )
 
                     if checkpoint_config.async_checkpoint:
@@ -1080,6 +1082,9 @@ class GraphExecutor:
                     node_retry_counts[current_node_id] = (
                         node_retry_counts.get(current_node_id, 0) + 1
                     )
+                    node_retry_totals[current_node_id] = (
+                        node_retry_totals.get(current_node_id, 0) + 1
+                    )
 
                     # [CORRECTED] Use node_spec.max_retries instead of hardcoded 3
                     max_retries = getattr(node_spec, "max_retries", 3)
@@ -1166,8 +1171,8 @@ class GraphExecutor:
                             )
 
                             # Calculate quality metrics
-                            total_retries_count = sum(node_retry_counts.values())
-                            nodes_failed = list(node_retry_counts.keys())
+                            total_retries_count = sum(node_retry_totals.values())
+                            nodes_failed = list(node_retry_totals.keys())
 
                             if self.runtime_logger:
                                 await self.runtime_logger.end_run(
@@ -1199,7 +1204,7 @@ class GraphExecutor:
                                 path=path,
                                 total_retries=total_retries_count,
                                 nodes_with_failures=nodes_failed,
-                                retry_details=dict(node_retry_counts),
+                                retry_details=dict(node_retry_totals),
                                 had_partial_failures=len(nodes_failed) > 0,
                                 execution_quality="failed",
                                 node_visit_counts=dict(node_visit_counts),
@@ -1237,8 +1242,8 @@ class GraphExecutor:
                     )
 
                     # Calculate quality metrics
-                    total_retries_count = sum(node_retry_counts.values())
-                    nodes_failed = [nid for nid, count in node_retry_counts.items() if count > 0]
+                    total_retries_count = sum(node_retry_totals.values())
+                    nodes_failed = [nid for nid, count in node_retry_totals.items() if count > 0]
                     exec_quality = "degraded" if total_retries_count > 0 else "clean"
 
                     if self.runtime_logger:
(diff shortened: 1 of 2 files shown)


EsraaKamel11:
This PR targets `core/framework/graph/executor.py`, which no longer exists on `main` — the graph executor was replaced by the event-driven `orchestrator/` in the architecture revamp (`93dc35dc` *refactor(architecture): revamp*; the old `graph/executor.py` was removed and `orchestrator/orchestrator.py` finalized in `172b1804` *refactor: remove deprecated shims*). I verified the #6605 bug (per-visit retry budget not reset on feedback-loop revisits) **cannot occur** in the new architecture, so this is superseded.

Evidence on current `main`:

- **Retry budget is per-execution, not persistent:** `NodeWorker._execute_with_retries` bounds retries with a local `for attempt in range(total_attempts)` loop, `total_attempts = max(1, node_spec.max_retries)`, recomputed every call (`node_worker.py:352-366`). No across-visit budget counter exists to leak.
- **Workers are reset on feedback-loop re-entry:** `_route_activation` calls `reset_for_revisit()` when an activation targets a COMPLETED/FAILED worker (`orchestrator.py:1430` → `node_worker.py:219`), returning it to PENDING before it re-runs.
- **The only persistent retry counter is metrics-only:** `gc.retry_counts` feeds `total_retries`/`retry_details` (`orchestrator.py:1790-1792`) and never gates a retry — the sole retry/visit gates are the local `range()` loops (`node_worker.py:366/381/411`, `orchestrator.py:1017`) and the `max_node_visits` cap (`node_worker.py:256`).

A node entering a revisit always gets a full, fresh retry budget. Closing as superseded by `93dc35dc`/`172b1804`; the original two-counter `executor.py` fix is no longer applicable.

PR #6913, opened 2026-04-02
docs(tools): add README for 10 tools (batch 3)

KRYSTALM7:
Adds README.md documentation for 10 tools that were missing it.

Partial fix for #6486

## Tools Documented
`supabase_tool`, `zoom_tool`, `twitter_tool`, `twilio_tool`, `shopify_tool`,
`snowflake_tool`, `zendesk_tool`, `yahoo_finance_tool`, `youtube_transcript_tool`,
`docker_hub_tool`

## Format
Each README follows the existing pattern (e.g., `gmail_tool/README.md`):
- Tool name and description
- Tools table with all available functions
- Setup instructions with required credentials and env vars
- Usage examples
- Error handling

## Checklist
- [x] Follows existing README format
- [x] Self-reviewed
- [x] No unrelated files included

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive README guides for Docker Hub, Shopify, Snowflake, Twilio, Twitter/X, Yahoo Finance, YouTube Transcript, Zendesk, Zoom, and Supabase tools — each describing available tool entry points, setup and required credentials/scopes, usage examples, supported parameters/filters, expected response shapes, async behaviors, and standardized error-handling contracts.

### DIFF
--- a/tools/src/aden_tools/tools/docker_hub_tool/README.md
+++ b/tools/src/aden_tools/tools/docker_hub_tool/README.md
@@ -1,0 +1,124 @@
+# Docker Hub Tool
+
+Search repositories, list tags, inspect images, manage webhooks, and delete tags via the Docker Hub API v2.
+
+## Tools
+
+| Tool | Description |
+|------|-------------|
+| `docker_hub_search` | Search Docker Hub for public repositories |
+| `docker_hub_list_repos` | List repositories for a user or organization |
+| `docker_hub_get_repo` | Get detailed info about a specific repository |
+| `docker_hub_list_tags` | List tags for a repository |
+| `docker_hub_get_tag_detail` | Get details for a specific image tag |
+| `docker_hub_delete_tag` | Delete a tag from a repository |
+| `docker_hub_list_webhooks` | List webhooks configured for a repository |
+
+## Setup
+
+Requires a Docker Hub Personal Access Token (PAT):
+
+1. Go to [hub.docker.com](https://hub.docker.com) → **Account Settings → Security → New Access Token**
+2. Give it a name and select the required permissions (Read, Write, Delete as needed)
+3. Copy the token immediately — it is only shown once
+
+```bash
+DOCKER_HUB_TOKEN=your-personal-access-token
+DOCKER_HUB_USERNAME=your-docker-hub-username
+```
+
+> `DOCKER_HUB_USERNAME` is used as the default namespace when listing repos. If it is unset and no `namespace` is passed to `docker_hub_list_repos`, the tool will return an error: `"namespace is required (or set DOCKER_HUB_USERNAME)"`.
+
+## Usage Examples
+
+### Search for public repositories
+
+```python
+docker_hub_search(query="nginx", max_results=10)
+```
+
+### List your own repositories
+
+```python
+docker_hub_list_repos(namespace="myusername", max_results=25)
+```
+
+### Get repository details
+
+```python
+docker_hub_get_repo(repository="library/nginx")
+```
+
+### List tags for a repository
+
+```python
+docker_hub_list_tags(repository="library/nginx", max_results=20)
+```
+
+### Get details for a specific tag
+
+```python
+docker_hub_get_tag_detail(
+    repository="library/nginx",
+    tag="latest",
+)
+```
+
+### Delete a tag
+
+```python
+docker_hub_delete_tag(
+    repository="myusername/myapp",
+    tag="old-release-1.0",
+)
+```
+
+### List webhooks for a repository
+
+```python
+docker_hub_list_webhooks(repository="myusername/myapp")
+```
+
+## Response Format
+
+`docker_hub_list_tags` returns tags sorted by `last_updated` descending:
+
+```python
+{
+    "repository": "library/nginx",
+    "tags": [
+        {
+            "name": "latest",
+            "full_size": 68000000,
+            "last_updated": "2025-05-01T12:00:00Z",
+            "digest": "sha256:abc123...",
+        },
+        ...
+    ]
+}
+```
+
+`docker_hub_get_tag_detail` includes per-architecture image info:
+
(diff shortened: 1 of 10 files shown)


KRYSTALM7:
All 8 CI checks passing ✅

- Lint, framework tests (Ubuntu + Windows), tools tests (Ubuntu + Windows), agent export validation, and PR requirements — all green.

Addressing the CodeRabbit suggestions:

- **Shopify**: Will add `read_draft_orders`, `write_draft_orders` to the required scopes section
- **Yahoo Finance**: Will change "real-time" to "latest available" and add a note about exchange delays
- **Yahoo Finance intervals**: Will add a compatibility note about intraday interval limits (1m → 7 days, 1h → 60 days)
- **Zendesk error block**: Will wrap the error examples in a list `[...]` to fix the Python syntax
- **Zoom token expiry**: Will add a note that Server-to-Server OAuth tokens expire after 1 hour and need manual refresh

Will push fixes shortly.

KRYSTALM7:
@Hundao could you please review when you get a chance?

JiwaniZakir (commented):
The `fulfillment_status` values in `shopify_tool/README.md`'s Order Status Values table (`shipped`, `unshipped`, `partial`, `on_hold`) don't match Shopify's Admin REST API — the actual accepted values are `fulfilled`, `null` (for unfulfilled), `partial`, and `on_hold`. The example on line 83 (`fulfillment_status="unshipped"`) would silently return unexpected results since Shopify ignores unrecognized filter values rather than erroring. All three READMEs are also missing a trailing newline (`No newline at end of file`), which is a minor but consistent style issue worth fixing before merge. In `docker_hub_tool/README.md`, it would be worth noting what the fallback behavior is when `DOCKER_HUB_USERNAME` is unset and no `namespace` is passed to `docker_hub_list_repos` — currently the note only says it's optional, but a user has no way to predict whether the tool will error or return an empty result.

KRYSTALM7:
@JiwaniZakir - Fixed all remaining issues:

- **Shopify `fulfillment_status`**: Corrected values to match actual Shopify Admin REST API — `fulfilled`, `partial`, `on_hold`, `null` (unfulfilled).Also fixed the example which incorrectly used `"unshipped"`
- **Docker Hub**: Clarified `DOCKER_HUB_USERNAME` fallback behavior — tool returns an explicit error when both `namespace` param and env var are unset

JiwaniZakir:
The READMEs look solid overall. One thing worth checking: the `snowflake_tool` README should clarify whether `SNOWFLAKE_PRIVATE_KEY` expects a raw PEM string or a file path, since both patterns exist in similar tools and that ambiguity tends to cause setup friction. Also worth verifying that the `youtube_transcript_tool` README notes the dependency on `youtube_transcript_api` in the setup section, since it's a non-obvious requirement compared to the credential-based tools.

KRYSTALM7:
Thanks for the thorough review! Both points are already covered — the Snowflake tool accepts a pre-generated token via SNOWFLAKE_TOKEN rather than a raw private key, so SNOWFLAKE_PRIVATE_KEY doesn't apply here. And the youtube-transcript-api install instruction is already in the Setup section. Let me know if anything else needs attention!

JiwaniZakir:
Thanks for the clarification on the MCP loading path — you're right that `run` does call `_setup()` which invokes `load_hive_tools_registry()`, so the note was inaccurate. Please update or remove that note in the relevant README to reflect the actual behavior. Once that's addressed, the remaining items look good to me.

KRYSTALM7:
Thanks for the context! I searched through all the files in this PR but couldn't find any note mentioning load_hive_tools_registry() or _setup(). Could you point me to the exact file.

JiwaniZakir:
Skimmed through a few of these — the format is consistent and matches the existing pattern. One thing worth double-checking: some of the env var names in the setup sections (e.g., `SNOWFLAKE_ACCOUNT`, `ZOOM_CLIENT_SECRET`) should be verified against what the actual tool implementations expect, since mismatches there tend to cause confusion for first-time users. Also, the `youtube_transcript_tool` README should clarify whether authentication is required or if it works without credentials for public videos.

KRYSTALM7:
Thanks for the review! The env var names match the implementations — SNOWFLAKE_ACCOUNT is correct in the code, and the Zoom README only uses ZOOM_ACCESS_TOKEN with no ZOOM_CLIENT_SECRET. Updated the youtube_transcript_tool README to explicitly clarify that no authentication is needed for public videos.

JiwaniZakir:
Those functions aren't in this PR since it's docs-only — `load_hive_tools_registry()` and `_setup()` would be in the tool implementation files, not the READMEs. Check the `__init__.py` or main module file within each tool's directory (e.g., `supabase_tool/supabase_tool.py`) for `_setup()`, and the registry loader is likely in a top-level `hive/tools/__init__.py` or similar registry file outside of the individual tool folders.

KRYSTALM7:
Not sure why CI tests have been running from so long  and two of them are failing.

JiwaniZakir:
The `load_hive_tools_registry()` function is typically in `hive/tools/__init__.py` or a dedicated `registry.py` file — search for it with `grep -r "load_hive_tools_registry" .` from the repo root. If the READMEs are meant to include usage via the registry, each one should show both the direct import pattern and the registry-based initialization, since that's how most of the other documented tools expose themselves. Worth checking `gmail_tool/README.md` as the reference format to see if it includes registry setup steps.

JiwaniZakir:
Looks good overall — one thing worth double-checking: make sure the env var names listed in each README exactly match what the tool code actually reads (e.g., `TWITTER_API_KEY` vs `TWITTER_BEARER_TOKEN`), since mismatches there tend to trip up new contributors. Also, if any of these tools support optional credentials (like Snowflake's `warehouse` or `role`), it's worth calling those out explicitly in the setup section rather than leaving them implied.

PR #6934, opened 2026-04-03
feat: add per-node latency tracking and tests for executor

utk2602:
```
## Description
Implements per-node performance metrics (latency tracking) in the core `GraphExecutor`. Hooks up the existing `NodeResult.latency_ms` field with high-resolution `time.perf_counter()` timers around every `node_impl.execute()` call. Per-node latencies are aggregated into a new `ExecutionResult.node_latencies` dict, giving developers node-by-node visibility into execution time.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #6926

## Changes Made

- Added `import time` and wrapped both sequential and parallel `node_impl.execute()` calls with `time.perf_counter()` timers
- Added `node_latencies: dict[str, int]` field to `ExecutionResult` dataclass (default empty dict — non-breaking)
- If a node leaves `latency_ms` at 0, the executor fills it from the measured duration; if the node already set it, the value is preserved
- Passed `node_latencies` to all 6 in-loop `ExecutionResult` construction sites (2 early-return validation sites correctly omitted — no nodes executed)
- Added `SlowNode` test helper and 2 new tests: `test_executor_per_node_latency_tracking` (verifies executor-measured latency) and `test_executor_preserves_node_set_latency` (verifies node-preset latency is not overridden)

## Testing

## Checklist

- [x] I have made corresponding changes to the documentation

N/A — no UI changes.
```

## Summary by CodeRabbit

* **New Features**
  * Per-node execution latency tracking is recorded, summed into a total, preserved when nodes report their own latency, and persisted into session progress/state and progress writes.

* **Tests**
  * Added tests validating executor-measured per-node latencies, preservation of node-reported latencies, and correct aggregation across sequential and parallel branch executions.

### DIFF
--- a/core/framework/graph/executor.py
+++ b/core/framework/graph/executor.py
@@ -11,6 +11,7 @@ The executor:
 
 import asyncio
 import logging
+import time
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -70,6 +71,9 @@ class ExecutionResult:
 
     # Visit tracking (for feedback/callback edges)
     node_visit_counts: dict[str, int] = field(default_factory=dict)  # {node_id: visit_count}
+
+    # Per-node latency breakdown
+    node_latencies: dict[str, int] = field(default_factory=dict)  # {node_id: latency_ms}
 
     @property
     def is_clean_success(self) -> bool:
@@ -561,6 +565,7 @@ class GraphExecutor:
         total_latency = 0
         node_retry_counts: dict[str, int] = {}  # Track retries per node
         node_visit_counts: dict[str, int] = {}  # Track visits for feedback loops
+        node_latencies: dict[str, int] = {}  # Per-node latency breakdown
         _is_retry = False  # True when looping back for a retry (not a new visit)
 
         # Restore node_visit_counts from session state if available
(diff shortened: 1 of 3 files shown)


utk2602:
yeaaa gotcha , wait let me just push the changes accordingly

utk2602:
Acknowledged — changing `NodeResult.latency_ms` from `int` to `Optional[int]` would be a cleaner contract, but it's a breaking change across all NodeResult consumers. For this PR, 0ms is a safe sentinel since sub-millisecond nodes are effectively instant. Happy to do the `Optional[int]` migration as a separate PR if desired.

utk2602:
Fixed — `total_latency` accumulation is now inside the `finally` block, so it captures elapsed time even when `execute()` raises. The old `total_latency += result.latency_ms` after the try/finally was removed to avoid double-counting.

Re: parallel branch retries — the fan-out retry loop already sets `result.latency_ms` in the `finally` block per attempt. The aggregation into `node_latencies` at the fan-out merge point uses the final successful result's latency, which is consistent with how `total_tokens` is handled in the existing code. Accumulating every retry attempt's latency is a broader design change to `_execute_parallel_branches` that I'd prefer to handle in a follow-up.

PR #6970, opened 2026-04-07
fix(runtime): prevent dual execution on session after forced cancel timeout

akramcodez:
## Description
Fixes a race condition where a force-cancelled execution that exceeded its cancel timeout
would allow a new execution to start on the same session while the previous task was still
running. This caused concurrent writes to shared session state, overlapping events, and
non-deterministic behavior.

## Type of Change

## Related Issues
Fixes #6967

## Changes Made
- Keep bookkeeping (`_active_executions`, `_execution_tasks`) in place after cancel timeout
  instead of eagerly removing it; mark execution status as `"cancelling"` so the stream
  remains locked until the task actually terminates
- Add a `still_running` guard inside `self._lock` in `execute()` that raises
  `ExecutionAlreadyRunningError` if any task is still alive, preventing concurrent starts
- Move `create_task` inside the same lock as the guard check and bookkeeping writes so
  concurrent `execute()` calls cannot interleave between check and registration
- Register a `done_callback` on the stuck task via `_schedule_deferred_cleanup` to clear
  all per-execution state (`_active_executions`, `_execution_tasks`, `_completion_events`,
  `_active_executors`) once the task eventually terminates
- Add `_CANCEL_GRACE_SECONDS = 0.25` class constant and a grace window `asyncio.wait`
  before the lock check, giving cooperative tasks a short window to finish cleanly before
  raising
- Add regression test covering the full lifecycle: stream stays locked during timeout,
  `status == "cancelling"` is set, and all registries are cleared after task exits

## Testing

**Regression test added:** `tests/runtime/test_forced_cancel_regression.py`

The test simulates a non-cooperative task that ignores `CancelledError`, monkeypatches
`asyncio.wait` to always time out, and asserts:
1. `ExecutionAlreadyRunningError` is raised when a second execution attempts to start
2. The stuck task and its bookkeeping remain in place during the timeout
3. All three registries are fully cleared once the stuck task terminates

## Checklist

## Release Notes

* **Bug Fixes**
  * Improved execution stream cancellation handling to gracefully manage stuck tasks without forcing premature cleanup.
  * Enhanced concurrent execution protection to prevent race conditions when multiple executions are initiated simultaneously.
  * Refined timeout behavior during task cancellation to maintain stream integrity while awaiting task completion.

### DIFF
--- a/core/framework/runtime/execution_stream.py
+++ b/core/framework/runtime/execution_stream.py
@@ -164,6 +164,8 @@ class ExecutionStream:
         result = await stream.wait_for_completion(exec_id)
     """
 
+    _CANCEL_GRACE_SECONDS: float = 0.25
+
     def __init__(
         self,
         stream_id: str,
@@ -536,6 +538,14 @@ class ExecutionStream:
                         node.cancel_current_turn()
             await self.cancel_execution(eid, reason="Restarted with new execution")
 
+        # Give cancelling tasks a short grace window to finish before failing
+        # fast with ExecutionAlreadyRunningError.
+        running_tasks = [
+            task for task in self._execution_tasks.values() if task is not None and not task.done()
+        ]
+        if running_tasks:
+            await asyncio.wait(running_tasks, timeout=self._CANCEL_GRACE_SECONDS)
+
         # When resuming, reuse the original session ID so the execution
         # continues in the same session directory instead of creating a new one.
         resume_session_id = session_state.get("resume_session_id") if session_state else None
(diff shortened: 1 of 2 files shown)


iamsanjaii (commented):
This is a really solid fix especially the deferred cleanup approach, which aligns well with the regression test for non-cooperative cancellation.

I had a couple of thoughts while reviewing:

1. In `cancel_execution`, we call `task.cancel()` and wait with `asyncio.wait`, but we don’t explicitly `await task`. Would it be safer to await the task (handling `CancelledError`) to ensure full termination and avoid silent exceptions or background leaks?

2. Since cleanup happens both in `_run_execution` (finally block) and `_deferred_cleanup`, are we fully safe from race conditions between these two paths? It might be worth confirming that cleanup is strictly idempotent under all interleavings.

3. In `execute()`, after the grace wait, we check for `still_running` tasks — just wondering if there’s any edge case where a task finishes between the wait and the check that could allow a brief race condition.

Overall, really nice handling of a tricky concurrency issue, especially ensuring the stream remains locked until true termination.

PR #6999, opened 2026-04-08
fix: reduce OutputValidator false positives for natural language [micro-fix]

### LINKED ISSUE #6997: [Bug]: OutputValidator falsely flags natural language as code due to naive keyword matching
### Description
The OutputValidator overly aggressively detects conversational English phrases as "hallucinated code". Any sentence containing common programming keywords like "class", "import", "const", or "export" triggers the block mechanism, heavily penalizing models that are legitimately describing actions or discussing topics without writing code.
For example, a phrase like: *"Let me update the class schedule for next week"* or *"We need to import all data from the database"* currently triggers a hallucination flag due to simple substring matching for `"class "` or `"import "`.
### Expected Behavior
The OutputValidator should only flag actual strings of code. Natural language use of programming keywords should not prematurely block execution.
### Proposed Solution
Implement a two-tier indicator mechanism in `OutputValidator._contains_code_indicators()`.
- **Strong indicators** (such as `<script`, `try:`, `async def`) remain unchanged and flag immediately.
- **Weak indicators** (such as `import `, `class `, `var `) will now require a line-anchor (e.g. `\nimport `) and must co-occur at least twice within a string segment to validate a positive match for code hallucination.

### LINKED ISSUE #7000: OutputValidator._contains_code_indicators() produces false positives on natural language
## Description

The `_contains_code_indicators()` method in `core/framework/graph/validator.py` uses simple substring matching for keywords like `"class "`, `"import "`, `"const "`, etc. This causes false positives when conversational English naturally contains these words.

## Examples

- "Let me update the **class** schedule for next week" → flagged as code
- "We need to **import** more goods from overseas" → flagged as code
- "She will **return** the book tomorrow" → flagged as code
- "The **function** starts at 3pm" → flagged as code

## Suggested Fix

Implement a two-phase detection approach:
1. **Strong indicators** (e.g. `"def "`, `"func "`) → match directly
2. **Weak indicators** (e.g. `"import "`, `"class "`) → require line anchoring or code context characters

This preserves sensitivity to real code snippets while reducing false positives on natural language.

Bahtya:
Fixes #7000

## Problem

The `OutputValidator._contains_code_indicators()` method uses simple substring matching for keywords like `"class "`, `"import "`, `"const "`, etc. This causes false positives when conversational English contains these words:

- "Let me update the **class** schedule for next week" → flagged
- "We need to **import** all data from the database" → flagged
- "Please **select** your preferred option" → flagged

## Fix

Implement a two-tier indicator system:

- **Strong indicators** (`<script`, `try:`, `async def`, `<?php`, `<%`) flag immediately — these rarely appear in prose
- **Weak indicators** (`class `, `import `, `function `, SQL keywords, etc.) only flag when:
  - At the start of a line (`\n` prefix or position 0), OR
  - Preceded by a non-alpha character (not part of a word), OR
  - Followed by a code-context character (parentheses, braces, colons, etc.)

Added new test `test_no_false_positive_for_natural_language_keywords` to verify the fix.

Fixes #6997

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced false positives by refining code-like indicator detection with a clearer distinction between strong and weak markers and improved handling around line starts and contextual delimiters.
  * Improved accuracy for code patterns appearing after line breaks and in sampled/long inputs.

* **Tests**
  * Added/updated tests to ensure natural-language sentences no longer trigger code detection while real code snippets still do.

### DIFF
--- a/core/framework/graph/validator.py
+++ b/core/framework/graph/validator.py
@@ -84,7 +84,17 @@ class OutputValidator:
             "DROP ",
         ]
 
-        def _check_segment(segment: str) -> bool:
+        def _check_segment(segment: str, before_char: str = "") -> bool:
+            """Check a segment for code indicators.
+
+            Args:
+                segment: The text segment to check.
+                before_char: The character immediately preceding this segment in
+                    the original string (empty string if this segment starts at
+                    position 0 of the original).  Used so that idx==0 inside a
+                    sampled chunk is not incorrectly treated as a true line-start
+                    when the chunk was taken from the middle of a larger string.
+            """
             # Strong indicators: immediate match
             if any(indicator in segment for indicator in strong_indicators):
                 return True
@@ -98,7 +108,15 @@ class OutputValidator:
             for indicator in weak_indicators:
                 idx = segment.find(indicator)
                 while idx != -1:
-                    at_line_start = idx == 0 or segment[idx - 1] == "\n"
+                    # True line-start: either we're at position 0 of the
+                    # original string (before_char is empty) or the preceding
+                    # character — either inside the segment or the carried-in
+                    # before_char — is a newline.
+                    if idx == 0:
+                        prev_char = before_char
+                    else:
+                        prev_char = segment[idx - 1]
+                    at_line_start = prev_char == "" or prev_char == "\n"
                     if at_line_start:
                         return True
                     # Mid-line: check if code context follows the indicator
@@ -109,17 +127,19 @@ class OutputValidator:
                     # non-whitespace character (true punctuation / syntax context).
                     # Whitespace alone (e.g. "update the class schedule") is NOT
                     # treated as code context to avoid false positives.
-                    before_char = segment[idx - 1] if idx > 0 else ""
-                    if before_char and not before_char.isalpha() and not before_char.isspace():
+                    if prev_char and not prev_char.isalpha() and not prev_char.isspace():
                         return True
                     idx = segment.find(indicator, idx + 1)
             return False
 
-        # For strings under 10KB, check the entire content
+        # For strings under 10KB, check the entire content (before_char="" since
+        # we're checking from position 0 of the original string).
         if len(value) < 10000:
             return _check_segment(value)
 
-        # For longer strings, sample at strategic positions
+        # For longer strings, sample at strategic positions.  Pass the character
+        # immediately before each chunk so _check_segment can correctly determine
+        # whether the very first character of the chunk is a true line-start.
         sample_positions = [
             0,  # Start
             len(value) // 4,  # 25%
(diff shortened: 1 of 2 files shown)


Bahtya:
Note: I'm not a repo member so I can't self-assign to the linked issue. Marking this as `micro-fix` if allowed, or requesting maintainer to assign me to #7000.

Bahtya:
Fixed all CodeRabbit review issues:

**Critical fix** (validator.py:110): The `before_char` check was treating whitespace as code context, causing `"update the class schedule"` to still match `"class "`. Fixed by adding `and not before_char.isspace()` — now only true punctuation/syntax characters (e.g. `(`, `{`, `=`) count as code context for weak indicators mid-line.

**Minor fixes**:
- Docstring updated to accurately describe the implemented logic (no occurrence counting; conditions are: line-start OR punctuation-prefix OR code-suffix)
- Added `-> None` return type annotation to `test_no_false_positive_for_natural_language_keywords`

Local test run: `pytest core/tests/test_hallucination_detection.py` → **20 passed** ✅

Regarding the PR requirement: I'm not a repo member so I can't self-assign to #7000. Could a maintainer assign me to the issue, or add the `micro-fix` label to bypass the requirement?

Bahtya:
Fixed the last CodeRabbit review issue (chunk boundary false positives):

**Problem**: When sampling long strings (>10 KB), each 2000-char chunk is passed to `_check_segment`. Inside the chunk, `idx == 0` was being treated as a line-start — but the chunk may have been sliced from the middle of the original string, so `idx == 0` does not mean the indicator is actually at the start of a line.

**Fix**: Added a `before_char` parameter to `_check_segment` (defaults to `""`). For sampled chunks, we now pass `value[pos - 1]` as `before_char`. The line-start check becomes:
```python
prev_char = before_char if idx == 0 else segment[idx - 1]
at_line_start = prev_char == "" or prev_char == "\n"
```

This ensures:
- A chunk at `pos == 0` (start of string): `before_char = ""` → `idx == 0` is a true line-start ✅  
- A chunk at `pos > 0` mid-string: `before_char = value[pos-1]` → `idx == 0` is only a line-start if the preceding char in the original was `\n` ✅

All existing tests continue to pass locally.

Bahtya:
Hi maintainers — all CodeRabbit review comments have been addressed (latest check shows ✅ Review completed, 5 passed pre-merge checks). Could a maintainer either:

1. Assign me (@Bahtya) to issue #7000 (or #6997) so the PR requirements check passes, **or**
2. Add the `micro-fix` label to this PR to bypass the requirement

The only remaining blocker is the contributor requirement check — all CI is green, no open review issues. Thanks!

PR #7001, opened 2026-04-08
fix(runtime): prevent session restart until cancelled execution fully terminates

kowshik-04:
## Summary

Fixes a race condition in ExecutionStream where a force-cancelled execution could time out during cancellation, have its bookkeeping removed early, and allow a new execution to start on the same session while the previous task was still running.

This change ensures the session remains blocked until the original task has fully terminated.

---

## Type of Change

Bug fix (non-breaking change that fixes an issue)

---

## Related Issues

Closes #6967

---

## Changes Made

Keep execution bookkeeping on cancel timeout instead of removing it early  
Mark executions as cancelling while shutdown is in progress  
Treat any still-alive execution task as blocking for new executions  
Make execution registration fully atomic under the stream lock  
Return 409 Conflict from trigger/resume/replay routes when execution is still shutting down  
Add regression tests covering forced-cancel restart scenarios  

---

## Testing

Focused regression tests pass:

.venv/bin/pytest core/framework/runtime/tests/test_forced_cancel_dual_execution.py -q

Modified files also compile cleanly.

The following scenarios are covered:

forced cancel timeout keeps the stream blocked until task exit  
repeated forced restarts do not allow overlapping executions  

Note: Full aiohttp server tests could not be executed in this environment due to socket binding restrictions, but the runtime behavior relevant to this fix was validated by focused regression tests.

---

## Checklist

My code follows the project's style guidelines  
I have performed a self-review of my code  
I have added tests that prove my fix is effective  
Relevant focused tests pass locally  

## Summary by CodeRabbit

* **New Features**
  * Cancellation is tri-state ("cancelled", "cancelling", "not_found"); pause/stop responses include both cancelled and cancelling lists. Trigger/resume/replay now return HTTP 409 when an execution conflict exists.

* **Bug Fixes**
  * Strengthened locking and execution tracking to prevent duplicate/accumulating execution tasks, preserve "cancelling" state until tasks exit, and avoid premature cleanup.

* **Tests**
  * Added regression tests for forced-cancel/restart overlap and updated API tests for new cancellation semantics.

### DIFF
--- a/core/framework/host/agent_host.py
+++ b/core/framework/host/agent_host.py
@@ -1672,7 +1672,7 @@ class AgentHost:
         entry_point_id: str,
         execution_id: str,
         graph_id: str | None = None,
-    ) -> bool:
+    ) -> str:
         """
         Cancel a running execution.
 
@@ -1682,11 +1682,11 @@ class AgentHost:
             graph_id: Graph to search (defaults to active graph)
 
         Returns:
-            True if cancelled, False if not found
+            Cancellation outcome from the stream.
         """
         stream = self._resolve_stream(entry_point_id, graph_id)
         if stream is None:
-            return False
+            return "not_found"
         return await stream.cancel_execution(execution_id)
 
     # === QUERY OPERATIONS ===
--- a/core/framework/host/execution_manager.py
+++ b/core/framework/host/execution_manager.py
@@ -1209,9 +1209,7 @@ class ExecutionManager:
         """Get execution context."""
         return self._active_executions.get(execution_id)
 
-    async def cancel_execution(
-        self, execution_id: str, *, reason: str | None = None
-    ) -> CancelExecutionResult:
+    async def cancel_execution(self, execution_id: str, *, reason: str | None = None) -> CancelExecutionResult:
         """
         Cancel a running execution.
 
@@ -1249,8 +1247,7 @@ class ExecutionManager:
             # here because resumed executions reuse the same id; a delayed pop
             # could otherwise delete bookkeeping that belongs to the new run.
             logger.warning(
-                "Execution %s did not finish within cancel timeout; "
-                "leaving bookkeeping in place until task exit",
+                "Execution %s did not finish within cancel timeout; leaving bookkeeping in place until task exit",
                 execution_id,
             )
             return "cancelling"
--- a/core/framework/host/execution_manager.py
+++ b/core/framework/host/execution_manager.py
@@ -16,7 +16,7 @@ from collections import OrderedDict
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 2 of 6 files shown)


Hundao (commented):
Diagnosis is right and the fix is the right shape. Tri-state return + `Literal`, 409 in the routes, `_get_blocking_execution_ids_locked` catching the "task still alive but missing from active_executions" gap, `create_task` moved inside the lock, the `stop()` rewrite. All the right calls. The tests with a real `BlockingExecutor` mock are also way better than monkeypatching `asyncio.wait` against a `SimpleNamespace`.

One thing I'd want before merging: a comment on `cancel_execution` (and probably `_get_blocking_execution_ids_locked` too) explaining *why* the stream stays locked until the task actually exits. If the task is truly stuck it can still mutate session state, so silently letting a replacement start would guarantee corruption. The permanent lock is doing real work, it's not an oversight.

The reason I want this written down: the competing PR #6970 added a `done_callback` based "deferred cleanup" that looks like a safety net but actually re-introduces the race, because `resume_session_id` makes the new execution reuse the same `execution_id`, and the deferred pop ends up running after the new execution has registered its bookkeeping. Without a comment explaining the intent, the next person looking at this is going to be tempted to add the same thing.

Minor / non-blocking: it might be worth emitting a periodic warning when a stream sits in `cancelling` for >N seconds, so we can spot a leaked task before someone notices the stream is wedged. Follow-up is fine.

Also when you push the comment update, please run `ruff check` and `ruff format` over the changed files so CI passes cleanly.

Other than that, looks ready to land.

kowshik-04:
@Hundao 
Added the intent comments around `_get_blocking_execution_ids_locked()` and `cancel_execution()` to make the locking behavior explicit and document why deferred cleanup would reintroduce the race with reused `execution_id`s.

I also ran `ruff format` / `ruff check` on the changed Python files and re-ran the focused regression tests before pushing the update.

kowshik-04:
@Hundao 
I’ve pushed the latest updates:

- added the intent comments around `_get_blocking_execution_ids_locked()` and `cancel_execution()`
- added the missing `meta.json` phase update for the `successful stop_graph_and_plan()` transition
- ran `ruff format` / `ruff check` on the changed Python files
- re-ran the focused regression tests

Could you please take another look when you have a moment?

kowshik-04:
Rebased onto the latest upstream/main and resolved the merge conflicts.

Preserved the execution lifecycle fixes, aligned the changes with recent upstream updates, and re-ran the focused forced-cancel regression tests locally.

Please let me know if anything else needs adjustment.

kowshik-04:
Rebased onto the latest upstream/main and resolved the remaining conflict in `routes_execution.py.`

Preserved the tri-state cancellation semantics and aligned with the updated upstream structure. 
Re-ran lint and focused regression tests locally.

PR #7060, opened 2026-04-16
Docs: clarify allowed-tools field with examples and behavior explanation

kevinbenty:
## What I did
Added a dedicated section explaining the `allowed-tools` field.

## Why
The field was referenced but not properly documented, causing confusion about:
- What qualifies as a tool
- What "pre-approved" means
- How it affects execution and permissions

## Changes
- Added definition of `allowed-tools`
- Included usage example
- Clarified behavior of tools not listed
- Explained security implications

## Impact
Improves clarity and helps prevent misuse of a potentially security-relevant configuration field.

## Summary by CodeRabbit

* **Documentation**
  * Added an "allowed-tools" section to the skills guide: defines what counts as a tool, includes a YAML example, explains that the platform parses/validates the field and may emit warnings (e.g., malformed values or missing tools), and clarifies the field documents expected/pre-approved tools for transparency and policy support rather than enforcing runtime allow/deny behavior.

### DIFF
--- a/docs/skills-user-guide.md
+++ b/docs/skills-user-guide.md
@@ -116,6 +116,41 @@ Use this skill when the user needs to extract text from PDFs or merge documents.
 | `metadata` | No | Arbitrary key-value pairs (author, version, etc.). |
 | `allowed-tools` | No | Space-delimited list of pre-approved tools. |
 
+### allowed-tools
+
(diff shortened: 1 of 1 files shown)


kevinbenty:
Thanks for the feedback! I've updated the documentation to reflect the current behavior and removed the implication of runtime enforcement.

PR #7097, opened 2026-04-20
fix(credentials): raise CredentialError when Aden sync init fails with API key set

kostasuser01gr:
## Summary

`CredentialStore.with_aden_sync()` silently falls back to local storage when `ADEN_API_KEY` is configured but initialization fails (connection errors, auth failures, etc.). This masks real configuration problems — users think they have Aden sync but are actually running local-only.

## Changes

**`core/framework/credentials/store.py`**  
- The broad `except Exception` handler now raises `CredentialError` with a descriptive message when `ADEN_API_KEY` is present but the sync provider cannot start.
- `except ImportError` fallback preserved (graceful degradation when Aden package is not installed).
- Log level changed from `warning` to `error` for the failure case.

**`core/tests/test_aden_sync_error.py`** (new)  
- `test_raises_when_api_key_set_and_init_fails` — client init raises → CredentialError
- `test_raises_when_api_key_set_and_sync_fails` — sync_all raises → CredentialError
- `test_falls_back_when_no_api_key` — no key → local storage (unchanged behavior)
- `test_falls_back_when_aden_not_installed` — ImportError → local storage (unchanged behavior)

## Validation

- `make check` — ruff lint + format: clean
- Core tests: 1603 passed, 0 failed
- Pre-commit hooks: passed

Fixes #5859

## Summary by CodeRabbit

* **Bug Fixes**
  * Credential sync now logs and raises a clear credential error when external Aden sync is configured but fails to initialize or run; fallback remains when the Aden API key is not set or Aden is unavailable.

* **Tests**
  * Added tests to verify Aden sync failure handling and fallback scenarios.

* **Style**
  * Minor formatting tidy-ups in internal agent loop code (no behavioral changes).

### DIFF
--- a/core/framework/agent_loop/internals/cursor_persistence.py
+++ b/core/framework/agent_loop/internals/cursor_persistence.py
@@ -162,9 +162,7 @@ async def drain_injection_queue(
     conversation: NodeConversation,
     *,
     ctx: NodeContext,
-    caption_image_fn: (
-        Callable[[str, list[dict[str, Any]]], Awaitable[tuple[str, str] | None]] | None
-    ) = None,
+    caption_image_fn: (Callable[[str, list[dict[str, Any]]], Awaitable[tuple[str, str] | None]] | None) = None,
 ) -> int:
     """Drain all pending injected events as user messages. Returns count.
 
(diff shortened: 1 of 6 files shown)


kostasuser01gr:
Rebased cleanly onto `main` at v0.10.5 (`ea707438`). Single commit, no conflicts. Tests in `core/tests/test_aden_sync_error.py` (4/4) green locally on the rebased tree.

Happy to wait for issue assignment per the policy before this is reviewed; flagging only that the branch is now in a mergeable state.

PR #7122, opened 2026-04-22
Fix: standardize needs_more_research condition expressions in agent.py

reuelll:
Description:

## What was the problem?
condition_expr in agent.py used:
    needs_more_research == True

But agent.json used the safer version:
    str(needs_more_research).lower() == 'true'

This mismatch could break the feedback loop if the value 
comes back as a string instead of a boolean.

## What I changed
- agent.py: updated both condition_expr strings to match agent.json
- config.py: improved intro message and bumped version to 1.1.0
- agent.json: updated source target from >=5 to >=8

## Testing
- Agent structure validated
- condition_expr now consistent across agent.py and agent.json

## Summary by CodeRabbit

* **Updates**
  * Version updated to 1.1.0
  * Enhanced onboarding message with improved workflow guidance, including clarification on HTML output format and research process expectations
  * Improved research routing decision logic for better handling of research continuation decisions

### DIFF
--- a/examples/templates/deep_research_agent/agent.py
+++ b/examples/templates/deep_research_agent/agent.py
@@ -106,23 +106,23 @@ edges = [
         priority=1,
     ),
     # review -> research (feedback loop)
-    EdgeSpec(
-        id="review-to-research-feedback",
-        source="review",
-        target="research",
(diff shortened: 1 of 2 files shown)


reuelll:
I have identified this bug and submitted a PR with a fix.

PR #7139, opened 2026-04-26
fix(dnssec): Implement custom resolver for reliable AD flag validation (micro-fix)

omichauhan-lgtm:
Fixes #5112.

The \_check_dnssec\ tool was returning false positives because the system resolver doesn't always validate DNSSEC, or strips the AD flag. This PR ensures we:
1. Initialize an isolated \dns.resolver.Resolver()\
2. Target public resolvers known to perform validation (8.8.8.8, 1.1.1.1)
3. Request validation using EDNS DO bit
4. Check the AD flag on the returned SOA records

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced DNSSEC validation in the DNS security scanner. The scanner now confirms DNSSEC enablement only when cryptographic validation succeeds, provides clearer "enabled"/"disabled" reporting, and gives more informative issue messages when DNSSEC is absent or cannot be validated. Improvements reduce false positives and improve the accuracy of validation status shown to users.

### DIFF
--- a/tools/src/aden_tools/tools/dns_security_scanner/dns_security_scanner.py
+++ b/tools/src/aden_tools/tools/dns_security_scanner/dns_security_scanner.py
@@ -11,6 +11,7 @@ from fastmcp import FastMCP
 
 try:
     import dns.exception
+    import dns.flags
     import dns.name
     import dns.query
     import dns.rdatatype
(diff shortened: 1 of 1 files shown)


omichauhan-lgtm:
Thanks for the review @coderabbitai! I've removed the unused 
esolver argument from _check_dnssec and merged the exception blocks as suggested.

PR #7162, opened 2026-05-03
fix: TOCTOU race condition in fan-out write conflict check (#7125)

### LINKED ISSUE #7125: [Bug]:TOCTOU race condition in fan-out write conflict check (_fanout_written_keys)
## Bug: Race condition in `_write_outputs()` fan-out conflict check

**File:** `hive/core/framework/orchestrator/node_worker.py`  
**Lines:** 556–580

---

### What's wrong

The check-then-write on `gc._fanout_written_keys` is not atomic. 
Two concurrent fan-out workers can both pass the conflict check before 
either records its write, silently defeating all three conflict strategies.

```python
# Lines 556–580 (current — unsafe)
prior_worker = gc._fanout_written_keys.get(key)   # CHECK
if prior_worker and prior_worker != node_spec.id:
    if conflict_strategy == "error":
        raise RuntimeError(...)          # never fires if two workers race here
    elif conflict_strategy == "first_wins":
        continue                         # both workers "win" simultaneously
gc._fanout_written_keys[key] = node_spec.id        # WRITE
gc.buffer.write(key, value, validate=False)
```

---

### Why this is clearly a bug

The same file already uses locks consistently for other shared state:
```python
# Line 251 — node_visit_counts is protected
async with gc._visits_lock:
    visit_count = gc.node_visit_counts.get(node_spec.id, 0) + 1
    gc.node_visit_counts[node_spec.id] = visit_count

# Line 299 — path is protected  
async with gc._path_lock:
    gc.path.append(node_spec.id)
```

---

### Proposed fix

1. Add _fanout_lock: asyncio.Lock to GraphContext in context.py (alongside _path_lock and _visits_lock)
2. Wrap the check-and-write block:
```python
async with gc._fanout_lock:
    prior_worker = gc._fanout_written_keys.get(key)
    if prior_worker and prior_worker != node_spec.id:
        if conflict_strategy == "error":
            raise RuntimeError(...)
        elif conflict_strategy == "first_wins":
            continue
        else:
            logger.debug(...)
    gc._fanout_written_keys[key] = node_spec.id
gc.buffer.write(key, value, validate=False)  # buffer has its own lock
```

---

cc: @RichardTang-Aden - flagging since this touches the concurrency pattern you established in this file.

I'd like to take this if you can assign it to me.

dong888cs:
## Summary

Fixes #7125 — a **TOCTOU (Time-of-Check-Time-of-Use)** race condition in `_write_outputs()` where two concurrent fan-out workers could both pass the `_fanout_written_keys` conflict check before either recorded its write, silently defeating all three conflict strategies (`error`, `first_wins`, `last_wins`).

## Root Cause

The check-then-write on `gc._fanout_written_keys` was not atomic:

```python
prior_worker = gc._fanout_written_keys.get(key)   # CHECK
# ... race window: another worker writes the same key here ...
gc._fanout_written_keys[key] = node_spec.id        # WRITE
```

## Fix

1. **context.py**: Added `_fanout_lock: asyncio.Lock` to `GraphContext` (alongside the existing `_path_lock` and `_visits_lock`).
2. **node_worker.py**: Made `_write_outputs` async and wrapped the check-and-write block with `async with gc._fanout_lock:`.

The buffer write itself (`gc.buffer.write`) is left *outside* the lock since the buffer has its own internal locking — only the conflict-tracking metadata access needs serialization.

## Consistency

This follows the same locking pattern already established in this file:
- `_visits_lock` protects `gc.node_visit_counts` (line 251)
- `_path_lock` protects `gc.path` (line 299)
- `_fanout_lock` now protects `gc._fanout_written_keys`

## Testing

- [x] All three conflict strategies (`error`, `first_wins`, `last_wins`) are correctly serialized under the lock
- [x] Existing lock patterns in the same class are followed consistently
- [x] `continue` inside `async with` block correctly skips the buffer write while holding the lock
- [x] Non-fan-out writes are unaffected (no lock overhead)

Co-Authored-By: Claude Opus 4.7 &lt;noreply@anthropic.com&gt;

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronized concurrent fan-out buffer writes with a lock and ensured deterministic conflict strategies (first_wins, last_wins, error), with a consistent default to avoid race conditions.
  * Output writes are now awaited to ensure proper ordering and conflict detection across workers.

* **Tests**
  * Added async concurrency tests covering contention scenarios, non-fan-out behavior, repeated-writer cases, multi-key writes, and default strategy behavior.

### DIFF
--- a/core/framework/orchestrator/context.py
+++ b/core/framework/orchestrator/context.py
@@ -81,6 +81,7 @@ class GraphContext:
     _visits_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
     # Fan-out buffer conflict tracking: key → worker_id that wrote it
     _fanout_written_keys: dict[str, str] = field(default_factory=dict)
+    _fanout_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
     # Retry tracking: worker_id → retry_count (for execution quality assessment)
     retry_counts: dict[str, int] = field(default_factory=dict)
     nodes_with_retries: set[str] = field(default_factory=set)
--- a/core/framework/orchestrator/node_worker.py
+++ b/core/framework/orchestrator/node_worker.py
@@ -311,7 +311,7 @@ class NodeWorker:
             # Handle result
             if result.success:
                 # Validate and write outputs
-                self._write_outputs(result)
+                await self._write_outputs(result)
 
                 # Evaluate outgoing edges
                 activations = await self._evaluate_outgoing_edges(result)
@@ -517,7 +517,7 @@ class NodeWorker:
     # Output handling
     # ------------------------------------------------------------------
 
-    def _write_outputs(self, result: NodeResult) -> None:
+    async def _write_outputs(self, result: NodeResult) -> None:
         """Validate and write node outputs to buffer."""
         gc = self._gc
         node_spec = self.node_spec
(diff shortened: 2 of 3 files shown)


dong888cs:
Hi @RichardTang-Aden, could you please assign me to issue #7125? The bot requires the PR author to be assigned to the linked issue, but I don't have the necessary permissions on the upstream repo to do it myself. Thanks!

PR #7189, opened 2026-05-14
fix(config): handle null max_tokens in configuration.json

### LINKED ISSUE #7117: [Bug]: null llm.max_tokens in configuration.json causes queen loop crash on startup
## Bug Report

**Describe the bug**
If llm.max_tokens or llm.max_context_tokens are set to null in 
configuration.json, the queen loop crashes immediately on startup with:
  TypeError: '>' not supported between instances of 'NoneType' and 'int'

**To Reproduce**
1. Set llm.max_tokens: null in configuration.json
2. Run hive open
3. Send any message

**Expected behavior**
Should fall back to default token values instead of crashing.

**Environment**
- OS: Windows 11
- Provider: Gemini

enaktes9-hub:
## Problem

When `llm.max_tokens` or `llm.max_context_tokens` is explicitly set to `null` in `configuration.json`, Python's `dict.get(key, default)` returns `None` instead of the default value. This is because `.get()` only falls back when the key is **missing**, not when it's `null`.

This causes two crash paths:
1. `get_max_tokens()` and `get_max_context_tokens()` return `None` instead of defaults
2. `publish_context_usage()` crashes with `TypeError: '>' not supported between instances of 'NoneType' and 'int'`

## Fix

**`core/framework/config.py`**:
- `get_max_tokens()`: Extract value, fall back to `DEFAULT_MAX_TOKENS` if `None`
- `get_max_context_tokens()`: Same pattern with `DEFAULT_MAX_CONTEXT_TOKENS`
- `get_worker_max_tokens()` / `get_worker_max_context_tokens()`: Use `is not None` check

**`core/framework/agent_loop/internals/event_publishing.py`**:
- `publish_context_usage()`: Add `or 32000` fallback for null `_max_context_tokens`

Closes #7117

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of default token limits when not explicitly configured, ensuring more consistent fallback behavior.
  * Enhanced context usage metrics calculation for more accurate reporting of token consumption and usage ratios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/aden-hive/hive/pull/7189)

### DIFF
--- a/core/framework/agent_loop/internals/event_publishing.py
+++ b/core/framework/agent_loop/internals/event_publishing.py
@@ -184,7 +184,7 @@ async def publish_context_usage(
     from framework.host.event_bus import AgentEvent, EventType
 
     estimated = conversation.estimate_tokens()
-    max_tokens = conversation._max_context_tokens
+    max_tokens = conversation._max_context_tokens or 32000
     ratio = estimated / max_tokens if max_tokens > 0 else 0.0
     await event_bus.publish(
         AgentEvent(
--- a/core/framework/config.py
+++ b/core/framework/config.py
@@ -322,7 +322,7 @@ def get_worker_llm_extra_kwargs() -> dict[str, Any]:
 def get_worker_max_tokens() -> int:
     """Return max_tokens for the worker LLM, falling back to default."""
     worker_llm = get_hive_config().get("worker_llm", {})
-    if worker_llm and "max_tokens" in worker_llm:
+    if worker_llm and worker_llm.get("max_tokens") is not None:
         return worker_llm["max_tokens"]
     return get_max_tokens()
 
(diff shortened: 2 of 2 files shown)


kowshik-04 (commented):
Nice fix overall especially the `is not None` checks in the config helpers. That correctly handles the `null` case while still preserving valid falsy values.

One small production-level suggestion:

```python id="k2d9x1"
max_tokens = conversation._max_context_tokens or 32000
```

I think this fallback may be better handled entirely at the config layer instead of inside `event_publishing.py`.

Right now `get_max_tokens()` and `get_max_context_tokens()` already normalize `None` values correctly, so adding another hardcoded fallback here creates an additional source of truth for the default context size.

If the default changes later, this runtime fallback could silently drift from the actual config defaults.

A cleaner approach could be:

* keep all normalization inside the config helpers
* guarantee those helpers never return `None`
* let downstream runtime code trust that contract

That keeps the fallback behavior centralized and makes future maintenance/refactors easier.

Other than that, the fix direction looks solid and addresses the original crash path well.

PR #7234, opened 2026-06-01
Fix/6391 use multi sse session reconnect

### LINKED ISSUE #6391: [Bug]: useMultiSSE does not reconnect when session ID changes
## Describe the Bug

The `useMultiSSE` hook in `core/frontend/src/hooks/use-sse.ts` does not reconnect when the session ID changes for an already connected agent type.

When a session is restarted or recreated, the frontend continues to hold an `EventSource` connection to the old session. As a result, the UI stops receiving real-time events (streaming responses, typing indicators, node updates, etc.) even though the backend agent continues running.

---

## To Reproduce

1. Open the Hive workspace
2. Start a conversation with an agent
3. Restart or recreate the agent session
4. Send another message

---

## Expected Behavior

The frontend should close the old SSE connection and reconnect to the new session automatically so real-time events continue to flow.

---

## Actual Behavior

The UI stops receiving SSE events:

- No streaming response
- No typing indicator
- No tool call updates

Refreshing the page restores functionality because all SSE connections are recreated.

---

## Environment

- OS: macOS
- Browser: Chrome
- Hive repository: latest `main`

---

## Root Cause

In `useMultiSSE`, the connection diff logic only checks if the agent type already exists in the connection map:
`if (!sessionId || current.has(agentType)) continue;`

This only checks if the **agent type key exists**, not whether the **session ID changed**.

Example scenario:

Old state
`researcher-agent → session_abc123`

New state
`researcher-agent → session_xyz789`

Since `current.has("researcher-agent")` is `true`, the hook skips creating a new `EventSource`.

This causes:

- The old `EventSource` to remain connected to `session_abc123`
- No connection to be opened for `session_xyz789`

---

## Suggested Fix

Track the session ID alongside the `EventSource` instance so session changes can be detected.

Example approach:
`Map<agentType, { es: EventSource, sessionId: string }>`

Then reconnect when the session ID changes:

- If the `agentType` exists but the `sessionId` differs:
  - close the old `EventSource`
  - create a new `EventSource` for the new session

---

## Additional Context

This bug causes the frontend to lose real-time updates after session restarts. The backend agent continues processing, but the UI becomes unresponsive to streaming events until the page is refreshed.

swikarthanekar:
Fixes #6391

## Summary
Reconnect `useMultiSSE` when the session ID changes for the same agent type so SSE events continue after session restart without a page refresh.

## Changes
- Add `multi-sse-diff.ts` with explicit close/open rules
- Refactor `useMultiSSE` to use those helpers (removes debug console logging)
- Add Vitest tests in `multi-sse-diff.test.ts`

## Testing
- [x] `cd core/frontend && npm test`
- [x] `cd core/frontend && npm run build`
- [ ] Manual: restart session in queen DM / colony chat, confirm streaming works

## Summary by CodeRabbit

* **Refactor**
  * Improved SSE connection management for more reliable open/close and reconnect behavior; reduced noisy logging and simplified per-connection event handling.

* **Tests**
  * Added comprehensive tests covering SSE connection lifecycle, session changes, reconnect flows, and edge cases to ensure stable behavior.

### DIFF
--- a/core/frontend/src/hooks/multi-sse-diff.test.ts
+++ b/core/frontend/src/hooks/multi-sse-diff.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from "vitest";
+import { agentTypesToClose, connectionsToOpen } from "./multi-sse-diff";
+
+describe("multi-sse-diff", () => {
+  it("closes when agent type is removed from sessions", () => {
+    const current = new Map([["queen-dm", { sessionId: "s1" }]]);
+    expect(agentTypesToClose(current, {})).toEqual(["queen-dm"]);
+  });
+
+  it("closes when session ID changes for the same agent type", () => {
+    const current = new Map([["researcher-agent", { sessionId: "session_abc123" }]]);
+    const sessions = { "researcher-agent": "session_xyz789" };
+    expect(agentTypesToClose(current, sessions)).toEqual(["researcher-agent"]);
+  });
+
+  it("does not close when agent type and session ID are unchanged", () => {
+    const current = new Map([["queen-dm", { sessionId: "s1" }]]);
+    expect(agentTypesToClose(current, { "queen-dm": "s1" })).toEqual([]);
+  });
+
+  it("opens a new connection when agent type is new", () => {
+    expect(connectionsToOpen(new Map(), { "queen-dm": "s1" })).toEqual([
+      { agentType: "queen-dm", sessionId: "s1" },
+    ]);
+  });
+
+  it("opens when session ID changes after stale entry would be closed", () => {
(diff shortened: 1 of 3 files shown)


kowshik-04 (approved):
Thanks for working on this. The changes align with the root cause described in #6391.

I verified that the new diff logic now considers session ID changes in addition to agent type presence, which was the missing behavior causing stale SSE connections after session restarts.

The added test coverage for:

* removed agent types
* session ID changes
* unchanged sessions
* reconnect flows

looks good and directly validates the reported bug scenario.

I don't see any functional concerns with the removal of the previous `onopen`/`onerror` console logging since those handlers were only used for debugging and were not part of the reconnect logic itself.

PR #7265, opened 2026-06-20
fix(llm): prevent mid-stream retry from duplicating published tokens

### LINKED ISSUE #5923: [Bug]: LiteLLMProvider.stream() retries mid-stream without guard — clients receive duplicated partial response concatenated with full retry response

EsraaKamel11:
## Description

Fixes #5923. When the LiteLLM streaming layer retried after a mid-stream `RateLimitError` or transient error, it re-streamed from token 1, duplicating content that had already been yielded to callers and published to the event bus. Because published `TextDeltaEvent`s cannot be recalled, the retry must be abandoned once a partial stream has already been emitted.

**Root cause.** `LiteLLMProvider.stream()` (`core/framework/llm/litellm.py`) runs an internal retry loop that resets `accumulated_text` at the top of each attempt and yields `TextDeltaEvent`s in real time. Both the `RateLimitError` handler (`L2694`) and the transient `except Exception` handler (`L2759`) called `continue` unconditionally — restarting the stream from the beginning even when text had already been published.

```python
# Before — restarts the stream, re-publishing every prior token:
except RateLimitError as e:
    if attempt < RATE_LIMIT_MAX_RETRIES:
        wait = _compute_retry_delay(attempt, exception=e)
        await asyncio.sleep(wait)
        continue
```

**Fix.** In both handlers, check `accumulated_text` before retrying. If any text was already yielded, emit a `recoverable=True` `StreamErrorEvent` and return instead of retrying.

```python
except RateLimitError as e:
    if attempt < RATE_LIMIT_MAX_RETRIES:
        if accumulated_text:               # L2705
            yield StreamErrorEvent(error=str(e), recoverable=True)
            return
        wait = _compute_retry_delay(attempt, exception=e)
        ...
        continue
```

The cooperating consumer already exists in `AgentLoop`: a `recoverable=True` error is not raised (`agent_loop.py:2634`), and `ConnectionError` is raised only when the partial response is empty (`agent_loop.py:2857`). So a non-empty `accumulated_text` commits the partial turn and skips the outer retry, while an empty one falls through to the outer transient retry (`agent_loop.py:1089`). The guard keys on `accumulated_text` only — tool-call deltas are buffered and never published before stream completion, so tool-only streams remain safe to retry internally.

> Supersedes #6541, whose branch predated the graph→orchestrator revamp; this is the same fix reworked on current `main` (the consumer is now `AgentLoop`, not `EventLoopNode`).

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #5923

## Changes Made

- `core/framework/llm/litellm.py`: add an `accumulated_text` guard to the `RateLimitError` handler (L2694 → yields a recoverable error at L2705) and the transient `Exception` handler (L2759 → L2766) so a mid-stream failure after text was published stops retrying instead of re-streaming.
- `core/tests/test_stream_retry_duplication.py` (new): 5 provider unit tests covering rate-limit/transient after text (no duplication, recoverable error, single stream call), a 50-chunk no-duplication case, a tool-only stream that still retries internally, and the before-any-text path that still retries.
- `core/tests/test_event_loop_node.py` (new `TestMidStreamRetryNoDuplication`): 3 AgentLoop integration tests proving the end-to-end contract — a recoverable error after text commits the partial turn with no outer retry (`llm._call_index == 1`, each delta published once), while a recoverable error before any text triggers the outer retry (`llm._call_index == 2`).

## Testing

- [x] Unit tests pass (`cd core && pytest tests/test_stream_retry_duplication.py tests/test_event_loop_node.py::TestMidStreamRetryNoDuplication`) — 8 new tests pass; 145 existing `test_litellm_provider.py` tests pass.
- [x] Lint passes (`cd core && ruff check .` and `ruff format --check`) — clean on changed files.
- [x] Manual testing performed — verified via TDD: with the guards removed, the duplication tests fail (`call_count` becomes 2, deltas double); with the guards in place they pass.

Pure-Python change with no OS-specific code; developed/run on Windows (Python 3.11 via `uv`). macOS/Linux covered by CI.

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A (internal bug fix, no public API or user-facing behavior change)

## Release Notes

* **Bug Fixes**
  * Prevented duplicated streaming output when rate-limit or transient connection errors occur mid-stream. The stream now stops retrying after partial content has already been delivered, emits a recoverable error, and avoids restarting already-published text.

* **Tests**
  * Added regression coverage to ensure mid-stream retry behavior never duplicates published text, and that retries still occur when failures happen before any content is emitted.

### DIFF
--- a/core/framework/llm/litellm.py
+++ b/core/framework/llm/litellm.py
@@ -2693,6 +2693,17 @@ class LiteLLMProvider(LLMProvider):
 
             except RateLimitError as e:
                 if attempt < RATE_LIMIT_MAX_RETRIES:
+                    if accumulated_text:
+                        # Text deltas were already yielded to the caller (and
+                        # published to the event bus) in real time — they cannot
+                        # be recalled.  Retrying restarts the stream from token 1
+                        # (accumulated_text is reset per attempt above),
+                        # duplicating content the client already received.
+                        # Signal a recoverable error instead: the consumer
+                        # (AgentLoop) commits the partial text and skips the
+                        # outer retry when accumulated_text is non-empty.
+                        yield StreamErrorEvent(error=str(e), recoverable=True)
+                        return
                     wait = _compute_retry_delay(attempt, exception=e)
                     logger.warning(
                         f"[stream-retry] {self.model} rate limited (429): {e!s}. "
@@ -2746,6 +2757,14 @@ class LiteLLMProvider(LLMProvider):
                         yield event
                     return
                 if _is_stream_transient_error(e) and attempt < STREAM_TRANSIENT_MAX_RETRIES:
+                    if accumulated_text:
+                        # Same guard as RateLimitError: text already streamed to
+                        # the caller in real time cannot be recalled.  Stop
+                        # retrying to avoid client-visible duplication; signal a
+                        # recoverable error so the consumer commits the partial
+                        # text instead of re-streaming from token 1.
+                        yield StreamErrorEvent(error=str(e), recoverable=True)
+                        return
                     wait = _compute_retry_delay(attempt, exception=e)
                     logger.warning(
                         f"[stream-retry] {self.model} transient error "
--- a/core/tests/test_event_loop_node.py
+++ b/core/tests/test_event_loop_node.py
@@ -2257,3 +2257,90 @@ class TestReplayDetector:
         await node.execute(ctx)
 
         assert captured == []
+
+
+# ===========================================================================
+# Mid-stream retry must not duplicate already-published tokens (#5923)
+# ===========================================================================
+
+
+def _partial_then_recoverable_error(tokens: list[str], error: str = "429 mid-stream") -> list:
+    """A stream that yields text deltas then fails mid-stream with a recoverable
+    error and no FinishEvent — what LiteLLMProvider.stream now emits instead of
+    silently re-streaming after a mid-stream RateLimitError/transient error.
+    """
+    events: list = []
+    snapshot = ""
+    for tok in tokens:
+        snapshot += tok
+        events.append(TextDeltaEvent(content=tok, snapshot=snapshot))
+    events.append(StreamErrorEvent(error=error, recoverable=True))
+    return events
+
+
+class TestMidStreamRetryNoDuplication:
+    """End-to-end (AgentLoop + provider contract): a recoverable error that
+    arrives AFTER text was published commits the partial turn and does NOT re-run
+    the LLM; one that arrives BEFORE any text triggers the outer retry."""
+
+    @pytest.mark.asyncio
+    async def test_partial_stream_recoverable_error_no_outer_retry(self, runtime, node_spec, buffer):
+        node_spec.output_keys = []
+        llm = MockStreamingLLM(scenarios=[_partial_then_recoverable_error(["Hel", "lo ", "world"])])
+        bus = EventBus()
+        deltas: list = []
+        bus.subscribe(
+            event_types=[EventType.LLM_TEXT_DELTA],
+            handler=lambda e: deltas.append(e),
+        )
+
+        ctx = build_ctx(runtime, node_spec, buffer, llm, is_subagent_mode=True)
+        node = EventLoopNode(event_bus=bus, config=LoopConfig(max_iterations=5))
+        result = await node.execute(ctx)
+
+        # The stream ran exactly once — the partial turn was committed, not
+        # re-run (a re-run would re-publish every token).
+        assert llm._call_index == 1
+        # Each of the three tokens was published exactly once (no duplication).
+        assert len(deltas) == 3
+        assert result.success is True
+
+    @pytest.mark.asyncio
+    async def test_partial_stream_many_chunks_no_duplicate_deltas(self, runtime, node_spec, buffer):
+        node_spec.output_keys = []
+        tokens = [f"t{i} " for i in range(20)]
+        llm = MockStreamingLLM(scenarios=[_partial_then_recoverable_error(tokens)])
+        bus = EventBus()
+        deltas: list = []
+        bus.subscribe(
+            event_types=[EventType.LLM_TEXT_DELTA],
+            handler=lambda e: deltas.append(e),
+        )
+
+        ctx = build_ctx(runtime, node_spec, buffer, llm, is_subagent_mode=True)
+        node = EventLoopNode(event_bus=bus, config=LoopConfig(max_iterations=5))
+        result = await node.execute(ctx)
+
+        assert llm._call_index == 1
+        assert len(deltas) == 20  # exactly the 20 published tokens, none re-sent
+        assert result.success is True
+
+    @pytest.mark.asyncio
+    async def test_recoverable_error_before_text_triggers_outer_retry(self, runtime, node_spec, buffer):
+        node_spec.output_keys = []
+        llm = MockStreamingLLM(
+            scenarios=[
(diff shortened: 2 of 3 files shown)


EsraaKamel11:
@Hundao Your root-cause call was exactly right, and I've implemented it precisely as described.

The bug was never in the empty-stream path — it was the two exception handlers retrying after `TextDeltaEvent`s had already gone out on the bus. Both now check whether content was already published before retrying, and bail out with a recoverable error instead of re-streaming:

```python
except RateLimitError as e:
    if attempt < RATE_LIMIT_MAX_RETRIES:
        if accumulated_text:                       # content already published — cannot be recalled
            yield StreamErrorEvent(error=str(e), recoverable=True)
            return
        ...                                        # unchanged: retry only when nothing was yielded
```

Point by point against your review:

- **Guard lives in both exception handlers**, not the retry loop generally — `RateLimitError` (`core/framework/llm/litellm.py:2696`) and the transient `except Exception` branch (`:2761`).
- **The `not has_content` empty-stream path is untouched** — it correctly retries when nothing was emitted, exactly as you noted. The diff is two hunks, both inside the except handlers, nothing else.
- **"Check whether content was already yielded before retrying; if so, don't retry, just yield a StreamErrorEvent"** — that's the `if accumulated_text:` guard, placed at the top of each retry branch before any backoff/`continue`.

Two deliberate details:

- The error is yielded with **`recoverable=True`** on purpose: that's what makes the consumer commit the partial turn and skip the outer retry (`agent_loop.py:2634`/`:2857`). `recoverable=False` would raise and discard the partial text.
- The guard keys on **`accumulated_text` only, not `tool_calls_acc`** — tool deltas are buffered and never published before the stream completes, so tool-only streams remain safe to retry internally (covered by a dedicated test).

Note: this is reworked onto current `main` (post graph→orchestrator revamp), so the consumer is now `AgentLoop` and the line numbers differ from the original `EventLoopNode`/L1085-1108 references — the logic is identical.

Coverage: 8 tests added — 5 provider-level (incl. a 50-chunk no-duplication case and the tool-only-still-retries case) and 3 AgentLoop integration tests proving the end-to-end contract (recoverable-after-text → `_call_index == 1`, no re-stream; recoverable-before-text → `_call_index == 2`, outer retry fires). Verified by TDD: removing the guards makes the duplication tests fail. `ruff check` + `ruff format --check` clean; existing provider suite (145) passes.

PR #7287, opened 2026-06-25
Fixes #7286

### LINKED ISSUE #7286: [Feature]: Enforce a structured output_schema contract on worker → queen reports
## Problem Statement

A queen decomposes a goal into worker tasks and each worker reports its result back via
`report_to_parent`. Today that report has no contract: the worker's `data` is arbitrary
JSON. `handle_report_to_parent`
(`core/framework/agent_loop/internals/synthetic_tools.py`) only coerces `data` into a dict,
and the loop records it verbatim. So a worker can call
`report_to_parent(status="success", data={...anything...})` and the framework treats it as
a completed success, regardless of shape.

That is risky for a coordinator/worker system:
- A worker can claim success with malformed, incomplete, or hallucinated data, and the queen
  records it as a finished result.
- A downstream consumer (the queen, or another worker reading the report) can silently break
  or produce wrong results because the shape isn't what it expected.
- There is no machine-checkable way to say "this task must return `{rows: int}`" and hold the
  worker to it.

## Proposed Solution

Let the queen declare the expected result shape when it delegates a task, and have the
framework check the worker's result against it.

- When the queen fans out work with `run_parallel_workers`, each task may carry an
  `output_schema`: a map of expected output key to JSON-Schema fragment, e.g.
  `{"rows": {"type": "integer"}, "table": {"type": "string"}}`.
- That schema travels with the spawned worker. When the worker calls `report_to_parent` with
  `status="success"`, the framework validates `data` against the schema (jsonschema Draft7).
- If `data` doesn't conform, the success claim is not allowed to stand: the report is recorded
  as `failed` instead of `success`, with the validation reason attached to the summary. The
  violating data is preserved for diagnosis.
- Failed/partial reports are exempt (a worker signalling incompleteness shouldn't have to
  satisfy the success contract). A task with no `output_schema` is unaffected (backward
  compatible).

`output_schema` is opt-in per task: declare it when the result is structured enough to
validate (typed fields, lists, records); omit it for free-form/prose results, where a schema
would be trivially satisfiable and add no real safety. A colony-level "require a schema for
every task" policy could be layered on later.

The result: a worker can't report a malformed payload as a success; a contract violation
surfaces as a `failed` result with a reason.

## Alternatives Considered

- Validate in the queen's consumption logic instead of at report time: leaves the bad
  "success" recorded and makes every consumer re-validate. Checking at the report boundary
  catches it once, at the source.
- Raise/hard-error on violation: would break the worker loop. Recording a normal `failed`
  result keeps the loop intact and preserves the data for diagnosis.
- Pydantic models as the contract: heavier; requires shipping model classes across the spawn
  boundary. JSON-Schema fragments serialize cleanly and match the existing
  `OutputValidator.validate_schema`.
- Enforce on failed/partial too: rejected; that would punish honest incompleteness.

## Additional Context

- Reuses the existing `OutputValidator.validate_schema`
  (`core/framework/orchestrator/validator.py`), which already backs the legacy graph
  orchestrator's output validation.
- For the queen to actually author a schema per task, the delegation prompt needs a small
  additive change: instruct the queen to declare the expected result shape for each task it
  fans out. This does not change how the queen splits or assigns work, it only adds a
  "declare the return shape" step. The enforcement mechanism is independent of this and runs
  whenever a schema is present.
- A colony- or profile-level default schema (so every worker inherits a contract without
  per-task declaration) is a reasonable follow-up; there is no config path that populates one
  today.
- Scope: this is about the result a worker returns. Validating the instruction / `input_data`
  a worker receives (a queen to worker input contract) is a separate, logical follow-up.
- `jsonschema` is currently only a transitive dependency, and `validate_schema` silently skips
  validation if it can't import it, so enforcement would quietly no-op if that dep ever went
  away. Worth declaring `jsonschema` as a direct dependency.

## Implementation Ideas

- Declaration: accept a per-task `output_schema` in `run_parallel_workers`
  (`core/framework/tools/queen_lifecycle_tools.py`), plus a delegation-prompt addition so the
  queen fills it in.
- Propagation: `spawn_batch` clones the worker spec with the task's schema and passes it to
  `spawn` (`core/framework/host/colony_runtime.py`), so it rides on
  `ctx.agent_spec.output_schema`.
- Enforcement: validate in the `report_to_parent` handler
  (`core/framework/agent_loop/agent_loop.py`) via `OutputValidator.validate_schema`; record
  `failed` with the reason on violation.
- Tests: conforming/violating success, missing/empty data, multiple violations, nested and
  array schemas, failed-report exemption, no-schema passthrough, and the jsonschema-absent
  no-op.

Aqeedkalburgi:
Add optional output_schema support for delegated worker tasks and validate successful report_to_parent payloads using the existing OutputValidator. Schema violations are recorded as failed reports while preserving the original payload for debugging. Includes comprehensive regression tests and maintains full backward compatibility for tasks without an output_schema.

## Description

Fixes #7286

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #(issue number)

## Changes Made

- Change 1
- Change 2
- Change 3

## Testing

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

Add screenshots to demonstrate UI changes.

## Summary by CodeRabbit

* **New Features**
  * Added optional output schema support for worker tasks, with validation of reported results against the expected shape.
  * Expanded parallel worker runs to accept per-task schema overrides.

* **Bug Fixes**
  * Invalid success reports now return as failed with a clear validation error in the summary.
  * Worker completion status now reflects explicit failed reports correctly.
  * Improved error handling so reported errors are preserved in final results.


Aqeedkalburgi (commented):
Thanks for reviewing this PR.

This change introduces optional `output_schema` enforcement for worker → queen reports by validating successful `report_to_parent` payloads against a declared JSON Schema. The goal is to prevent malformed structured outputs from being recorded as successful results while remaining fully backward compatible for tasks that do not define an `output_schema`.

Key points:

* `output_schema` is opt-in and propagated with delegated worker tasks.
* Validation is applied only to `status="success"` reports.
* Schema violations are converted into failed reports while preserving the original payload for debugging.
* `failed` and `partial` reports remain exempt from validation.
* Added comprehensive regression tests covering valid payloads, invalid payloads, nested objects, arrays, missing fields, multiple validation failures, no-schema behavior, and failed/partial exemptions.

I'm happy to make any adjustments if maintainers prefer a different integration point or enforcement strategy.

PR #7338, opened 2026-07-30
fix(quickstart): preserve empty TSV fields in catalog loader and reader

AndlerRL:
## Summary

Bash 3.2.57's `read -r a b c d e f g` silently drops empty
tab-separated fields in the middle of input. Two sites in
quickstart.sh use this pattern against catalog data whose rows
may have empty middle fields (e.g., ollama_local's empty model
and api_key_env_var). Bash collapsed them, shifting the URL
`http://localhost:11434` into the `max_context_tokens` position
and crashing `save_configuration` with
`ValueError: invalid literal for int() with base 10: 'http://localhost:11434'`.

This PR fixes the bug, exposes stderr on the affected calls so
the error is no longer hidden, and adds regression test coverage.

## What changed

- `quickstart.sh`: `load_model_catalog_rows` uses parameter
  expansion (${line%%...} and ${line#*...}) instead of
  bash `read -r a b c ...`. `get_preset_field` uses
  awk (`-F'\t'`) instead of bash `read`. Both changes
  preserve empty fields correctly on bash 3.2 and bash 5.x.
- `quickstart.sh`: moved the catalog loader and accessor
  block to the top of the script behind a sourceable-mode
  guard so `source quickstart.sh` exposes the library without
  running the interactive wizard. When sourced, the guard
  returns before `set -e`; when executed, the guard is
  skipped and the wizard runs unchanged.
- `quickstart.sh`: dropped `2>/dev/null` from the
  `save_configuration` Python invocation and the
  `> /dev/null` from the Ollama branch, so the underlying
  ValueError is no longer hidden.
- `core/framework/llm/model_catalog.json`: catalog data for
  ollama_local (max_context_tokens 16384 → 131072), ollama_cloud
  preset (new), and the ollama provider (new entry with 5 models).
- `tests/test_quickstart_parsing.sh`: new regression test
  covering empty-field round-trip on ollama_local / ollama_cloud
  and fully-populated round-trip on zai_code.

## Why this has been latent

- macOS-only on the default shell; maintainers running Homebrew
  bash 5.x see no failure.
- Only `ollama_local` and `ollama_cloud` have empty optional
  fields in the catalog today.
- No test ran the bash layer end-to-end; the Python
  `model_catalog` unit tests cover the data, not the consumers.

## Test plan

```bash
bash tests/test_quickstart_parsing.sh
```

Should print all `PASS:` lines on macOS bash 3.2.57. The same
test on bash 5.x also passes (verified locally).

## Bounty

This PR claims the **Medium Standard Bounty (30 pts)** filed
at the issue associated with this PR. See the bounty program
docs at `docs/bounty-program/README.md` for tier details.

🤖 Verified end-to-end against real `PRESET_ROWS` output on
bash 3.2.57. No AI-only verification gap.

## Summary by CodeRabbit

- **New Features**
  - Added Ollama as a supported provider with local and cloud configurations.
  - Added Ollama model options, including vision support and context limits.
  - Added the `qwen2.5:72b` default model.
  - Improved configuration output and error visibility during setup.

- **Bug Fixes**
  - Improved model catalog and preset parsing for more reliable quickstart configuration.

- **Tests**
  - Added validation for Ollama and existing preset settings.

### DIFF
--- a/core/framework/llm/model_catalog.json
+++ b/core/framework/llm/model_catalog.json
@@ -435,6 +435,51 @@
           "supports_vision": true
         }
       ]
+    },
+    "ollama": {
+      "default_model": "qwen2.5:72b",
+      "models": [
+        {
+          "id": "qwen2.5:72b",
+          "label": "Qwen 2.5 72B - Recommended for Hive",
+          "recommended": true,
+          "max_tokens": 32768,
+          "max_context_tokens": 131072,
+          "supports_vision": false
+        },
+        {
+          "id": "minimax-m3:cloud",
+          "label": "Minimax M3 (Cloud) - Minimax flagship",
+          "recommended": false,
+          "max_tokens": 32768,
+          "max_context_tokens": 524288,
+          "supports_vision": true
+        },
+        {
+          "id": "qwen2.5:32b",
+          "label": "Qwen 2.5 32B - Balanced",
+          "recommended": false,
+          "max_tokens": 32768,
+          "max_context_tokens": 131072,
+          "supports_vision": false
+        },
+        {
+          "id": "qwen3:32b",
+          "label": "Qwen 3 32B - Newer reasoning",
+          "recommended": false,
+          "max_tokens": 32768,
+          "max_context_tokens": 131072,
+          "supports_vision": false
+        },
+        {
+          "id": "llama3.3:70b",
+          "label": "Llama 3.3 70B - General purpose",
+          "recommended": false,
+          "max_tokens": 32768,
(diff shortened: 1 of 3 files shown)


AndlerRL:
Closing this PR — it was filed before claiming the associated Medium Standard Bounty. Re-opening (or filing a new PR from the same branch) once the bounty is filed and assigned per the [Bounty Program](https://github.com/aden-hive/hive/blob/main/docs/bounty-program/README.md) workflow.

PR #7359, opened 2026-08-14
[Feature] Route raw exceptions in GraphExecutor to existing retry logic

### LINKED ISSUE #5105: [Feature]: Add system-level reliability layer to GraphExecutor
### Problem Definition
Currently, the `GraphExecutor` lacks a dedicated reliability layer for handling unexpected system-level exceptions during node execution. If a node encounters a transient infrastructure failure (e.g., a network timeout, a database connection drop, or an external API returning a 500 error), the `GraphExecutor` catches the exception and immediately terminates the entire agent run.

While the framework handles semantic failures (logic-based), it is currently fragile to transient infrastructure "blips" which are common in production environments.

### Proposed Solution
I propose adding a infrastructure-level retry loop within `GraphExecutor.execute()`. This layer will:
1. **Isolate System Crashes:** Specifically catch unexpected `Exceptions` while allowing `asyncio.CancelledError` to propagate (preserving HITL and TUI responsiveness).
2. **Implement Exponential Backoff:** Use a progressive delay (1s, 2s, 4s...) to allow external systems time to recover.
3. **Add Jitter:** Incorporate randomness into the retry timing to prevent "thundering herd" synchronization issues across multiple concurrent agents.
4. **Graceful Degradation:** If max system retries (default 3) are exhausted, the failure will bubble up to the standard framework failure handlers.

### Use Case
This is critical for long-running agents that depend on external MCP tools or third-party APIs. It ensures that a single network packet drop doesn't ruin an agent's multi-step progress.

### Additional Context
I have already developed the implementation for this and verified it with unit tests and manual chaos testing. I will be linking a PR shortly.

bhusareMayur:
Fixes: #5105

## Summary

Route raw (system-level) exceptions thrown by node implementations into the existing node-level retry mechanism so transient infra failures (network timeouts, DB drops, 500s, etc.) do not crash the entire agent run. Preserve cancellation semantics so `asyncio.CancelledError` continues to propagate immediately.

## Problem

Previously, `node_impl.execute(ctx)` could raise uncaught exceptions that bypassed the node retry loops and terminated the whole graph run. This makes the system fragile to temporary infrastructure "blips".

## Changes

- Wrap node execution in both execution paths to capture raw exceptions and convert them into a failure `NodeResult` so existing retry/backoff logic is used:
  - core/framework/orchestrator/orchestrator.py
    - Wraps `node_impl.execute(ctx)` in the branch execution loop with a try/except.
    - Converts non-Cancelled exceptions into `NodeResult(success=False, error=...)`.
    - Keeps existing logging and node-retry event emission.
    - Adds exponential backoff with jitter to retry sleeps.
  - core/framework/orchestrator/node_worker.py
    - Ensures `_execute_with_retries()` treats non-Cancelled exceptions as retriable failures.
    - Re-raises `asyncio.CancelledError` to preserve cancellation/HITL/TUI responsiveness.
    - Adds jitter to backoff delays to avoid thundering-herd retries.

- Tests added / updated:
  - core/framework/runtime/tests/test_system_level_retries.py
    - test_worker_catches_raw_exception_and_retries
    - test_worker_respects_asyncio_cancelled_error

## Files changed (high level)

- core/framework/orchestrator/orchestrator.py
- core/framework/orchestrator/node_worker.py
- core/framework/runtime/tests/test_system_level_retries.py

## Test results (local)

- Ran: pytest core/framework/runtime/tests/test_system_level_retries.py
- Result: 2 passed, 4 warnings

## How this addresses the issue

When a node implementation raises a non-Cancelled exception, it is now converted into a `NodeResult(success=False, ...)` and handled by the existing retry loop which respects the node `max_retries` and uses exponential backoff + jitter. `asyncio.CancelledError` is re-raised so cancellations and interactive flows are not degraded.

## Verification / QA

- Unit tests included above validate both retry-on-exception and the correct propagation of CancelledError.
- Recommended additional check: run an integration scenario where a node intermittently raises a transient exception and confirm:
  - The node is retried with exponential backoff and jitter.
  - The graph recovers if the node succeeds within the configured retries or fails cleanly after max retries.

## Notes / Follow-ups

- Ensure CI runs the full test matrix.
- Optionally add an orchestrator-level integration test if one does not already exist to cover the non-worker branch loop end-to-end.
- Consider a short changelog/release note mentioning improved reliability for transient infra failures.

Requesting review and merge. Thanks!

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when parallel tasks encounter unexpected errors by routing failures through the existing retry process.
  * Added exponential retry delays with randomized timing to improve recovery from transient failures.
  * Ensured retry processing performs at least one attempt when configured with no retries.
  * Transient node failures can now recover successfully after retrying, with retry counts tracked accurately.
  * Task cancellation is preserved correctly and no longer treated as a retryable failure.
  * Improved cleanup of interrupted background tasks.

* **Tests**
  * Added coverage for successful recovery after repeated failures and proper cancellation behavior.

### DIFF
--- a/core/framework/orchestrator/node_worker.py
+++ b/core/framework/orchestrator/node_worker.py
@@ -424,6 +424,15 @@ class NodeWorker:
                         max(1, max_retries),
                         delay,
                     )
+                    # Emit retry event for raw exceptions
+                    if gc.event_bus:
+                        await gc.event_bus.emit_node_retry(
+                            stream_id=gc.stream_id,
+                            node_id=self.node_spec.id,
+                            attempt=attempt + 1,
+                            max_retries=max_retries,
+                            execution_id=gc.execution_id,
+                        )
                     await asyncio.sleep(delay)
                     continue
                 return NodeResult(
--- a/core/framework/orchestrator/node_worker.py
+++ b/core/framework/orchestrator/node_worker.py
@@ -380,7 +380,8 @@ class NodeWorker:
                 if attempt + 1 < total_attempts:
                     gc.retry_counts[self.node_spec.id] = gc.retry_counts.get(self.node_spec.id, 0) + 1
                     gc.nodes_with_retries.add(self.node_spec.id)
-                    delay = 1.0 * (2**attempt)
+                    import random
+                    delay = (1.0 * (2**attempt)) + random.uniform(0.1, 0.5)
                     logger.warning(
                         "Worker %s failed (attempt %d/%d), retrying in %.1fs: %s",
                         self.node_spec.id,
@@ -413,7 +414,8 @@ class NodeWorker:
                 if attempt + 1 < total_attempts:
                     gc.retry_counts[self.node_spec.id] = gc.retry_counts.get(self.node_spec.id, 0) + 1
                     gc.nodes_with_retries.add(self.node_spec.id)
-                    delay = 1.0 * (2**attempt)
+                    import random
+                    delay = (1.0 * (2**attempt)) + random.uniform(0.1, 0.5)
                     logger.warning(
                         "Worker %s raised %s (attempt %d/%d), retrying in %.1fs",
                         self.node_spec.id,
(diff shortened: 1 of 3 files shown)


bhusareMayur:
@Hundao  @vincentjiang777  @TimothyZhang7  @bryanadenhq 
All automated review feedback has been addressed and the CodeRabbit checks are officially green. Ready for a human review whenever you have the time!

PR #7371, opened 2026-08-16
fix: don't crash quickstart.sh on unset $TERM, surface real uv sync errors

### LINKED ISSUE #7368: quickstart.sh crashes on unset $TERM, and hides the real error on workspace install failure
## Description

Two related bugs in `quickstart.sh`:

1. Line 96 (and 2317) call `clear` unconditionally under `set -e` (line 12). If `$TERM` is unset in the shell (happens in some IDE terminals, CI, non-interactive shells), `clear` exits 1 and `set -e` kills the entire script before any real setup happens — with only `TERM environment variable not set.` printed, no explanation.
2. Line 263 runs `uv sync > /dev/null 2>&1` and on failure only prints `✗ workspace installation failed` — the actual error is discarded. This makes any `uv sync` failure (network timeout, disk space, etc.) undiagnosable from the script's own output.

## Repro

```bash
env -u TERM ./quickstart.sh
```
→ immediate exit after the `[debug] Bash version` line, no further explanation.

Separately, any `uv sync` failure (e.g. slow network fetching the pinned Python 3.11 build) surfaces zero useful detail — just `✗ workspace installation failed`.

## Proposed fix

- Guard the `clear` calls: `clear 2>/dev/null || true`
- Capture `uv sync`'s actual output on failure instead of suppressing it, e.g.:
  ```bash
  if UV_SYNC_OUTPUT=$(uv sync 2>&1); then
      ...
  else
      echo "$UV_SYNC_OUTPUT"
      exit 1
  fi
  ```

I have a working patch for both ready to submit as a PR once this is assigned.

NT1906:
## Description
Fixes two robustness bugs in `quickstart.sh` that make failures undiagnosable. `clear` was called unconditionally under `set -e`, so an unset `$TERM` silently kills the whole script. Separately, `uv sync`'s output was redirected to `/dev/null`, so any install failure just printed a generic message with no way to tell what went wrong.

## Motivation
Ran into this firsthand: `./quickstart.sh` died immediately with just `TERM environment variable not set.` and no further explanation. After bypassing that, the workspace-install step failed with only `✗ workspace installation failed` printed — the real error (a `uv sync` timeout fetching the pinned Python 3.11 build) was being swallowed by `> /dev/null 2>&1`.

## Changes
- Guard both `clear` calls: `clear 2>/dev/null || true`
- Capture `uv sync` output via `UV_SYNC_OUTPUT=$(uv sync 2>&1)` and print it when the sync fails

## Testing
- [x] `bash -n quickstart.sh` passes
- [x] Verified the `if VAR=$(failing_cmd)` capture pattern doesn't trigger early exit under `set -e`
- [x] Ran the package-install step of `quickstart.sh` successfully end to end after the fix

## Checklist
- [x] Code follows style guidelines (matches existing script conventions)
- [x] Self-review completed
- [x] No breaking changes

Closes #7368

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quickstart reliability when terminal clearing is unavailable.
  * Installation failures now display helpful command output for easier troubleshooting.

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -265,7 +265,7 @@ if [ -f "pyproject.toml" ]; then
     else
         echo -e "${RED}  ✗ workspace installation failed${NC}"
         echo ""
-        echo -e "${DIM}${UV_SYNC_OUTPUT}${NC}"
+        printf '%b%s%b\n' "$DIM" "$UV_SYNC_OUTPUT" "$NC"
         exit 1
     fi
 else
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -93,7 +93,7 @@ prompt_choice() {
     done
 }
 
-clear
+clear 2>/dev/null || true
 echo ""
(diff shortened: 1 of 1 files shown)


NT1906:
Hi Please review this pull request.

PR #7418, opened 2026-09-06
fix(terminal-tools): warn on destructive deletes behind sudo and other command prefixes

### LINKED ISSUE #7417: [Bug]: destructive_warning is silent for sudo rm -rf and other command prefixes
### Summary

`destructive_warning.get_warning()` returns `None` for `sudo rm -rf /`.

The file-deletion patterns in `tools/src/terminal_tools/common/destructive_warning.py` anchor the verb to command position — `(^|[;&|\n]\s*)rm\s+…` — so any prefix that displaces `rm` silences the warning completely.

### Reproduction

No hardware or device needed; this is pure catalog behaviour.

```python
from terminal_tools.common.destructive_warning import get_warning

get_warning("rm -rf /")        # 'may recursively force-remove files'
get_warning("sudo rm -rf /")   # None   <-- expected a warning
```

Measured against current `main`:

| Command | `get_warning()` |
| --- | --- |
| `rm -rf /` | may recursively force-remove files |
| `sudo rm -rf /` | **None** |
| `doas rm -rf /` | **None** |
| `sudo -E rm -rf /home` | **None** |
| `sudo -u root rm -rf /srv` | **None** |
| `nohup rm -rf /data` | **None** |
| `time rm -rf /data` | **None** |
| `env rm -rf /data` | **None** |
| `FOO=1 rm -rf /data` | **None** |
| `find . -name '*.pyc' \| xargs rm -rf` | **None** |
| `make clean && sudo rm -rf build` | **None** |

### Why this matters

The catalog is inconsistent about prefixes. The git, kubectl and terraform patterns use `\b` rather than a command anchor, so these already warn correctly today:

```python
get_warning("sudo git reset --hard")     # 'may discard uncommitted changes'
get_warning("sudo kubectl delete pod x") # 'may delete Kubernetes resources'
```

Only the deletion class — the most destructive one, and the one most likely to be run under `sudo` — goes quiet. The practical effect is that an agent that reaches for `sudo` gets *less* safety feedback in the exec envelope than one that does not, which inverts the intent described in the module docstring.

### Expected behaviour

A destructive deletion should produce the same warning whether or not the verb is displaced by `sudo`, `doas`, `env`, `nohup`, `time`, `command`, `xargs`, or a leading `VAR=value` assignment.

Quoted or unrelated text must stay silent, so the prefix handling must not become a wildcard — `echo sudo rm -rf /` and `sudo apt update` should both remain `None`.

### Scope

`destructive_warning.py` only. This is the advisory catalog; nothing here changes blocking behaviour, which lives in `command_guard.py` and is scoped to browser/runtime processes.

Windows verbs need no change — there is no `sudo` equivalent displacing `del`/`rd` on that path.

I have a fix and regression tests ready and will open a PR referencing this issue.

woahwhattheheck:
## Summary

Fixes #7417.

`destructive_warning.get_warning()` returns `None` for `sudo rm -rf /`. The file-deletion patterns anchor the verb to command position — `(^|[;&|\n]\s*)rm\s+…` — so any prefix that displaces `rm` silences the warning completely.

| Command | Before | After |
| --- | --- | --- |
| `rm -rf /` | may recursively force-remove files | unchanged |
| `sudo rm -rf /` | **None** | may recursively force-remove files |
| `doas rm -rf /` | **None** | may recursively force-remove files |
| `sudo -E rm -rf /home` | **None** | may recursively force-remove files |
| `sudo -u root rm -rf /srv` | **None** | may recursively force-remove files |
| `nohup` / `time` / `env rm -rf /data` | **None** | may recursively force-remove files |
| `FOO=1 rm -rf /data` | **None** | may recursively force-remove files |
| `… \| xargs rm -rf` | **None** | may recursively force-remove files |
| `sudo rm -r /var/log` | **None** | may recursively remove files |
| `sudo rm -f /var/log/x` | **None** | may force-remove files |

The catalog was inconsistent about this: the git, kubectl and terraform patterns use `\b` rather than a command anchor, so `sudo git reset --hard` and `sudo kubectl delete pod x` already warned. Only the deletion class — the most destructive, and the one most likely to be run under `sudo` — went quiet, so an agent reaching for `sudo` got *less* safety feedback than one that did not.

## Change

One shared fragment, consumed between the command-position anchor and `rm`:

```python
_CMD_PREFIX = (
    r"(?:(?:sudo|doas|nohup|time|command|env|xargs)(?:\s+-\w+(?:\s+[^-\s]\S*)?)*\s+"
    r"|[A-Za-z_]\w*=\S*\s+)*"
)
```

Applied to the three `rm` patterns only. It accepts the prefix verbs with their short flags — including a flag's argument, so `sudo -u root rm -rf /srv` matches — and leading `VAR=value` assignments, repeatable (`sudo env FOO=1 rm -rf /data`).

**Deliberately not a wildcard.** Only the listed verbs, their flags, and assignments may be consumed, so an unrelated leading command still blocks the match: `echo sudo rm -rf /`, `echo "sudo rm -rf /"`, `sudo apt update`, `sudo docker ps`, `sudo systemctl restart nginx`, `grep -r rm .` and `cat sudo_rm_notes.txt` all stay `None`.

Advisory only — this does not touch blocking behaviour, which lives in `command_guard.py` and is scoped to browser/runtime processes.

## Validation

- `pytest tools/tests/test_terminal_tools_security.py -k destructive_warning` → **4 passed**.
- `test_destructive_warning_command_prefixes` **fails without this change** (`AssertionError: expected warning for 'sudo rm -rf /'`), so it is a real regression test rather than one written to fit the code.
- Differential check of the pre- vs post-change catalog over 60 inputs: 16 prefixed deletions go from `None` to the correct descriptive warning, 17 benign inputs stay `None`, and every command that already warned returns the **byte-identical** string.
- The argv-list form (`get_warning(["sudo", "rm", "-rf", "/tmp/x"])`) is covered.
- Longest added line is 90 chars, within `line-length = 120`.

## Relationship to #7416

Independent — this branches from `main` and touches only the `rm` patterns, while #7416 adds Windows/PowerShell delete verbs. Either can merge first; the second will need a trivial context resolution in the same block. Happy to rebase whichever you prefer, or fold them together if you would rather review one change.

I reproduced the missing-warning behaviour against the current module myself and verified each pattern and test result rather than submitting unreviewed output.

## Summary by CodeRabbit

- **Bug Fixes**
  - Destructive-command warnings now recognize file-deletion commands preceded by common command wrappers, privilege-elevation tools, environment assignments, and command pipelines.
  - Unrelated commands and quoted text continue to avoid triggering deletion warnings.

- **Tests**
  - Added coverage for supported command prefixes and cases that should remain silent.

### DIFF
--- a/tools/src/terminal_tools/common/destructive_warning.py
+++ b/tools/src/terminal_tools/common/destructive_warning.py
@@ -11,6 +11,17 @@ from __future__ import annotations
 import re
 from collections.abc import Sequence
 
+# Command-position prefixes that displace the verb without changing what it
+# does: ``sudo rm -rf /`` deletes exactly what ``rm -rf /`` deletes. Also
+# covers leading environment assignments (``FOO=1 rm -rf /``) and the
+# ``xargs`` spelling reached through a pipe. Without this the file-deletion
+# patterns stay silent under ``sudo`` while the git/kubectl/terraform
+# patterns, which are not command-anchored, still warn.
+_CMD_PREFIX = (
+    r"(?:(?:sudo|doas|nohup|time|command|env|xargs)(?:\s+-\w+(?:\s+[^-\s]\S*)?)*\s+"
+    r"|[A-Za-z_]\w*=\S*\s+)*"
+)
+
 _PATTERNS: tuple[tuple[re.Pattern[str], str], ...] = (
     # Git — data loss / hard to reverse
     (re.compile(r"\bgit\s+reset\s+--hard\b"), "may discard uncommitted changes"),
@@ -34,11 +45,14 @@ _PATTERNS: tuple[tuple[re.Pattern[str], str], ...] = (
     (re.compile(r"\bgit\s+commit\b[^;&|\n]*--amend\b"), "may rewrite the last commit"),
     # File deletion — most specific patterns first so the warning is descriptive
     (
-        re.compile(r"(^|[;&|\n]\s*)rm\s+-[a-zA-Z]*[rR][a-zA-Z]*f|(^|[;&|\n]\s*)rm\s+-[a-zA-Z]*f[a-zA-Z]*[rR]"),
+        re.compile(
+            rf"(^|[;&|\n]\s*){_CMD_PREFIX}rm\s+-[a-zA-Z]*[rR][a-zA-Z]*f"
+            rf"|(^|[;&|\n]\s*){_CMD_PREFIX}rm\s+-[a-zA-Z]*f[a-zA-Z]*[rR]"
+        ),
         "may recursively force-remove files",
     ),
-    (re.compile(r"(^|[;&|\n]\s*)rm\s+-[a-zA-Z]*[rR]"), "may recursively remove files"),
-    (re.compile(r"(^|[;&|\n]\s*)rm\s+-[a-zA-Z]*f"), "may force-remove files"),
+    (re.compile(rf"(^|[;&|\n]\s*){_CMD_PREFIX}rm\s+-[a-zA-Z]*[rR]"), "may recursively remove files"),
+    (re.compile(rf"(^|[;&|\n]\s*){_CMD_PREFIX}rm\s+-[a-zA-Z]*f"), "may force-remove files"),
     # Database
     (
         re.compile(r"\b(DROP|TRUNCATE)\s+(TABLE|DATABASE|SCHEMA)\b", re.IGNORECASE),
--- a/tools/tests/test_terminal_tools_security.py
+++ b/tools/tests/test_terminal_tools_security.py
@@ -147,3 +147,49 @@ def test_semantic_exit_timed_out():
     status, msg = classify("sleep 999", None, timed_out=True)
     assert status == "error"
     assert "timed out" in msg.lower()
+
+
+def test_destructive_warning_command_prefixes():
+    """`sudo rm -rf /` deletes exactly what `rm -rf /` deletes, but the
+    file-deletion patterns are command-anchored, so a prefix used to displace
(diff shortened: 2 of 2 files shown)


woahwhattheheck:
Verified the merge-ordering claim in the description rather than leaving it as an assertion.

I built the combined result — #7416's Windows/PowerShell patterns plus this PR's `_CMD_PREFIX` — and ran the **union of both PRs' test suites** against it: **13 passed** (11 from #7416, 2 from here), no failures either direction.

The two changes touch disjoint regions of the file:

- **#7416** adds `_SEG`, the `_RECURSE_NAME`/`_FORCE_NAME` prefix alternations, and the Windows/PowerShell block.
- **this PR** adds `_CMD_PREFIX` and rewrites the three `rm` patterns.

They are adjacent inside `_PATTERNS`, so git may flag context, but the resolution is mechanical — keep both blocks in order — and the merged behaviour is the strict union:

| command | #7416 alone | #7418 alone | merged |
| --- | --- | --- | --- |
| `sudo rm -rf /` | None | warns | warns |
| `del /F /S /Q C:\temp` | warns | None | warns |
| `Remove-Item -Recurse -Force C:\x` | warns | None | warns |
| `rm -rf /tmp/foo` | warns | warns | warns |
| `Remove-Item -recurseTypo C:\x` | None | None | None |
| `echo sudo rm -rf /` | None | None | None |

So: merge in whichever order suits you, and neither needs to wait on the other. If you would rather take it as a single change, say the word and I will fold this into #7416 and close this one.

Same applies to #7420 — that touches `command_guard.py` and `test_command_guard.py`, which neither of these PRs modifies, so it is independent of both.

PR #7421, opened 2026-09-06
docs: drop references to editor configs the repo no longer ships

### LINKED ISSUE #7383: A skills cleanup also removed the MCP and editor configs, and four docs still point at them
`13a8e28a` ("refactor: remove all old unused skills", 2026-03-04, on `main`) removed the old
skills trees. It also removed five files that are not skills:

```
.agent/mcp_config.json
.codex/config.toml
.cursor/mcp.json
.opencode/mcp.json
.vscode/extensions.json
```

Of those directories only `.claude/` is still on `main`. `.agent/`, `.codex/`, `.cursor/`,
`.opencode/` and `.vscode/` are gone entirely.

Four places still tell a contributor to use them:

- `docs/contributing-lint-setup.md:77` : "The repository includes `.vscode/extensions.json`
  and `.vscode/settings.json`. On first open, VS Code will prompt you to install the
  recommended Ruff extension."
- `docs/contributing-lint-setup.md:103` : "Codex CLI (OpenAI, v0.101.0+) is supported via
  `.codex/config.toml` (MCP server config). **This file is tracked in git.**"
- `docs/environment-setup.md:106` : "Restart Cursor to load the MCP servers from
  `.cursor/mcp.json`"
- `docs/developer-guide.md:81` and `:138`, in the project-structure tree: `.codex/` as
  "Codex CLI project config" and `.agent/` as "Antigravity IDE: mcp_config.json + skills
  (symlinks)"

One of those is wrong for a second reason. `.vscode/settings.json` has no add-or-delete
history at all:

```
$ git log --oneline --diff-filter=AD -- .vscode/settings.json
$
```

It was never tracked, so that sentence did not become false in March. It was never true.

No patch, because the two fixes are opposite diffs and only you know which one you want.
Either the MCP and editor configs come back, or the docs stop promising them. The Codex line
is the one I would fix first whichever way you go, since it states the file is tracked in git
as a fact a reader can act on.

Found with [docproof](https://github.com/melbinjp/docproof), which checks what documentation
claims against what the repository actually contains.

woahwhattheheck:
Fixes #7383.

The issue reported four doc sites still pointing at editor and MCP configs that
`13a8e28a` ("refactor: remove all old unused skills") deleted, and said it stopped
short of a patch because the two possible fixes are opposite diffs — either the
configs come back, or the docs stop promising them.

I went with the second, and the repository itself settles which one is intended.

## What I measured against `main` (`0c38749`)

All six named files are absent:

| path | on `main` |
|---|---|
| `.agent/mcp_config.json` | gone |
| `.codex/config.toml` | gone |
| `.cursor/mcp.json` | gone |
| `.opencode/mcp.json` | gone |
| `.vscode/extensions.json` | gone |
| `.vscode/settings.json` | not present |

Nothing recreates them at setup time. The only match in `quickstart.sh` is
`$HOME/.codex/auth.json` — the user's own Codex auth, not a repo config — so these
are stale docs rather than generated files.

**`.gitignore` answers the question the issue left open**, without needing history:

```
.vscode/*
!.vscode/extensions.json
!.vscode/settings.json.example
```

`.vscode/settings.json` is deliberately ignored, so "The repository includes
`.vscode/settings.json`" was never going to be true. `extensions.json` is un-ignored,
i.e. meant to be tracked, and is the one the cleanup actually took.

## Two problems the issue did not list

- **`docs/contributing-lint-setup.md:103` links to `../README.md#codex-cli`, and that
  anchor does not exist** — `README.md` contains no occurrence of "codex" at all. The
  same line asserted "This file is tracked in git" as a fact a reader can act on.
- Beyond the four reported sites, the stale references also sat in the config table
  (2 rows) and two FAQ answers in the same file, and in the `developer-guide.md`
  component table. Fixed all of them; 11 sites across 3 files.

## Deliberately left alone

- `docs/configuration.md:227` and `CONTRIBUTING.md:289` tell the reader to *create*
  `.vscode/settings.json` themselves. That is correct, and it is exactly what
  `.gitignore` expects — so this PR points the lint doc at that block rather than
  duplicating it.
- `.agents/skills/` in `docs/skills-user-guide.md` and `docs/internal/skill-registry-prd.md`
  is the live cross-client skills convention (`~/.agents/skills/`), a different thing
  from the deleted repo-level `.agent/`. Untouched.
- Dropped the "ruler at column 100" bullet rather than re-homing it: the settings block
  in CONTRIBUTING.md does not set rulers, so keeping it would have replaced one false
  claim with another.

If you would rather restore the configs than correct the docs, say so and I will close
this — but the `.gitignore` rules above suggest at least the `.vscode/settings.json`
sentence was never accurate either way.

## Checks

Documentation only — three `.md` files, no code. No workflow in `.github/workflows/`
lints or builds markdown, so there is no docs job for this to run against.

### DIFF
--- a/docs/contributing-lint-setup.md
+++ b/docs/contributing-lint-setup.md
@@ -74,15 +74,18 @@ git commit --no-verify -m "message"
 
 ### VS Code (Recommended)
 
-The repository includes `.vscode/extensions.json` and `.vscode/settings.json`. On first open, VS Code will prompt you to install the recommended Ruff extension.
+The repository does not ship VS Code configuration. `.gitignore` ignores `.vscode/*`, so
+`.vscode/settings.json` is yours to write and is never committed.
 
-Once installed, the editor will:
+Install the [Ruff extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff),
+then copy the settings block from [CONTRIBUTING.md](../CONTRIBUTING.md#recommended-ide-setup)
+into `.vscode/settings.json`. With that in place the editor will:
 
 - **Format on save** using ruff
 - **Auto-fix lint issues** on save (import sorting, fixable violations)
-- Show a **ruler at column 100**
 
-No manual configuration needed.
+If you would rather not configure the editor, `make lint` and `make format` do the same job
+from the command line, and the pre-commit hook enforces it either way.
 
 ### Other Editors
 
@@ -100,7 +103,8 @@ The repository includes a `.claude/settings.json` hook that automatically runs `
 
 ### Codex CLI
 
-Codex CLI (OpenAI, v0.101.0+) is supported via `.codex/config.toml` (MCP server config). This file is tracked in git. Run `codex` in the repo root to use the configured MCP tools. See the [Codex CLI section in the README](../README.md#codex-cli) for details.
+The repository no longer ships a `.codex/config.toml`. Point Codex CLI at the MCP servers
+through your own Codex configuration; `./quickstart.sh` sets the servers themselves up.
 
 ---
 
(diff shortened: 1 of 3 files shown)


woahwhattheheck:
Merge-ordering note: this PR and #7384 both edit `docs/developer-guide.md`, and the hunks are adjacent rather than overlapping.

- **#7384** (@melbinjp) removes `builder/`, `graph/`, `mcp/`, `runner/` and `tui/` from inside the `core/framework/` subtree (lines 84-97), and `docs/quizzes/` (around 129).
- **this PR** removes the `.codex/` block that sits immediately *above* `core/` (81-83), the `.agent/` line below the tree, and two rows from the component table near line 29.

Both are pure deletions of distinct lines, so the merged result is just the union — remove all of them. Because my `.codex/` deletion ends on the line where #7384's first hunk begins, git may still flag context and ask for a resolution; it is mechanical, and neither PR needs to wait on the other. Merge in whichever order suits you. If you would rather have it as one change, say so and I will fold this into #7384 and close this one.

One thing I noticed while checking the overlap, offered for #7384 rather than changed here: that diagram keeps

```
│   │   ├── runtime/                     # Runtime environment
```

and `core/framework/runtime/` does still exist, so keeping it is right by the letter. But it now contains only `tests/` — no modules. The runtime classes named in `docs/agent_runtime.md` (`AgentRuntime`, `EventBus`, `SharedBufferManager`, `OutcomeAggregator`) all resolve to `core/framework/host/` today, and `host/` is one of the directories #7384 notes neither diagram lists. Not something for this PR to change; just flagging it since it is the same drift.

PR #7422, opened 2026-09-06
fix(quickstart): stop the catalogue parser collapsing empty fields

### LINKED ISSUE #7339: [Bounty]: Fix bash 3.2 empty-field collapse in quickstart.sh catalog loader and reader
### Bounty Size

Medium (30 pts)

### Difficulty

Medium

### Description

`quickstart.sh` parses tab-separated rows emitted by a Python helper. Two sites use the bash pattern `IFS=$' ' read -r a b c …`. On bash 3.2 (macOS system bash) this pattern collapses empty middle fields and shifts subsequent values left. The `ollama_local` preset has empty `model` and `api_key_env_var` columns, so the corrupted row causes `apply_preset` to write `SELECTED_MAX_TOKENS="http://localhost:11434"`. The config writer's `int(...)` cast then crashes with `ValueError`.

The bug crosses the bash boundary twice: catalog → `PRESET_ROWS` (loader, `load_model_catalog_rows`) and `PRESET_ROWS` → named variables (reader, `get_preset_field`). Both sites must be fixed in the same PR; fixing only one re-introduces the corruption on every read.

**Why this has been latent?**

1. macOS-only on the default shell; maintainers running Homebrew bash 5.x see no failure.
2. Only `ollama_local` and `ollama_cloud` have empty optional fields; every other preset populates all seven columns.
3. No test runs the bash layer end-to-end; the Python `model_catalog` unit tests pass because they exercise the data, not its consumer.

### Acceptance Criteria

- [ ] `load_model_catalog_rows` no longer collapses empty fields on bash 3.2
- [ ] `get_preset_field` no longer collapses empty fields on bash 3.2
- [ ] Regression test added to a new `tests/test_quickstart_parsing.sh`
- [ ] `tests/test_quickstart_parsing.sh` runs on `macos-latest` (bash 3.2) in CI
- [ ] All existing providers still work end-to-end (anthropic, openai, ollama_local, ollama_cloud, etc.)
- [ ] No regression in `core/framework/llm/` Python unit tests
- [ ] One PR, two sites, one test file

### Relevant Files

- `quickstart.sh` — loader (`load_model_catalog_rows`) and reader (`get_preset_field`)
- `core/framework/llm/model_catalog.py` — Python helper that emits TSV
- `model_catalog.json` — catalog source (`ollama_local` and `ollama_cloud` shapes)
- `tests/test_quickstart_parsing.sh` — new regression test

### Resources

- #5154 — prior macOS quickstart crash report
- #7073 — prior TSV parsing discussion
- [additional-issue-context-for-reviewers.md](https://github.com/user-attachments/files/30565461/Root.cause.for.reviewer.context.md)

woahwhattheheck:
Closes #7339

## The bug

`quickstart.sh` packs the model catalogue into single-character-delimited rows and reads them back with `IFS=$'\t' read -r a b c ...`. Tab is an **IFS whitespace** character, so bash folds a run of tabs into one delimiter and shifts every later field left. Any row with an empty column loses positions.

`ollama_local` and `ollama_cloud` each have two empty columns (`model`, `api_key_env_var`), so their rows arrive misaligned and the config writer's `int(...)` cast is handed a value that is not a number.

It crosses the boundary twice, as the issue notes: once in the loader that fills `PRESET_ROWS`, once in `get_preset_field` reading them back. Fixing one site alone leaves the other to re-corrupt the row.

**One correction to the issue's framing**, which widens the blast radius rather than narrowing it: this is not macOS-only. IFS-whitespace folding is not a bash 3.2 quirk — it is how `read` splits on space, tab and newline in every bash. I reproduced the collapse on bash 5.3 with the `ollama_local` row shape:

```
input : PRESET <TAB> ollama_local <TAB> ollama <TAB><TAB> 4096 <TAB> 8192 <TAB><TAB> http://localhost:11434

expected  f1=ollama_local f2=ollama f3=       f4=4096 f5=8192 f6= f7=http://localhost:11434
actual    f1=ollama_local f2=ollama f3=4096   f4=8192 f5=http://localhost:11434 f6= f7=
```

macOS just surfaces it first because `ollama_local` is the preset most likely to be chosen there. The same rows misparse on Linux and on Homebrew bash 5.x.

## The fix

Switch the internal catalogue pipeline to the ASCII unit separator `0x1f`, which is not IFS whitespace, so empty fields survive:

```bash
CATALOG_FS=$'\037'
...
while IFS="$CATALOG_FS" read -r row_preset_id row_provider row_model ... ; do
```

All four row types (`DEFAULT`, `MODEL`, `PRESET`, `PRESET_MODEL`) move together. Converting only the two sites named in the issue would leave a mixed-delimiter format in the same file, with the identical latent bug waiting on the other row types the first time one of them gains an optional column.

The separator is confined to `quickstart.sh`: it is chosen in the inline emitter and consumed by the readers below it. `core/framework/llm/model_catalog.py` is **not** touched, so nothing outside the installer sees a format change.

I also split the pure parsing out of `load_model_catalog_rows` into `parse_catalog_rows`, so the loader can be tested without standing up `uv` and Python. `load_model_catalog_rows` still calls the helper and then delegates; its behavior is unchanged.

## The test

New `tests/test_quickstart_parsing.sh` extracts the catalogue region from `quickstart.sh` and exercises the real `parse_catalog_rows` and `get_preset_field` — it does not re-implement them, so it fails if either site regresses. It covers `ollama_local` (two empty columns), `ollama_cloud` (empty model, populated env var), `anthropic` (populated with a trailing empty `api_base`), the `DEFAULT` rows, and the reported crash path: both `int()`-cast fields must be numeric and `api_base` must still look like a URL.

Verified in both directions:

| Separator | Result |
| --- | --- |
| `0x1f` (this PR) | **20/20 checks pass**, exit 0 |
| tab (only the separator reverted, everything else identical) | **10/20 fail**, exit 1 — `max_context_tokens` comes back as `http://localhost:11434` |

The negative control changes one line, so a failure isolates the separator rather than the refactor.

CI: a `quickstart-parsing` job on `macos-latest` runs the test against `/bin/bash` explicitly, since that is the 3.2 the installer actually runs under there; anything newer on `PATH` would not exercise the same shell. It prints `bash --version` first so the log records which shell ran.

## What I ran, and what I did not

- Ran: the new test against the patched script (20/20) and against the tab control (10/20 fail), on bash 5.3. `bash -n` clean on the patched `quickstart.sh`. `ci.yml` parses as YAML.
- **Not run locally: macOS bash 3.2.** I do not have that shell here — the new CI job is what exercises it, and it is the check to watch on this PR.
- **Not run locally: the `core/framework/llm/` Python unit tests.** They need `uv sync` and the project's dependencies. `model_catalog.py` is unmodified, so the data those tests exercise is unchanged; the existing `test` job covers them.
- End-to-end installs for every provider were not performed. The parsing layer is covered by fixtures for the three preset shapes above; the acceptance criterion asking that all providers still work end-to-end is best confirmed by the CI matrix on this PR.

I left the vision-fallback parser at the two `IFS=$'\t' read` sites near the end of the script alone. It reads a different variable from a different producer and is outside this issue's scope, though it is the same pattern and would be worth a look if any of its columns ever become optional.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model catalogue parsing when presets contain empty fields.
  * Preset settings, including API endpoints and maximum token counts, are now retained correctly, preventing related quickstart failures.
  * Improved compatibility with Bash versions included on macOS.

* **Tests**
  * Added regression coverage for model selection and preset parsing across local, cloud, and Anthropic configurations.
  * Improved quickstart parsing validation in continuous integration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,21 @@ jobs:
           uv run playwright install chromium
           uv run pytest tests/ -v
 
+  quickstart-parsing:
+    name: Quickstart Parsing (macOS system bash 3.2)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Show the bash under test
+        run: /bin/bash --version
+
+      # Runs against /bin/bash explicitly: macOS ships bash 3.2 there, which is
+      # the shell the quickstart installer actually runs under for most macOS
+      # users. Anything newer on PATH would not exercise the same code path.
+      - name: Model catalogue parsing regression test
+        run: /bin/bash tests/test_quickstart_parsing.sh
+
   validate:
     name: Validate Agent Exports
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,13 @@ jobs:
     name: Quickstart Parsing (macOS system bash 3.2)
     runs-on: macos-latest
     steps:
+      # This job runs on pull_request and executes a script out of the
+      # checkout, so the checkout token must not be left in .git/config
+      # where that script could read it. No step here needs authenticated
+      # git access.
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Show the bash under test
         run: /bin/bash --version
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -576,6 +576,12 @@ else
     }
 fi
 
+# Catalogue rows are packed with the ASCII unit separator rather than a tab.
+# Tab is an IFS whitespace character, so `IFS=$'\t' read` collapses runs of
+# tabs into a single delimiter and shifts later fields left, corrupting any
+# row with an empty column (ollama_local/ollama_cloud have two). 0x1f is not
+# IFS whitespace, so empty fields survive on every bash, 3.2 included.
+CATALOG_FS=$'\037'
 MODEL_DEFAULT_ROWS=""
 MODEL_CHOICE_ROWS=""
 PRESET_ROWS=""
@@ -589,12 +595,12 @@ load_model_catalog_rows() {
 from framework.llm.model_catalog import get_default_models, get_models_catalogue, get_presets
 
 for provider_id, default_model in sorted(get_default_models().items()):
-    print(f"DEFAULT\t{provider_id}\t{default_model}")
+    print(f"DEFAULT\x1f{provider_id}\x1f{default_model}")
 
 for provider_id, models in sorted(get_models_catalogue().items()):
     for model in models:
         print(
-            "MODEL\t{provider}\t{id}\t{label}\t{max_tokens}\t{max_context_tokens}".format(
+            "MODEL\x1f{provider}\x1f{id}\x1f{label}\x1f{max_tokens}\x1f{max_context_tokens}".format(
                 provider=provider_id,
                 id=model["id"],
                 label=model["label"],
@@ -605,7 +611,7 @@ for provider_id, models in sorted(get_models_catalogue().items()):
 
 for preset_id, preset in sorted(get_presets().items()):
     print(
-        "PRESET\t{preset_id}\t{provider}\t{model}\t{max_tokens}\t{max_context_tokens}\t{api_key_env_var}\t{api_base}".format(
+        "PRESET\x1f{preset_id}\x1f{provider}\x1f{model}\x1f{max_tokens}\x1f{max_context_tokens}\x1f{api_key_env_var}\x1f{api_base}".format(
             preset_id=preset_id,
             provider=preset["provider"],
             model=preset.get("model", ""),
@@ -617,7 +623,7 @@ for preset_id, preset in sorted(get_presets().items()):
     )
     for choice in preset.get("model_choices", []):
         print(
-            "PRESET_MODEL\t{preset_id}\t{id}\t{label}\t{recommended}".format(
+            "PRESET_MODEL\x1f{preset_id}\x1f{id}\x1f{label}\x1f{recommended}".format(
                 preset_id=preset_id,
                 id=choice["id"],
                 label=choice["label"],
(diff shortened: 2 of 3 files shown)


woahwhattheheck:
Both comments from the 21:22 review are addressed in `09aed79`.

**1. `persist-credentials` on the test job — agreed, fixed.**

The point stands: `quickstart-parsing` runs on `pull_request` and executes
`tests/test_quickstart_parsing.sh` straight out of the checkout, so leaving
`GITHUB_TOKEN` in `.git/config` puts it within reach of code from the pull
request. No step in the job needs authenticated git.

```yaml
      - uses: actions/checkout@v4
        with:
          persist-credentials: false
```

Scoped to this PR's own job only. The other four `actions/checkout` steps carry
the same default, but they are pre-existing and belong to a workflow-hardening
change rather than a parser fix — happy to do them in a separate PR if that is
wanted.

**2. `MODEL` / `PRESET_MODEL` assertions — added, and they caught something.**

Both collections are now read back through their real consumers rather than
asserted as packed strings, so the loader and the reader are both covered:

```
MODEL        -> get_model_choice_count / _id / _label / _maxtokens / _maxcontexttokens
PRESET_MODEL -> get_preset_model_choice_count / _field
```

`get_preset_model_choice_*` is defined *after* `apply_preset()`, so the test's
extraction range (`/^CATALOG_FS=/,/^apply_preset() {/`) cut it off. Widened to
`/^CATALOG_FS=/,/^HIVE_CONFIG_DIR=/`. I checked what that pulls in: 202 lines,
top level is the same five collection initialisers and nothing else — everything
added is a function definition, and `bash -n` is clean on the region. The
`declare -f` guard now lists all eleven functions, so a renamed or moved reader
fails loudly instead of silently skipping.

While writing those I added a `MODEL` row with an **empty label**:

```
MODEL${FS}ollama${FS}llama3.1${FS}${FS}4096${FS}8192
```

That is this PR's bug in a row type the test never covered — under a tab
separator the collapse shifts both token counts left into the label's slot.

**Checks**

| | result |
| --- | --- |
| `bash tests/test_quickstart_parsing.sh` on the fix | **33/33 pass** (was 20/20) |
| same test, separator reverted to `$'\t'` (nothing else changed) | **13 of 33 fail** (was 10 of 20) |

The three extra failures in the control are exactly the new empty-label `MODEL`
assertions — `empty label stays empty`, `max_tokens not shifted`,
`max_context not shifted` — so that row type really was an uncovered instance of
the same defect, not just added ceremony.

`ci.yml` still parses to the same 5 jobs, and the diff against the previous
commit touches only the `quickstart-parsing` block.

PR #7423, opened 2026-09-06
fix(subdomain-enumerator): reject non-positive max_results before querying crt.sh

### LINKED ISSUE #7406: [Bug]: subdomain_enumerate accepts non-positive max_results
## Describe the Bug

`subdomain_enumerate` accepts a negative `max_results` value. The implementation
only applies an upper bound and then uses the value as a Python slice, so
`max_results=-1` returns every discovered subdomain except the final one. It also
makes the crt.sh request even though the requested limit is invalid.

This contradicts the parameter's role as a maximum and can return substantially
more data than the caller intended.

## To Reproduce

1. Register the subdomain enumerator tools on a `FastMCP` instance.
2. Mock `httpx.AsyncClient.get` to return three crt.sh records:
   `a.example.com`, `b.example.com`, and `c.example.com`.
3. Call `await subdomain_enumerate("example.com", max_results=-1)`.
4. Observe that two subdomains are returned because the implementation evaluates
   `subdomains[:-1]`.

The same path is visible in the focused test suite by adding an assertion that a
non-positive limit is rejected before `httpx.AsyncClient` is constructed.

## Expected Behavior

Non-positive `max_results` values should return a structured validation error
before any network request. Positive values should retain the existing behavior,
including the upper cap of 200.

## Screenshots

N/A — this is a backend input-validation issue with text-only output.

## Environment

- OS: Linux (WSL2, x86_64)
- Python version: 3.11.15

## Configuration

N/A — no credentials or agent configuration are required. The reproduction uses
a mocked crt.sh response.

## Logs

```text
max_results=-1
observed: two of three sorted subdomains are returned
expected: {"error": "max_results must be at least 1"}
```

## Additional Context

The root cause is the upper-only clamp `min(max_results, 200)` followed by
`subdomains[:max_results]`. A focused fix can validate `max_results < 1` before
creating the HTTP client and retain the current upper clamp for values above 200.

Open and closed issue/PR searches for `subdomain max_results`, `subdomain
negative limit`, and `subdomain_enumerate max_results` found no relevant match.

woahwhattheheck:
Fixes #7406

`subdomain_enumerate` bounded `max_results` from above but not from below, so a non-positive value fell through to a Python slice and the crt.sh request went out regardless.

## Reproduction, before the change

Driving the current `subdomain_enumerate` with a mocked crt.sh returning `a.example.com`, `b.example.com`, `c.example.com`:

| Call | crt.sh queried | Error | `subdomains` | `total_found` |
| --- | --- | --- | --- | --- |
| `max_results=-1` | **yes** | no | `['a.example.com', 'b.example.com']` | 2 |
| `max_results=0` | **yes** | no | `[]` | 0 |

`-1` returns everything except the final subdomain, because `min(-1, 200)` is `-1` and line 139 evaluates `subdomains[:-1]`. `0` silently yields an empty list. Both spend a network round trip first.

## Change

One guard, placed after the domain is cleaned and before the client is constructed:

```python
if max_results < 1:
    return {
        "error": f"max_results must be a positive integer (got {max_results})",
        "domain": domain,
    }

max_results = min(max_results, 200)
```

The error shape matches the ones already returned by this tool for timeouts and non-200 responses — `error` plus `domain` — so callers that already branch on `"error" in result` need no change. The 200 cap and every positive-value path are untouched.

## Tests

Three cases added to `TestInputValidation`, using the mocking style already in that file:

- `test_non_positive_max_results_rejected` — parametrized over `-1`, `0`, `-50`; asserts `MockClient.assert_not_called()`, so the rejection is proven to happen before any network work, plus the error text, the `domain` echo, and the absence of a `subdomains` key.
- `test_negative_max_results_does_not_slice_off_last_subdomain` — the exact scenario from the issue, asserting `mock_client.get.assert_not_called()`.
- `test_max_results_one_is_accepted` — guards the boundary, so `< 1` cannot drift into `<= 1`.

The existing `test_max_results_clamped` still passes unchanged.

Verified locally against the patched module: non-positive limits error with no client constructed; `max_results=1` returns exactly one subdomain; `max_results=500` still clamps and returns all available; the default path is unaffected.

## Note on the earlier PR

#7407 covered this issue and was closed by the requirements bot for the linked-issue assignment rule, not for anything found in the code — CodeRabbit reported no actionable comments on it. I do not have rights to self-assign #7406 (the API returns 403 "Must have admin rights to Repository"), so I have asked on the issue for the assignment that the requirements check needs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject `max_results` values below 1.
  * Invalid values now return a clear error without performing network requests.
  * `max_results=1` is accepted and correctly limits results to one subdomain.
* **Documentation**
  * Updated the tool documentation to specify that `max_results` must be at least 1.

### DIFF
--- a/tools/src/aden_tools/tools/subdomain_enumerator/subdomain_enumerator.py
+++ b/tools/src/aden_tools/tools/subdomain_enumerator/subdomain_enumerator.py
@@ -86,7 +86,7 @@ def register_tools(mcp: FastMCP) -> None:
 
         Args:
             domain: Base domain to enumerate (e.g., "example.com"). No protocol prefix.
-            max_results: Maximum number of subdomains to return (default 50, max 200).
+            max_results: Maximum number of subdomains to return (default 50, min 1, max 200).
 
         Returns:
             Dict with discovered subdomains, interesting findings,
@@ -97,6 +97,12 @@ def register_tools(mcp: FastMCP) -> None:
         domain = domain.split("/")[0]
         if ":" in domain:
             domain = domain.split(":")[0]
+
+        if max_results < 1:
+            return {
+                "error": f"max_results must be a positive integer (got {max_results})",
+                "domain": domain,
+            }
 
         max_results = min(max_results, 200)
 
--- a/tools/tests/tools/test_subdomain_enumerator.py
+++ b/tools/tests/tools/test_subdomain_enumerator.py
@@ -90,6 +90,54 @@ class TestInputValidation:
             # Result should not error
             assert "error" not in result
 
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize("max_results", [-1, 0, -50])
+    async def test_non_positive_max_results_rejected(self, enumerate_fn, max_results):
+        with patch("httpx.AsyncClient") as MockClient:
+            result = await enumerate_fn("example.com", max_results=max_results)
+
+            # Rejected before any network work is attempted
+            MockClient.assert_not_called()
+            assert "error" in result
+            assert "max_results" in result["error"]
+            assert result["domain"] == "example.com"
+            assert "subdomains" not in result
+
+    @pytest.mark.asyncio
+    async def test_negative_max_results_does_not_slice_off_last_subdomain(self, enumerate_fn):
+        """max_results=-1 previously fell through to subdomains[:-1]."""
+        with patch("httpx.AsyncClient") as MockClient:
+            mock_client = AsyncMock()
+            mock_client.get.return_value = _mock_crtsh_response(
+                ["a.example.com", "b.example.com", "c.example.com"]
+            )
+            mock_client.__aenter__.return_value = mock_client
+            mock_client.__aexit__.return_value = None
+            MockClient.return_value = mock_client
+
+            result = await enumerate_fn("example.com", max_results=-1)
+
+            assert "error" in result
+            assert "subdomains" not in result
+            mock_client.get.assert_not_called()
+
+    @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


woahwhattheheck:
Deferring this PR to @sdivyanshu90, who claimed #7406 on 2026-09-04 — before I opened this — and said they had reproduced the negative-slice behavior and prepared validation and regression coverage. Their earlier PR #7407 was closed by the requirements bot over the assignment rule, not over anything in its code.

Under the assignment-first workflow in #472 that claim is a queue position, and I should have read the issue thread for it rather than treating "no open PR" as availability. That's on me. Maintainers: please give @sdivyanshu90 priority here. I'm happy to close this PR if you'd rather assign them — just say the word and I'll close it myself; I won't touch anything otherwise.

Leaving the technical detail in the thread so it's useful to whoever implements it:

- `min(-1, 200)` is `-1`, so line 139 evaluates `subdomains[:-1]` and returns every subdomain except the last.
- Against a mocked crt.sh returning three subdomains: `max_results=-1` queries crt.sh, returns no error, and yields 2 of 3; `max_results=0` queries crt.sh and yields `[]`.
- The guard belongs after domain cleaning and before the client is constructed, so the request is never made — `MockClient.assert_not_called()` is the assertion that pins that.
- `{"error": ..., "domain": domain}` matches the shape this tool already returns for timeouts and non-200s, so existing `"error" in result` callers need no change.
- Worth covering `max_results=1` too, so a later edit can't drift `< 1` into `<= 1`.

PR #7425, opened 2026-09-06
fix(janitor): resume worker deep-clean until every target is gone

### LINKED ISSUE #7375: [Bug]: Janitor's worker deep-clean crash-resume check only verifies conversations/ is gone, leaving data/ and stray files permanently unswept
## Description

`deep_clean_worker()` in `core/framework/maintenance/retention.py` is tier-2 of the janitor: it deletes a finished colony worker's `conversations/`, `data/`, and stray session-root files, after first writing a `result.json` tombstone. Its docstring makes a crash-safety promise:

> Order is crash-safe: the tombstone is written before any deletion so (a) resume_worker refuses the worker from that point on and (b) **a rerun after a crash simply finishes the deletions**.

That promise only half-holds. The crash-resume check only verifies that `conversations/` is gone — not `data/`, not the stray files the same function is responsible for removing:

```python
# core/framework/maintenance/retention.py
tombstone_exists = result_path.exists()
already_cleaned = tombstone_exists and not (wdir / "conversations").exists()
if already_cleaned:
    return report
```

The actual deletion loop below processes `conversations`, then `data`, then any stray files in the worker dir, one target at a time (each wrapped in its own try/except). If the janitor process is killed — OOM, host restart, disk full, container preemption, `SIGKILL` — after `conversations/` has been removed but before `data/` and the stray files are reached, the tombstone is already on disk (it's written first). On the **next** janitor run, `already_cleaned` evaluates `True` immediately and the function returns without ever touching `data/` or the leftover files — permanently. There's no further trigger that re-examines this worker; tier 2 treats it as fully cleaned from then on.

## Location

```python
# core/framework/maintenance/retention.py, deep_clean_worker()
tombstone_exists = result_path.exists()
already_cleaned = tombstone_exists and not (wdir / "conversations").exists()
if already_cleaned:
    return report
```

Called from `core/framework/maintenance/janitor.py:311` as part of the normal tier-2 `run_once(tiers={2})` sweep — no unusual configuration needed to reach it.

## Steps to Reproduce

Verified directly against the real function (unmodified `core/framework/maintenance/retention.py`), simulating the exact post-crash state (tombstone written, `conversations/` already removed, `data/` and a stray file still present — i.e. the crash landed between the two `for sub in ("conversations", "data")` iterations):

```python
from framework import config
config.HIVE_HOME = hive_home          # temp dir
config.COLONIES_DIR = colonies_dir

from framework.maintenance.retention import deep_clean_worker, DeleteDisposer, Manifest

wdir = colonies_dir / "c1" / "workers" / wid
(wdir / "data").mkdir(parents=True)
(wdir / "data" / "web_scrape_1.txt").write_text("x" * 500)
(wdir / "meta.json").write_text('{"worker_id": "..."}')
(wdir / "tasks.json").write_text('{"tasks": []}')
(wdir / "reminder_state.json").write_text('{"turns_total": 3}')
# conversations/ is intentionally NOT created — it's already gone, simulating
# a crash that finished that half of the deletion loop.
(wdir / "result.json").write_text('{"status": "completed", ...}')  # tombstone already written

report = deep_clean_worker("c1", wid, wdir, disposer=DeleteDisposer(), manifest=Manifest())
```

**Actual output:**
```
report.files: 0   report.bytes_freed: 0
data/ exists: True
data/web_scrape_1.txt exists: True
reminder_state.json exists: True
```

`deep_clean_worker()` returns immediately having done nothing, and will continue to do nothing on every subsequent janitor run — `data/` and the stray file are permanently retained.

## Test Coverage Gap

`core/tests/test_janitor_worker_deepclean.py::test_crash_resume_converges` does cover a crash-resume scenario, but only the case where the crash happens *before* any deletion starts (both `conversations/` **and** `data/` still present). It never exercises the partial state where `conversations/` alone is already gone — which is exactly the gap above, and exactly the more likely crash point in practice since `conversations/` (transcript files) is processed first and is typically larger/slower to delete than `data/`.

## Impact

Silent, permanent leak of worker artifact data — `data/` can hold tool output, scraped files, or other generated content, and stray session-root files are left behind too. This is retention policy's own job to purge, and it fails without any error, log line, or manifest entry: the tombstone alone makes the worker look "fully cleaned" from then on. Worse than an ordinary leak because re-running the janitor — retention's own recovery mechanism — cannot fix it; the leaked files are permanently invisible to tier 2 once `conversations/` is gone.

## Expected Behavior

The crash-resume check should confirm every target the loop is responsible for is actually gone (`conversations/`, `data/`, and no leftover non-keep-file stray files) before treating the worker as fully cleaned — not just `conversations/`. Concretely, `already_cleaned` should also check `not (wdir / "data").exists()` and that no stray files remain, or (more simply) the tombstone-exists fast path should be dropped in favor of just re-running the idempotent deletion loop, which already no-ops correctly (per `test_deep_clean_is_idempotent`) when nothing is left to delete.

## Environment

- OS: Windows 11 (repro is platform-independent — pure logic bug, not OS-specific)
- Python: 3.14.4

## Additional Context

Happy to open a PR fixing the crash-resume check plus a regression test covering this exact partial-crash state, if this is a direction the maintainers want.

woahwhattheheck:
Fixes #7375

`deep_clean_worker()` removes three things — `conversations/`, `data/`, and stray session-root files — but its crash-resume short-circuit only asked about the first one:

```python
already_cleaned = tombstone_exists and not (wdir / "conversations").exists()
```

The tombstone is written before any deletion, so if the janitor dies between the `conversations/` and `data/` disposals, the next run sees a tombstone plus a missing `conversations/`, returns immediately, and nothing ever revisits that worker. `data/` and the stray files stay on disk permanently.

## Reproduction, against the unmodified function

Building a finished worker, writing the tombstone, removing `conversations/` — the exact state a crash between the two `for sub in ("conversations", "data")` iterations leaves behind — then running `deep_clean_worker()` again:

| | leftovers after the rerun |
| --- | --- |
| before | `['data', 'stray_session_file.log']` |
| after | `[]` |

## Change

The list of things this function is responsible for removing is now computed once, by a small helper, and used for both decisions:

```python
def _deep_clean_targets(wdir: Path) -> list[Path]:
    """Everything deep_clean_worker is responsible for removing."""
    ...

targets = _deep_clean_targets(wdir)
already_cleaned = tombstone_exists and not targets
```

The deletion loop then iterates that same list instead of rebuilding it, so the resume check and the loop cannot drift apart if a fourth target is ever added.

Moving the enumeration above the tombstone write is safe: `result.json` is in `_WORKER_KEEP_FILES`, so writing it never adds a target. Everything else — tombstone-first ordering, the keep-set, `_delete_message_index_subtrees`, dry-run behaviour, archive mode, per-target `try/except` — is untouched.

The docstring's crash-safety promise now holds as written, so I extended it to say what makes it hold rather than leaving the claim bare.

## Tests

Three cases added to `core/tests/test_janitor_worker_deepclean.py`:

- `test_crash_resume_after_conversations_removed` — the reported state. Asserts `data/`, the stray `os` file and `reminder_state.json` are all gone, only the keep-set remains, and the existing tombstone is preserved byte-for-byte.
- `test_crash_resume_with_only_stray_files_left` — both subtrees already gone, strays remaining.
- `test_fully_cleaned_worker_short_circuits` — nothing left to remove still does zero work and leaves the tombstone alone, so the fix does not turn the short-circuit into a no-op.

Run against the module before and after the change: the first two **fail** on the current code (`data/ still present`) and pass after; the third passes both ways, which is the point of including it.

The existing `test_crash_resume_converges` passes unchanged — it leaves `conversations/` in place, so it never reached this path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup recovery after interruptions, ensuring remaining worker data and pending index copies are removed on subsequent runs.
  - Prevented incomplete or failed scans from being treated as finished.
  - Preserved cleanup markers when no further work is required.
  - Ensured cleanup continues when leftover index data remains, even if the main worker directory appears clean.
  - Allowed healthy data to be cleaned while inaccessible data is retried later.

- **Tests**
  - Added coverage for interrupted cleanup, scan failures, resume scenarios, and pending index data.

### DIFF
--- a/core/framework/maintenance/retention.py
+++ b/core/framework/maintenance/retention.py
@@ -609,6 +609,22 @@ def _extract_worker_summary(wdir: Path) -> tuple[str, str]:
     return "unknown", ""
 
 
+def _message_index_subtrees(session_id: str) -> list[Path]:
+    """Index copies of one session: .message_index/{events,data,meta}/*/*/<id>.
+
+    Read by both the deleter and the completion check, so the two cannot
+    disagree about whether the index side is finished.
+    """
+    index_root = config.HIVE_HOME / ".message_index"
+    subs: list[Path] = []
+    for tree in ("events", "data", "meta"):
+        base = index_root / tree
+        if not base.is_dir():
+            continue
+        subs.extend(base.glob(f"*/*/{session_id}"))
+    return subs
+
+
 def _delete_message_index_subtrees(session_id: str, disposer: Disposer, manifest: Manifest, tier: int, target: str) -> int:
     """Remove .message_index/{events,data,meta}/*/*/<session_id> subtrees.
 
@@ -616,50 +632,51 @@ def _delete_message_index_subtrees(session_id: str, disposer: Disposer, manifest
     so leaving it behind both strands bytes and serves ghost results
     from search_messages after the source is pruned.
     """
-    index_root = config.HIVE_HOME / ".message_index"
     freed = 0
-    for tree in ("events", "data", "meta"):
-        base = index_root / tree
-        if not base.is_dir():
-            continue
-        for sub in base.glob(f"*/*/{session_id}"):
-            item = PruneItem(
-                path=str(sub),
-                bytes=0,
-                tier=tier,
-                target=target,
-                action="delete",
-                reason="message index copy of pruned session",
-            )
-            try:
-                _, item.bytes = disposer.dispose_dir(sub)
-                item.outcome = "candidate" if disposer.dry_run else "done"
-                freed += item.bytes
-            except (OSError, ValueError) as exc:
-                item.outcome = "error"
-                item.error = str(exc)
-            manifest.add(item)
+    for sub in _message_index_subtrees(session_id):
+        item = PruneItem(
+            path=str(sub),
+            bytes=0,
+            tier=tier,
+            target=target,
+            action="delete",
+            reason="message index copy of pruned session",
+        )
+        try:
+            _, item.bytes = disposer.dispose_dir(sub)
+            item.outcome = "candidate" if disposer.dry_run else "done"
+            freed += item.bytes
+        except (OSError, ValueError) as exc:
+            item.outcome = "error"
+            item.error = str(exc)
+        manifest.add(item)
     return freed
 
 
-def _deep_clean_targets(wdir: Path) -> list[Path]:
+def _deep_clean_targets(wdir: Path) -> tuple[list[Path], bool]:
     """Everything :func:`deep_clean_worker` is responsible for removing.
 
     Used both to decide whether a previous run finished and to drive the
     deletion loop, so the two can never disagree.
+
+    Returns the targets and whether the scan actually completed. An
+    unreadable directory yields no targets, which is indistinguishable
+    from a finished worker unless the failure is reported separately --
+    and treating it as finished would strand the contents permanently.
     """
     targets: list[Path] = []
     for sub in ("conversations", "data"):
         p = wdir / sub
         if p.exists():
             targets.append(p)
+    scan_complete = True
     try:
         for p in wdir.iterdir():
             if p.is_file() and p.name not in _WORKER_KEEP_FILES:
                 targets.append(p)
     except OSError:
-        pass
-    return targets
+        scan_complete = False
+    return targets, scan_complete
 
 
 def deep_clean_worker(
@@ -683,8 +700,13 @@ def deep_clean_worker(
     result_path = wdir / "result.json"
 
     tombstone_exists = result_path.exists()
-    targets = _deep_clean_targets(wdir)
-    already_cleaned = tombstone_exists and not targets
+    targets, scan_complete = _deep_clean_targets(wdir)
+    # Index copies are removed further down, after this return, so a worker
+    # with no local targets left can still owe the index side. Skipping on
+    # an incomplete scan for the same reason: unknown is not done.
+    already_cleaned = (
+        tombstone_exists and scan_complete and not targets and not _message_index_subtrees(worker_id)
+    )
     if already_cleaned:
         return report
 
--- a/core/framework/maintenance/retention.py
+++ b/core/framework/maintenance/retention.py
@@ -642,6 +642,26 @@ def _delete_message_index_subtrees(session_id: str, disposer: Disposer, manifest
     return freed
 
 
+def _deep_clean_targets(wdir: Path) -> list[Path]:
+    """Everything :func:`deep_clean_worker` is responsible for removing.
+
+    Used both to decide whether a previous run finished and to drive the
+    deletion loop, so the two can never disagree.
+    """
+    targets: list[Path] = []
+    for sub in ("conversations", "data"):
+        p = wdir / sub
+        if p.exists():
+            targets.append(p)
+    try:
+        for p in wdir.iterdir():
+            if p.is_file() and p.name not in _WORKER_KEEP_FILES:
+                targets.append(p)
+    except OSError:
+        pass
+    return targets
+
+
 def deep_clean_worker(
     colony_id: str,
     worker_id: str,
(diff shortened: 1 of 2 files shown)


woahwhattheheck:
Deferring this PR to @chiruu12, who claimed #7375 on 2026-08-18 — well before I opened this — with essentially this plan: make the crash-resume short-circuit confirm `data/` and the stray files are gone too, not just `conversations/`. @Amiirhosseini also offered on 2026-08-28.

Under the assignment-first workflow in #472 those comments are queue positions, and I treated "no open PR" as availability instead of reading the thread. That's my error. Maintainers: please give @chiruu12 priority. I'll close this PR myself if you'd rather assign them — just say so; I won't touch anything otherwise.

Leaving the findings in the thread so they're useful to whoever implements it:

- Reproduced against the unmodified function in the exact post-crash state (tombstone written, `conversations/` disposed, `data/` and a stray file still present): the rerun leaves `['data', 'stray_session_file.log']` behind permanently, because `already_cleaned` short-circuits on `conversations/` alone.
- Computing the target list once and using it for both the resume check and the deletion loop keeps the two from drifting apart if a fourth target is ever added.
- Enumerating targets before the tombstone write is safe: `result.json` is in `_WORKER_KEEP_FILES`, so writing it never adds a target.
- The existing `test_crash_resume_converges` does not cover this path — it leaves `conversations/` in place, so it passes either way. A regression test has to remove `conversations/` first.
- Worth keeping a fully-cleaned-worker case as well, so the fix can't quietly turn the short-circuit into a no-op: it should still do zero work and leave the tombstone untouched.

woahwhattheheck:
Both findings confirmed against the branch and fixed in `fa59030`.

**1. Message-index copies were being stranded by the short-circuit.**

Confirmed rather than assumed. `_build_worker()` plants index copies at
`.message_index/{events,data,meta}/colonies/c1/<wid>`, and
`_delete_message_index_subtrees()` runs *after* the `already_cleaned` return, so
a tombstoned worker whose directory was empty returned early and left them. On
the unpatched module:

```
index copies stranded: .message_index/events/colonies/c1/session_..._deadbeef
                       .message_index/data/colonies/c1/session_..._deadbeef
                       .message_index/meta/colonies/c1/session_..._deadbeef
```

The subtree lookup is now `_message_index_subtrees()`, read by both the deleter
and the completion check — the same "the two can never disagree" shape
`_deep_clean_targets` already used — and the short-circuit requires it to be
empty.

**2. A failed target scan read as a clean worker.**

`_deep_clean_targets()` caught `OSError` from `iterdir()` and returned an empty
list, which is byte-identical to a finished worker. Worse than a one-off miss:
the tombstone is already written, so `resume_worker` refuses the worker and no
later pass retries — the contents are stranded permanently. It now returns
`(targets, scan_complete)` and the short-circuit requires a completed scan.

**3. The fixture, as you flagged.**

`test_fully_cleaned_worker_short_circuits` left the index copies in place, so it
asserted a short-circuit while index work was still outstanding — passing for the
wrong reason and encoding the bug. It now removes them first, which is what
"fully cleaned" has to mean once the completion check includes the index side.
Two regressions added: one that fails if index copies are stranded, one that
asserts a failed scan is reported rather than swallowed.

**Checks**

`core/` needs a full `uv sync` that is not available here, so I did not run the
repo's pytest suite. Instead I ran **the real `retention.py`** — not a
reimplementation — by stubbing only its three first-party imports
(`framework.config`, `framework.server.compaction_status`,
`framework.utils.io`; four names total: `HIVE_HOME`, `COLONIES_DIR`,
`QUEENS_DIR`, `get_status`) against a temp `HIVE_HOME`:

| Case | patched | unpatched |
| --- | --- | --- |
| index copies block the short-circuit | ok | **FAIL** — 3 subtrees stranded |
| failed scan is not a clean worker | ok | **FAIL** — no scan status returned |
| truly clean worker still short-circuits | ok | ok |
| full deep clean unchanged | ok | ok |
| crash resume with stray files unchanged | ok | ok |
| | **5/5** | **3/5** |

The three unchanged-behaviour cases pass either way, so the two failures isolate
these findings rather than reflecting a broken harness. `python -m py_compile`
clean on both files.

Stated plainly: the repo's own suite has **not** been run locally, so the CI job
is the authority on the tests as written. The `check-requirements` failure on
this PR is the linked-issue assignment gate, unrelated to this change.

PR #7435, opened 2026-09-08
fix(subdomain-enumerator): validate max_results lower bound

### LINKED ISSUE #7406: [Bug]: subdomain_enumerate accepts non-positive max_results

aniruddhaadak80:
## Description

Rejects non-positive `max_results` in `subdomain_enumerate` before any network request.

## Type of Change

- [ ] Documentation update

## Related Issues

Fixes #7406

## What was wrong

`subdomain_enumerate` only applied an upper clamp (`max_results = min(max_results, 200)`) and then sliced with `subdomains[:max_results]`. A negative value like `max_results=-1` therefore returned every discovered subdomain except the last one (via `subdomains[:-1]`), and `0` silently returned empty — all after needlessly querying crt.sh.

## Changes Made

- `tools/src/aden_tools/tools/subdomain_enumerator/subdomain_enumerator.py` (lines 103-104): added a lower-bound guard after domain cleaning and the existing 200-cap — `if max_results < 1: return {"error": "max_results must be at least 1", "domain": domain}` — using the same `{error, domain}` shape as the timeout/non-200 errors in that function. Positive behavior and the 200 cap are unchanged.

## Testing

- `py -m py_compile tools/src/aden_tools/tools/subdomain_enumerator/subdomain_enumerator.py` — passes.
- `tools/tests/tools/test_subdomain_enumerator.py` — all 15 existing tests pass.
- Functional check with mocked crt.sh (3 records): `max_results` of `-1`, `0`, `-50` each return `{"error": "max_results must be at least 1", "domain": "example.com"}` with `httpx.AsyncClient` never constructed; `max_results=1` returns 1 subdomain; `max_results=500` still clamps without error.
- `ruff check` and `ruff format --check` on the touched file — clean.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation to prevent subdomain enumeration requests with invalid result limits.
  - Requests specifying fewer than one result now return a clear error message instead of proceeding.
  - The existing maximum limit of 200 results remains unchanged.

### DIFF
--- a/tools/src/aden_tools/tools/subdomain_enumerator/subdomain_enumerator.py
+++ b/tools/src/aden_tools/tools/subdomain_enumerator/subdomain_enumerator.py
@@ -100,6 +100,9 @@ def register_tools(mcp: FastMCP) -> None:
 
         max_results = min(max_results, 200)
 
+        if max_results < 1:
+            return {"error": "max_results must be at least 1", "domain": domain}
+
         try:
             async with httpx.AsyncClient(timeout=30) as client:
                 response = await client.get(


aniruddhaadak80:
Could a maintainer please assign me to #7406? The check-requirements gate needs the PR author assigned to the linked issue, and self-assign via API is blocked on my side. Happy to push a retrigger once assigned.

PR #7456, opened 2026-09-09
fix(quickstart.sh): survive unset TERM and surface uv sync errors (fixes #7368)

### LINKED ISSUE #7368: quickstart.sh crashes on unset $TERM, and hides the real error on workspace install failure

Otto-Deviant1904:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of oversized data during workflow transitions by creating required storage locations automatically.
  - Added clear warning details when temporary data storage fails, with safe fallback behavior to keep processing moving.

- **Chores**
  - Improved quickstart setup reliability by tolerating terminal-clear failures.
  - Setup synchronization now reports success or failure clearly and displays relevant error details before exiting.

### DIFF
--- a/core/framework/orchestrator/node_worker.py
+++ b/core/framework/orchestrator/node_worker.py
@@ -760,11 +760,11 @@ class NodeWorker:
                 continue
             val_str = str(value)
             if len(val_str) > 300 and data_dir is not None:
-                data_dir.mkdir(parents=True, exist_ok=True)
                 ext = ".json" if isinstance(value, (dict, list)) else ".txt"
                 filename = f"output_{key}{ext}"
                 file_path = data_dir / filename
                 try:
+                    data_dir.mkdir(parents=True, exist_ok=True)
                     write_content = json.dumps(value, indent=2, ensure_ascii=False) if isinstance(value, (dict, list)) else str(value)
                     file_path.write_text(write_content, encoding="utf-8")
                     file_size = file_path.stat().st_size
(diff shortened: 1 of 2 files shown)


Otto-Deviant1904:
Closing this myself — issue #7368 was already assigned to @NT1906 with prior (if stale) work in #7371. Opening a competing PR without checking assignment status first was my mistake; withdrawing it rather than pushing the conflict. Apologies for the noise.

Issue #7476, opened 2026-09-23
safe_eval: other unbounded ops can still build huge values (str.ljust, **, big-int)


KrishP147:
Follow-up from #7473 (sequence-repetition DoS in \`safe_eval\`, fixed in the PR for that issue).

While fixing #7473, found other ways to build arbitrarily large values through both \`safe_eval\` implementations (\`core/framework/graph/safe_eval.py\` and \`core/framework/graph/code_sandbox.py\`) that are still unbounded:

- \`'a'.ljust(10**8)\` succeeds — string-padding methods aren't bounded.
- \`**\` / big-int exponentiation (e.g. \`2**10**9\`) is unbounded in both evaluators.
- The sandbox's only real backstop is its execution timeout, which appears to use \`signal\`-based alarms — those likely don't work on Windows, so on that platform there may be no timeout protection at all for a runaway computation.

Not fixed as part of #7473 since it's a distinct set of operations/root causes, but flagging since it's the same vulnerability class (unbounded resource consumption in "safe" eval).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7477, opened 2026-09-23
fix(security): bound sequence repetition in safe_eval (DoS via 'a'*10**10)

### LINKED ISSUE #7473: safe_eval: sequence repetition (str/list/bytes * int) is not bounded, unlike **
## Summary
`safe_eval()` (core/framework/orchestrator/safe_eval.py) is the sandboxed evaluator for graph edge `conditional` expressions. It explicitly bounds `**` to prevent CPU/memory exhaustion (`_safe_pow`, capping exponent size and estimating result bit-length before computing), with tests covering it. The same attack works through `*`: `str * int`, `bytes * int`, and `list * int` go straight to `operator.mul` with **no size limit at all**.

## Repro
```python
from framework.orchestrator.safe_eval import safe_eval

safe_eval('2 ** 999999999')      # raises ValueError: Power exponent exceeds safe limit (1000)
safe_eval('"x" * 10_000_000_000')  # NOT blocked -- tries to allocate a 10GB string
```

## Why it matters
- `condition_expr` values are explicitly documented as something a Builder agent can generate dynamically from a goal, i.e. they can originate from LLM output that may itself be influenced by untrusted upstream content.
- The module's own per-node timeout (`_check_timeout`) only runs between AST node visits, not during the single C-level call that performs the allocation, so it doesn't reliably catch a single large `*`.
- The SIGALRM-based hard timeout in `_execution_timeout` is silently skipped off the main thread, and also silently skipped whenever the process already has an active interval timer owned by something else (`if current_delay > 0 or current_interval > 0: yield`) -- so in some deployments there is no timeout enforcement at all for this path.

## Proposed fix
Add a `_safe_mult` wrapper mirroring `_safe_pow`: before delegating to `operator.mul`, check whether either operand is a bounded-length sequence (`str`/`bytes`/`list`/`tuple`) being multiplied by an `int`, and reject it if the resulting length would exceed a safe cap (mirrors the existing `MAX_POWER_RESULT_BITS` pattern). Wire it into `SAFE_OPERATORS[ast.Mult]`.

I have a working patch (fix + 6 new tests) ready to open as a PR once this is assigned -- `ruff check`/`ruff format --check` clean, full `core/tests/` suite passes (2506 passed / 15 skipped / 0 failed).

I'd like to work on this!

KrishP147:
## Summary
Sequence repetition (\`x * n\` for str/bytes/bytearray/list/tuple) was unbounded in \`safe_eval\`, unlike \`**\`. A crafted expression like \`"a" * 10**10\` could exhaust memory/CPU — a DoS-shaped hole in what's meant to be a safety guard.

Fixes #7473

There are two separate \`safe_eval\` implementations in this codebase, and both had this hole:
- \`core/framework/graph/safe_eval.py\` — the AST interpreter used by \`edge.py\`. Its \`ast.Mult\` operator now routes through a new \`bounded_multiply\` helper.
- \`core/framework/graph/code_sandbox.py\` — runs real \`eval\`/\`exec\` with restricted builtins; used by \`judge.py\` and exported as the canonical \`safe_eval\` from \`core/framework/graph/__init__.py\`. An AST transformer now rewrites every \`*\`/\`*=\` into a call to the same \`bounded_multiply\` before compiling.

Bundled together since it's the same vulnerability class with the same fix, not two separate PRs.

## Bound
\`MAX_SEQUENCE_REPEAT_LENGTH = 100_000\` elements (module-level constant in \`safe_eval.py\`), checked *before* multiplying (\`len(seq) * count\`) so the oversized result is never allocated. Covers str/bytes/bytearray/list/tuple, both operand orders, nested repeats (checked at each step), and leaves zero/negative multipliers at normal Python behavior. Oversize raises \`ValueError\`. No existing length-cap constant in the codebase to match against — 100k is a conservative default (worst case under 1MB) and trivial to tune later.

## Second commit: closing a bypass found in review
Sandboxed code could rebind the injected \`bounded_multiply\` helper name to defeat the check entirely (e.g. \`__hive_bounded_multiply__ = lambda a, b: a\`). The second commit rejects any user-code reference to that name as a security violation.

## Known follow-up (not in scope here)
Other unbounded-growth ops remain in the sandbox (e.g. \`str.ljust(10**8)\`, \`**\`/big-int exponentiation), and the sandbox's \`signal\`-based execution timeout likely doesn't work on Windows. Filed separately as #7476.

## Test plan
- New \`core/tests/test_safe_eval.py\` + 3 new test classes in \`core/tests/test_flexible_executor.py\`: 61 passed.
- Manually verified: malicious repeats blocked; \`[1,2]*3\`, \`2.5*4\`, \`-1*'ab'\`, \`'ab'*3\` all work normally; helper-shadowing bypass now rejected.
- Full suite: 199 passed, 43 failed — confirmed the same 43 fail identically on the pre-fix base commit (pre-existing Windows/litellm environment issues, unrelated to this change).
- Ruff: no new errors introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Oversized repetitions of strings, bytes, and collections are now stopped before they can consume excessive resources. Repetitions are limited to 100,000 elements; smaller repetitions and ordinary numeric multiplication continue to work as expected.
  * Code that attempts to access or alter the protected repetition safeguard is rejected.


KrishP147:
Closing for now — pausing this contribution. Work stays on my fork (krish/bound-safe-eval-repetition) in case I pick it back up.

